PR #1483, opened 2024-08-14
QA - fix regression testing

eunjisong:
## Description
- Going through android/ios regression daily run result and fixed them 
- tested on both iOS and Android 

## Checklist


packages/core-mobile/e2e/tests/settings/showRecoveryPhrase.e2e.smoke.ts
```diff
@@ -9,7 +9,7 @@ import { warmup } from '../../helpers/warmup'
 import CreatePinPage from '../../pages/createPin.page'
 import SecurityAndPrivacyPage from '../../pages/burgerMenu/securityAndPrivacy.page'
 
-describe('Show Recovery Phrase', () => {
+describe('Setting - Show Recovery Phrase', () => {
```
Any2suited66, line 12: Testrail usually doesn't like special characters so let's keep an eye on this

Any2suited66 (approved):
See comment, otherwise it looks good

PR #1498, opened 2024-08-20
Moved the pattern of SafeAreaView down the hierarchy to fix the letter box issue

sroy3:
## Description

**Ticket: []** 

There previously was a `SafeAreaView` at the root of the app, which caused problems down the chain by restricting views to the inside of this safe area.

Following React Navigation [recommendations](https://reactnavigation.org/docs/handling-safe-area/), the top `SafeAreaView` was removed, and some padding was added to children's views using the safe area insets. Ideally, these views with padding would be used even closer to the last descendants, but this would have been too big of a change. I recommend doing so incrementally, section by section, when possible.

### Before
Simulator Screenshot - iPhone 15 Pro - 2024-08-19 at 22 50 55

### After
Simulator Screenshot - iPhone 15 Pro - 2024-08-19 at 22 28 38

## Testing

It should be easier to review without whitespaces:
Screenshot 2024-08-19 at 10 54 11 PM

I tried checking all possible screens, but I probably forgot some.

## Checklist


packages/core-mobile/app/screens/defi/DeFiProtocolDetails.tsx
```diff
@@ -102,7 +103,7 @@ export const DeFiProtocolDetails = (): JSX.Element => {
   }
 
   return (
-    <View style={styles.container}>
+    <SafeLowerAreaView sx={styles.container}>
```
atn4z7, line 106: is the WalletScreenStack too complex to apply SafeLowerAreaView? I see that you are doing it manually for some of its screens

sroy3: It's because the `DrawerView` is in it and needs to bleed out as well.

Wrapping screen stack was a way to make this change not to big at the moment, but eventually it would be great if these safe area views moved down the hierarchy to the view that needs it.

atn4z7 (dismissed):

sroy3:
Fix the previously mentioned bu. Also, I wasn't satisfied with Android not behaving the same way, so I dig into this a little more and figured out how to fix that as well.

packages/core-mobile/app/ContextApp.tsx
```diff
@@ -48,7 +48,7 @@ const ContextApp = (): JSX.Element => {
 
   return (
     <Sentry.ErrorBoundary fallback={<TopLevelErrorFallback />}>
-      <StatusBar barStyle={'light-content'} backgroundColor="black" />
+      <StatusBar translucent backgroundColor="#00000000" />
```
sroy3, line 51: `ContextProviders` is created conditionally below, so it is not possible to get access to the theme at that point to get the `theme.transparent` variable here, but all it does is making the  `StatusBar` transparent so we can truly get edge to edge. Transparent is transparent, whatever the color.

packages/core-mobile/app/navigation/wallet/EarnScreenStack/EarnScreenStack.tsx
```diff
@@ -35,64 +36,66 @@ const EarnStack = createStackNavigator<EarnStackParamList>()
```
atn4z7, line 36: these intro popups are being stretched too long and there is no spacing between it and the bottom tab. you can see the figma design for it [here](https://www.figma.com/design/6ETlkl6Gi5cMvSRIvoRP6y/Mobile?t=yZldRWlY3NisUFHF-0) 
Screenshot 2024-08-26 at 5 04 04 PM

packages/core-mobile/app/navigation/wallet/EarnScreenStack/EarnScreenStack.tsx
```diff
@@ -35,64 +36,66 @@ const EarnStack = createStackNavigator<EarnStackParamList>()
 
 function EarnScreenStack(): JSX.Element {
   return (
```
atn4z7, line 38: Screenshot 2024-08-26 at 5 27 46 PM

atn4z7: here are a few screenshots from the current core mobile and also the slack app for references 
Screenshot 2024-08-26 at 5 30 44 PM

packages/core-mobile/app/components/SafeAreaViews.tsx
```diff
@@ -0,0 +1,31 @@
+import { View, type SxProp } from '@avalabs/k2-mobile'
+import React, { type FC, type PropsWithChildren } from 'react'
+import { useSafeAreaInsets } from 'react-native-safe-area-context'
+
+type Props = {
+  sx?: SxProp
+  noPaddingTop?: boolean
+}
+
```
atn4z7, line 9: can you add some comments in this file to briefly explain when to use SafeVertical vs SafeLower?

atn4z7 (changes requested):
I did some testing on iOS and Android and found a few UI issues. other than that, the full screen/edge to edge experience looks good.

packages/core-mobile/app/components/SafeAreaViews.tsx
sroy3: I added some comments. Let me know if these aren't clear enough.

packages/core-mobile/app/navigation/wallet/EarnScreenStack/EarnScreenStack.tsx
sroy3: I fixed a lot of issues. I still need to to reproduce the time and icons missing. So I'm going to do another pass tomorrow.

sroy3: I'm unable to reproduce the status bar bug and I've tried different simulators and ios versions. I don't think I would know where the bug is coming from to try to fix it blindly. Is it still doing it on your side?

atn4z7: yes it's still happening. my iphone is 13 pro max and version 16.1.2

atn4z7: also can you increase the spacing to match the slack app for these 2 things?
- bottom tab icon + text and the end of screen
- bottom tab bar and the floating acting button

ruijialin-avalabs:
I just tried out internal ios build 0.0.0.3393.  looks like the testnet banner is misaligned

IMG_0159

packages/core-mobile/app/navigation/wallet/EarnScreenStack/EarnScreenStack.tsx
atn4z7: and I also got the same issue about the testnet banner as @ruijialin-avalabs did.

sroy3:
> I just tried out internal ios build 0.0.0.3393. looks like the testnet banner is misaligned
> 
> Screenshot 2024-08-30 at 12 57 58 PM

Thanks, I fixed it.

sroy3:
Spacing and Testnet banner should be fixed. 

I also tried to blindly fix the status bar bug, let me know if that worked (confidence it did is 50-50).

packages/core-mobile/app/navigation/wallet/EarnScreenStack/EarnScreenStack.tsx
atn4z7: btw I found out how to reproduce the status bar issue. it seems to happen with:
- iOS version < 17 
- dark mode is turned off

you can download iOS version 16.1 or 16.2 in Xcode to test. here is a video:
https://github.com/user-attachments/assets/a0df9021-b7c9-401c-ab34-d5fddd79d37b

packages/core-mobile/app/navigation/TopNavigationHeader.tsx
```diff
@@ -111,7 +113,7 @@ const TopNavigationHeader: FC<Props> = ({
   )
 
   return (
-    <View>
+    <View style={{ paddingTop: insets.top }}>
```
atn4z7, line 116: Screenshot 2024-09-03 at 11 07 00 PM

I did some more testing on Android and it looks like the bottom tab + top navigation header + floating action button still need adjustments. In the pic above, I took a few Android screenshots of the current Core Mobile as well as some popular apps for your references. Basically, for those 3 components, we just want our app to always match the others as best as we can. Let me know if you have any questions.

packages/core-mobile/app/navigation/wallet/EarnScreenStack/EarnScreenStack.tsx
sroy3: Finally was able to reproduce and fix.

packages/core-mobile/app/navigation/TopNavigationHeader.tsx
sroy3: Padding wasn't dynamic for the button before so I simply adjusted it, but I'll make it dynamic and that should look good at all times.

sroy3:
Closing in favour of #1689

PR #1508, opened 2024-08-23
Update avax and bip122 namespace accounts based on isTestnet

ruijialin-avalabs:
## Description

- the AddressWithCaip2ChainId construction is incorrect (see screenshot below), testnet and mainnet addresses are both used in mainnet and testnet namespace
- the chainList and accounts in session namespace should only contain the chains and accounts that is either in testnet or mainnet.

## Screenshots/Videos

- before

Screenshot 2024-08-22 at 5 22 36 PM

- after

https://github.com/user-attachments/assets/cb468eb5-4d84-45fe-a140-eb1abe83b6a8

## Checklist

- [X] I have added/updated necessary unit tests 

### DIFF
--- a/packages/core-mobile/app/services/walletconnectv2/WalletConnectService.ts
+++ b/packages/core-mobile/app/services/walletconnectv2/WalletConnectService.ts
@@ -4,12 +4,17 @@ import { IWeb3Wallet, Web3Wallet } from '@walletconnect/web3wallet'
 import { getSdkError } from '@walletconnect/utils'
 import Config from 'react-native-config'
 import { RpcError } from '@avalabs/vm-module-types'
-import { BlockchainNamespace } from '@avalabs/core-chains-sdk'
+import {
+  AvalancheCaip2ChainId,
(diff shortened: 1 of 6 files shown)


packages/core-mobile/app/services/walletconnectv2/WalletConnectService.ts
```diff
@@ -280,19 +285,93 @@ class WalletConnectService implements WalletConnectServiceInterface {
...
+              AvalancheCaip2ChainId.C_TESTNET,
+              AvalancheCaip2ChainId.P_TESTNET,
+              AvalancheCaip2ChainId.X_TESTNET
+            ]
+          : [
+              AvalancheCaip2ChainId.C,
+              AvalancheCaip2ChainId.P,
+              AvalancheCaip2ChainId.X
+            ]
+
```
atn4z7, line 323: updateSession is more about adding additional chain/account actual. it's not about replacing/deleting.

PR #1524, opened 2024-08-28
QA - fix regression tests

eunjisong:
## Description
- tested on both iOS and Android 
- It should pass some tests for the daily regression run. 

## Checklist


packages/core-mobile/e2e/locators/collectibles.loc.ts
```diff
@@ -21,7 +21,7 @@ export default {
   nftlogo: 'avatar__logo_avatar',
   warningAddressRequired: 'Address required',
   warningInsufficientFee: 'Insufficient balance for fee.',
-  warningGasLimitIsInvalid: 'Gas limit is invalid',
+  warningGasLimitIsInvalid: 'Unable to send token, invalid gags limit.',
```
Any2suited66, line 24: you have a typo here is `gags` that should be `gas`

eunjisong: Thank you for catching that! I've pushed the change.

Any2suited66 (approved):

PR #1527, opened 2024-08-28
CP-8935: upgrade paraswap to 6.9.1

ruijialin-avalabs:
## Description

**Ticket: [CP-8935]** 

- upgrade paraswap from v5 to v6
- remove paraswapUtil file and move the logic into SwapService

## Screenshots/Videos

https://github.com/user-attachments/assets/470034dd-d69a-4305-9307-ef88e2931ead

## Checklist

[CP-8935]: https://ava-labs.atlassian.net/browse/CP-8935?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/core-mobile/app/utils/Utils.ts
```diff
@@ -148,7 +147,7 @@ export function calculateGasAndFees<T extends TokenBaseUnit<T>>({
 }
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-export function isAPIError(rate: any): rate is APIError {
+export function isAPIError(rate: any): rate is Error {
   return typeof rate?.message === 'string'
 }
```
onghwan, line 152: `isAPIError` function seems unnecessary.

ruijialin-avalabs: removed.

packages/core-mobile/app/services/swap/SwapService.ts
```diff
@@ -73,49 +95,65 @@ class SwapService {
...
-            swapSide,
-            {},
+            amount: srcAmount,
+            userAddress: account.addressC,
+            side: swapSide,
             srcDecimals,
             destDecimals
-          )
+          })
         }
```
onghwan, line 122: I compared the `getRate` functions between the old and new sdk, and it seems not to return `Error` anymore. same for the `getSpender()` in the function below.

packages/core-mobile/app/swap/getSwapRate.ts
```diff
@@ -80,8 +79,8 @@ export async function getSwapRate({
     }
   }
 
-  if ((priceResponse as APIError).message) {
-    return { error: (priceResponse as APIError).message }
+  if ((priceResponse as Error).message) {
+    return { error: (priceResponse as Error).message }
   }
```
onghwan, line 84: > I compared the getRate functions between the old and new sdk, and it seems not to return Error anymore. 

if this is true, then I guess we can make `swapService.getSwapRate` function not to return error with `priceResponse` object.

packages/core-mobile/app/services/swap/SwapService.ts
ruijialin-avalabs: updated [here](3f8a29109e9e02dfb3c3e6ad6e4210d4fe81c60c)

packages/core-mobile/app/swap/getSwapRate.ts
ruijialin-avalabs: update [here](3f8a29109e9e02dfb3c3e6ad6e4210d4fe81c60c)

packages/core-mobile/app/services/swap/SwapService.ts
```diff
@@ -141,14 +133,19 @@ class SwapService {
...
+          const { getSpender } = constructGetSpender({
+            apiURL: this.getParaSwapSDK(network.chainId).apiURL,
+            chainId: ChainId.AVALANCHE_MAINNET_ID,
+            version: ParaSwapVersion.V6,
+            fetcher
+          })
+          return getSpender()
+        } catch (error) {
+          Logger.error('Failed to get Paraswap spender', error)
+          return undefined
```
onghwan, line 147: shouldn't we just throw an error here?

packages/core-mobile/app/services/swap/SwapService.ts
```diff
@@ -108,30 +108,22 @@ class SwapService {
...
-            userAddress: account.addressC,
-            side: swapSide,
-            srcDecimals,
-            destDecimals
-          })
-        }
-
-        return await retry({
-          operation: optimalRates,
-          isSuccess: result => (result as Error).message !== 'Server too busy'
```
onghwan: so does it mean that we don't have to retry with the new function?

packages/core-mobile/app/swap/getSwapRate.ts
```diff
@@ -64,27 +63,13 @@ export async function getSwapRate({
...
+      optimalRate: priceResponse,
+      destAmount: priceResponse.destAmount
     }
-  }
-
-  if (!priceResponse) {
+  } catch (error) {
     return {
-      error: 'no price response'
+      error: (error as Error).message
```
onghwan, line 72:
```suggestion
      error: error instanceof Error ? error.message : 'unknown error'
```

packages/core-mobile/app/services/swap/SwapService.ts
ruijialin-avalabs: Screenshot 2024-08-29 at 12 33 59 PM

if spenderAddress is undefined, it will failed to get the allowance, and throw `allowance error`

packages/core-mobile/app/services/swap/SwapService.ts
ruijialin-avalabs: Not sure why we needed that in the first place.  we don't simply retry whenever we can't fetch because `Server was busy` in this case.  we should just let it error out, and user can retry manually.

packages/core-mobile/app/services/swap/SwapService.ts
onghwan: I think it makes more sense to throw an error instead of returning undefined. The `Allowance Error` represents a different error path, and throwing an error aligns naturally with the existing pattern in the `getParaswapSpender` function, where other errors are already being thrown.​

ruijialin-avalabs: update [here](https://github.com/ava-labs/core-mobile/pull/1527/commits/c5d3c2b37dc9de8a76d33a883f84902890c575f3)

packages/core-mobile/app/contexts/SwapContext/performSwap/performSwap.ts
```diff
@@ -77,6 +75,13 @@ export async function performSwap({
 
   // no need to approve native token
   if (!isSrcTokenNative) {
+    const [spenderAddress, spenderAddressError] = await resolve<string>(
+      SwapService.getParaswapSpender(activeNetwork) ?? Promise.resolve(null)
+    )
+    if (spenderAddressError || spenderAddress === null) {
+      throw new Error(`Spender Address Error: ${spenderAddressError}`)
+    }
+
```
onghwan, line 84: I guess `await resolve` were used below because `contract.allowance`, `contract.approve` can be undefined, but here, I think we just can call `await SwapService.getParaswapSpender` with try catch.

packages/core-mobile/app/contexts/SwapContext/performSwap/performSwap.ts
```diff
@@ -51,22 +50,15 @@ export async function performSwap({
...
-    partnerAddress = undefined,
-    partner = 'Avalanche',
-    receiver = undefined,
-    permit = undefined,
-    deadline = undefined,
-    partnerFeeBps = undefined
+  const partner = 'Avalanche'
 
   let approveTxHash: string | undefined
+  let spenderAddress: string
```
onghwan, line 61: I think you can declare `spenderAddress` inside the `if (!isSrcTokenNative)` statement.

onghwan (approved):
🔥

PR #1536, opened 2024-08-29
Enable btc bridging via core web

atn4z7:
## Description
not sure if we ever have btc bridging working via core web but this pr adjusts the code to make that work.

## Screenshots/Videos
bridging via core web
https://github.com/user-attachments/assets/0aaeb5df-e237-4444-b706-c2bbef458925

in app bridging
https://github.com/user-attachments/assets/66266eb3-7d83-4ad6-a010-e570b6858934

## Testing
please make sure all kinds of bridging still work

## Checklist


packages/core-mobile/app/screens/bridge/Bridge.tsx
```diff
@@ -320,8 +323,8 @@ const Bridge: FC = () => {
 
       if (error || !hash) {
         // do not show the error when the user denied the transfer
-        if (error === 'User declined the transaction') {
-          Logger.error(error)
+        if (isUserRejectedError(error)) {
```
atn4z7, line 326: the error string is different now since we are using rpc errors. updated to use the same logic in other places to check for rejected error.

packages/core-mobile/app/screens/bridge/Bridge.tsx
```diff
@@ -330,6 +333,9 @@ const Bridge: FC = () => {
           return
         }
         setBridgeError(TRANSFER_ERROR)
+        showTransactionErrorToast({
+          message: getJsonRpcErrorMessage(error)
+        })
```
atn4z7, line 338: we now also show detailed bridge error via a toast

packages/core-mobile/app/screens/rpc/components/v2/BridgeAsset.tsx
```diff
@@ -27,17 +29,17 @@ const BridgeAsset = (): JSX.Element => {
...
-
   const { onUserApproved: onApprove, onUserRejected: onReject } =
     useDappConnectionV2()
-  const { activeNetwork } = useNetworks()
-
+  const { activeNetwork, networks } = useNetworks()
+  const bridgeAppConfig = useSelector(selectBridgeAppConfig)
   const theme = useContext(ApplicationContext).theme
-
-  const { data: networkFee } = useNetworkFee(activeNetwork)
```
atn4z7, line 37: we can't just use active network here anymore. for example, when users try to bridge some btc to avalanche via core web and the active network in mobile is avalanche, we don't want to use avalanche's network fee for this transaction. instead we need to infer the correct network to use based on the blockchain param in the request.

packages/core-mobile/app/store/rpc/providers/walletConnect/walletConnect.ts
```diff
@@ -50,7 +51,8 @@ class WalletConnectProvider implements AgnosticRpcProvider {
       const requestId = request.data.id
       const dappName = request.peerMeta.name
 
-      shouldShowErrorToast && showDappToastError(error.message, dappName)
+      shouldShowErrorToast &&
+        showDappToastError(getJsonRpcErrorMessage(error), dappName)
```
atn4z7, line 55: we now also show detailed error for wallet connect's failed request approvals

packages/core-mobile/app/store/rpc/providers/walletConnect/utils.ts
```diff
@@ -9,5 +8,15 @@ export const isSessionProposal = (
   return request.method === RpcMethod.WC_SESSION_REQUEST
 }
 
-export const isUserRejectedError = (error: RpcError): boolean =>
-  error.code === providerErrors.userRejectedRequest().code
+export const isUserRejectedError = (error: unknown): boolean => {
```
neven-s, line 11: would it be possible to make typing bit stronger here?
By looking just at function signature one would think it could be used anywhere, but it's actually specific for metamask/rpc-errors group of errors, and code number could clash with some other group of errors, right?

atn4z7: it's actually being used everywhere right now (in both react component and redux) so that's why I adjusted it to unknown. the function now also checks for the correct type or else it will return false.

neven-s (approved):

PR #1689, opened 2024-09-09
Remove letterboxing effect in app (edge to edge)

sroy3:
## Description

Recreating #1498 with signed commits.


packages/core-mobile/app/navigation/WalletScreenStack/WalletScreenStack.tsx
```diff
@@ -240,177 +241,180 @@ function WalletScreenStack(props: Props): JSX.Element {
     <BridgeProvider>
       <NFTMetadataProvider>
         {isTestnet && <TestnetBanner />}
```
atn4z7, line 243: Screenshot 2024-09-10 at 9 42 21 AM

packages/core-mobile/app/navigation/WalletScreenStack/WalletScreenStack.tsx
```diff
@@ -240,177 +241,180 @@ function WalletScreenStack(props: Props): JSX.Element {
...
-        <WalletScreenS.Navigator
-          screenOptions={{
-            headerShown: false
-          }}>
-          <WalletScreenS.Screen
-            name={AppNavigation.Wallet.Drawer}
```
atn4z7, line 248: Screenshot 2024-09-10 at 9 45 50 AM

packages/core-mobile/app/navigation/WalletScreenStack/WalletScreenStack.tsx
```diff
@@ -240,177 +241,180 @@ function WalletScreenStack(props: Props): JSX.Element {
...
-            component={DrawerScreenStack}
-          />
-          <WalletScreenS.Screen
-            options={{
-              ...MainHeaderOptions({ title: 'Manage token list' })
-            }}
-            name={AppNavigation.Wallet.TokenManagement}
-            component={TokenManagement}
-          />
-          <WalletScreenS.Screen
```
atn4z7, line 258: when toggling testnet mode, the animation/transition is kinda glitchy
https://github.com/user-attachments/assets/b4588b9b-a146-4595-85b3-6b22adac4f9f

atn4z7: vs the current core mobile
https://github.com/user-attachments/assets/2999327b-2508-4c7b-9074-daab566fba5c

atn4z7 (changes requested):
I retested the app and the status bar plus the floating action button look good now. However, there are still some issues with the bottom tab bar, the network dropdown and toggling testnet mode. I did the testing on Android.

packages/core-mobile/app/navigation/WalletScreenStack/WalletScreenStack.tsx
sroy3: Should be fixed now

packages/core-mobile/app/navigation/WalletScreenStack/WalletScreenStack.tsx
sroy3: This is something I fixed before opening this new PR and I can't reproduce at all anymore. Can you give me some info on the device you're using? That could help me pinpoint the problem

packages/core-mobile/app/navigation/WalletScreenStack/WalletScreenStack.tsx
sroy3: This wasn't an easy fix, but I was able to figure out something.

sroy3:
@atn4z7 waiting for more info on the network dropdown bug, other comments were resolved.

packages/core-mobile/app/navigation/WalletScreenStack/WalletScreenStack.tsx
atn4z7: I was testing on my Samsung S22+

sroy3:
Found an emulator skin to test and fix.

packages/core-mobile/app/navigation/NavUtils.tsx
```diff
@@ -106,8 +107,15 @@ export const TabButton = ({
   image,
   focused
 }: TabButtonProps): JSX.Element => {
+  const { bottom } = useSafeAreaInsets()
```
atn4z7, line 110: let's use the useVariableSafeAreaInsets hook everywhere to avoid confusion since `bottom` or `top` is the same between these two

packages/core-mobile/app/hooks/useVariableSafeAreaInsets.ts
```diff
@@ -1,15 +1,14 @@
 import { useSafeAreaInsets } from 'react-native-safe-area-context'
-import { useNetworks } from './networks/useNetworks'
+import { useSelector } from 'react-redux'
+import { selectIsDeveloperMode } from 'store/settings/advanced'
 
 export const useVariableSafeAreaInsets = (): {
   top: number
   bottom: number
   conditionalTop: number
```
atn4z7, line 8: can you add a comment to document what conditionalTop is used for?

packages/core-mobile/app/navigation/RootScreenStack.tsx
```diff
@@ -144,6 +146,12 @@ const WalletScreenStackWithContext: FC = () => {
 
 const RootScreenStack: FC = () => {
   const walletState = useSelector(selectWalletState)
+  const { top } = useVariableSafeAreaInsets()
+  const dispatch = useDispatch()
+
+  useEffect(() => {
+    dispatch(setMainTopInset(top))
```
atn4z7, line 153: how often does `top` change? also, we don't really store UI state in redux. cannot we just use the `top` value as is?

sroy3: The top value is variable depending on its parent (and device of course. This main top inset value is unavailable where it's meant to be used. You can see it being used for the `NetworkDropdown`. 

There might not have been a use before to store anything UI related in the state, but it's not a bad thing. It may or may not be necessary for the redesign.

I had to do it this way because the value comes from a hook. Otherwise I would have created a basic variable for it.

Let me see if I can replicate what the hook does without the need of a React component context and just export the variable instead.

packages/core-mobile/app/hooks/useVariableSafeAreaInsets.ts
sroy3: You're right. I'll try finding it a better name as well. It can be confusing which one to take (conditional to what?).

PR #1705, opened 2024-09-11
CP-9200: Don't fetch stakes when wallet is not active

atn4z7:
## Description

**Ticket: [CP-9200]** 

This is an attempt to address the glacier rate limit issue reported by QA. 

Problem: We were fetching stakes every 30 seconds even when wallet is locked. This is not necessary and can easily lead to rate limit issue with Glacier due to these reasons:
- a lot of requests need to be made when the account has lots of transactions on P-chain
- on some Android devices, the app can continue to run even when it has been minimized/backgrounded

[CP-9200]: https://ava-labs.atlassian.net/browse/CP-9200?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


onghwan (approved):
really nice catch.

PR #1805, opened 2024-09-23
Fix iOS CI

atn4z7:
1/ we have multiple targets to set versions now (our app and the notification extension service) so we need to manually update the version + build number for them all.
2/ they need to match each other.


packages/core-mobile/ios/AvaxWallet/AvaxWalletRelease.entitlements
```diff
@@ -2,6 +2,8 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+	<key>aps-environment</key>
+	<string>development</string>
```
atn4z7, line 6: this config was missing for internal build

ruijialin-avalabs (approved):

PR #1910, opened 2024-10-09
QA - New: Test manage token & NFT and plusIcon options by networks

eunjisong:
## Description
New
1. ManageToken - Go to manage token screen > toggle ON and OFF > Verify the token is visible and NOT visible 
2. ManageNFT - pretty much same idea with manage token but for NFT 
3. PlusIcon option testing - SWAP & BUY are ONLY available on C-chain. Test the plusIcon options by network 

## Checklist

### DIFF
--- a/packages/core-mobile/app/screens/tokenManagement/TokenManagementItem.tsx
+++ b/packages/core-mobile/app/screens/tokenManagement/TokenManagementItem.tsx
@@ -20,6 +20,7 @@ const TokenManagementItem: FC<Props> = ({ id, name, image, symbol }) => {
 
   const isBlacklisted = useSelector(selectIsTokenBlacklisted(id))
 
+  // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
   function handleChange() {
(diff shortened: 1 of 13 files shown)


packages/core-mobile/app/screens/tokenManagement/TokenManagementItem.tsx
```diff
@@ -20,6 +20,7 @@ const TokenManagementItem: FC<Props> = ({ id, name, image, symbol }) => {
 
   const isBlacklisted = useSelector(selectIsTokenBlacklisted(id))
 
+  // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
```
Any2suited66, line 23: we shouldn't be ignoring this.  get co-pilot and it will tell you which return type to use here.

eunjisong: done!

Any2suited66 (approved):
Looks good

PR #1936, opened 2024-10-11
new navigation system - expo-router

onghwan:
## Description

**Ticket: [CP-9315]** 

### @react-navigation/native(default)
https://github.com/user-attachments/assets/40a4ecf7-8429-4f75-9d35-9cf01e380aa7

#### iOS
https://github.com/user-attachments/assets/d08d8c53-382b-41bd-a35b-903d92a6dd9f

#### Android
https://github.com/user-attachments/assets/f926e036-374d-4bf6-85c9-eea6813402d5

### Deeplinks
core://portfolio
core://portfolio/assets
core://contracts
core://stake
core://track
core://browser
core://settings
core://settings/account
core://receive
core://notifications

## Blurred Bars
https://github.com/user-attachments/assets/5968ecb2-bb70-4c8c-8ac5-5d9e8218ac22

## Checklist

[CP-9315]: https://ava-labs.atlassian.net/browse/CP-9315?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/core-mobile/app.json
```diff
@@ -1,4 +1,16 @@
 {
   "name": "AvaxWallet",
-  "displayName": "Core X"
+  "displayName": "Core X",
+  "slug": "core-wallet",
```
atn4z7, line 4: what is this slug used for?

packages/core-mobile/app.json
```diff
@@ -1,4 +1,16 @@
 {
   "name": "AvaxWallet",
-  "displayName": "Core X"
+  "displayName": "Core X",
```
atn4z7, line 3:
```suggestion
  "displayName": "Core Mobile",
```

packages/core-mobile/babel.config.js
```diff
@@ -14,13 +14,12 @@ module.exports = {
           '.ios.js',
           '.ios.tsx'
         ],
-        root: ['./app'],
+        root: ['./app', './src'],
```
atn4z7, line 17: looks like you have app nested inside src. can we have only one of them?

packages/core-mobile/package.json
```diff
@@ -102,6 +102,11 @@
     "es6-promise": "4.2.8",
     "ethers": "6.8.1",
     "events": "3.3.0",
+    "expo": "50.0.21",
+    "expo-constants": "15.4.6",
+    "expo-linking": "6.2.2",
+    "expo-router": "3.4.10",
+    "expo-status-bar": "1.11.1",
```
atn4z7, line 109: are we using expo-status-bar anywhere?

packages/core-mobile/src/hooks/useHomeScreenHeader.tsx
```diff
@@ -0,0 +1,37 @@
...
+    () => (
+      <View style={{ flexDirection: 'row', gap: 12, marginRight: 12 }}>
+        <Link href="/receive/">Receive</Link>
+        <Link href="/notifications/">Notifications</Link>
+      </View>
+    ),
+    []
+  )
+
+  useLayoutEffect(() => {
```
atn4z7, line 27: is it possible to statically configure these in advanced? that will be preferred as using hook can affect performance

packages/core-mobile/app.json
onghwan: https://docs.expo.dev/versions/latest/config/app/#slug

it should be a url friendly name for our app, and it's required field in app.json. I believe expo uses this in their urls, like `https://expo.dev/@your-username/{slug}`, but I don't think we're exposing our app via expo.

packages/core-mobile/src/utils/screenOptions.tsx
```diff
@@ -0,0 +1,83 @@
...
+    borderTopLeftRadius: 40,
+    borderTopRightRadius: 40,
+    marginTop: 75
+  },
+  gestureEnabled: true,
+  gestureDirection: 'vertical',
+  cardStyleInterpolator: forModalPresentationIOS
+}
+
+function forModalPresentationIOS({
```
atn4z7, line 38: can you add a short description for this presentation? also is the `cardStyleInterpolator` prop iOS only?

packages/core-mobile/src/utils/screenOptions.tsx
```diff
@@ -0,0 +1,83 @@
...
+  StackCardInterpolatedStyle,
+  StackCardInterpolationProps,
+  StackNavigationOptions,
+  TransitionPresets
+} from '@react-navigation/stack'
+import { Animated } from 'react-native'
+import Grabber from 'components/navigation/Grabber'
+import BackBarButton from 'components/navigation/BackBarButton'
+
+export const stackNavigatorScreenOptions: StackNavigationOptions = {
```
atn4z7, line 12: nit: can you add comments for some of these consts to make it clearer when we should use them?

packages/core-mobile/package.json
onghwan: not directly, but expo-router depends on it(as a peerDependency) so we should have it.

https://docs.expo.dev/router/installation/#install-dependencies

packages/core-mobile/src/utils/screenOptions.tsx
onghwan: renamed to `cardStyleInterpolatorForModalPresentation`. Fyi, the name `forModalPresentationIOS` was from `react-navigation` transition presets, meaning that it was showing modals in IOS _style_. it works for both platforms.

packages/core-mobile/app/utils/debugging/DevDebuggingConfig.ts
```diff
@@ -10,7 +10,8 @@ const DevDebuggingConfig = {
   LOGBOX_IGNORED_WARNINGS: [],
   SHOW_DEMO_NFTS: false,
   API_MOCKING: false,
-  SENTRY_SPOTLIGHT: false
+  SENTRY_SPOTLIGHT: false,
+  K2_ALPINE: true
```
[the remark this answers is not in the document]
onghwan: fixed.

packages/core-mobile/app.json
onghwan: fixed.

packages/core-mobile/babel.config.js
onghwan: moved the expo-router app root to `/app/new` and revered babel configuration.

packages/core-mobile/src/hooks/useHomeScreenHeader.tsx
onghwan: I'm not sure how we can achieve this at the moment. I don't want to replace the entire navigation bar(like we're currently doing with `<TopNavigationHeader />`) because I want to keep their animations and behavior with left/right/back bar buttons and title view. We can revisit this in another pr once we come up with a better solution.

packages/core-mobile/app/utils/navigation/Stack.tsx
```diff
@@ -0,0 +1,16 @@
+import { ParamListBase, StackNavigationState } from '@react-navigation/native'
+import {
+  createStackNavigator,
+  StackNavigationEventMap,
+  StackNavigationOptions
+} from '@react-navigation/stack'
+import { withLayoutContext } from 'expo-router'
+
+const { Navigator } = createStackNavigator()
+
```
atn4z7, line 10: these 2 files I think should be moved to the `new` folder since they are only used in there

packages/core-mobile/app/hooks/useHomeScreenHeader.tsx
```diff
@@ -0,0 +1,34 @@
+import React, { useCallback, useLayoutEffect } from 'react'
+import { View } from 'react-native'
+import { Link, useNavigation } from 'expo-router'
+
+const useHomeScreenHeader = (): void => {
```
atn4z7, line 5: let's move these inside the `new` folder as well

packages/core-mobile/src/hooks/useHomeScreenHeader.tsx
onghwan: configured statistically. :D

packages/core-mobile/app/utils/navigation/Stack.tsx
onghwan: oh, but the `app/new` folder should have files only for screens and layouts used by `expo-router`. if we want to use `app/new` more generic, I think we should add another sub-folder in it and move those files into it.

packages/core-mobile/app/hooks/useHomeScreenHeader.tsx
onghwan: I statically configured bar buttons so this hook has been removed.

atn4z7 (approved):
great work!!

PR #1973, opened 2024-10-18
Add volta settings

neven-s:


packages/core-mobile/package.json
```diff
@@ -305,5 +305,9 @@
     "stream": "stream-browserify",
     "vm": "vm-browserify",
     "zlib": "browserify-zlib"
+  },
+  "volta": {
+    "node": "18.14.1",
+    "yarn": "1.22.10"
```
atn4z7, line 311: yarn will need to be >= 3.6.4. also, can you add this for all the package.json in the repo?

PR #2029, opened 2024-10-28
Fix duration too short error

neven-s:


packages/core-mobile/app/screens/earn/Confirmation/Confirmation.tsx
```diff
@@ -90,6 +90,9 @@ export const Confirmation = (): JSX.Element | null => {
 
   const { minStartTime, validatedStakingEndTime, validatedStakingDuration } =
     useValidateStakingEndTime(stakingEndTime, validatorEndTimeUnix)
+  const localValidatedStakingEndTime = useMemo(() => {
+    return new Date(validatedStakingEndTime.getTime())
```
atn4z7, line 94: do we not need to use UTCDate here?

atn4z7 (approved):

PR #2111, opened 2024-11-22
CP-9541: Fix 'No asset selected' error when bridging BTC.b to BTC

onghwan:
## Description

**Ticket: [CP-9541]** 

* Uses the correct CAIP-2 ID for the Non-evm networks.
* Moves caip2ChainIds.ts to utils since it's no longer temporary.


packages/core-mobile/app/screens/bridge/hooks/useBridgeType.ts
```diff
@@ -11,6 +11,6 @@ export const useBridgeType = (
       return undefined
     }
 
-    return bridgeAsset.destinations[getEvmCaip2ChainId(targetChainId)]?.[0]
+    return bridgeAsset.destinations[getCaip2ChainId(targetChainId)]?.[0]
```
onghwan, line 14: Only this line caused the issue. The other changes are from refactoring.

atn4z7 (approved):

PR #2127, opened 2024-11-25
CP-9562: Fix unable to claim rewards on testnet

atn4z7:
## Description

**Ticket: [CP-9562]** 

Issue 1: we were rounding up fee calculation which results in a case where actual balance - fee (rounded up) doesn't equal to actual balance - fee (raw). @neven-s adjusted the round up logic in token unit in the utils sdk to make it round down instead. it seems to be working. 

Issue 2: when calculating the estimated export P fee, we were using a fake amount that is less than available balance to try to estimate the fee. however, this is not reliable and usually much smaller than the actual required fee. instead of doing that, I have adjusted it so that we would try with the actual available amount. if it errors out with the insufficient funds message, we will grab the `missing amount` from the message and deduct it from the available amount. this will guarantee we always have enough balance.

## Screenshots/Videos
claim rewards
https://github.com/user-attachments/assets/fc5d3d36-4ac9-4262-aba9-c67ec3147d52

disable claim button and show error when fee is higher than balance
https://github.com/user-attachments/assets/a85fa0d1-3c53-46f2-9efb-f80f39f95bde

## Testing
Please make sure claiming rewards works

## Checklist

[CP-9562]: https://ava-labs.atlassian.net/browse/CP-9562?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core-mobile/app/services/wallet/WalletService.tsx
+++ b/packages/core-mobile/app/services/wallet/WalletService.tsx
@@ -41,6 +41,11 @@ import {
 import WalletInitializer from './WalletInitializer'
 import WalletFactory from './WalletFactory'
 import MnemonicWalletInstance from './MnemonicWallet'
+import {
+  getAssetId,
+  TESTNET_AVAX_ASSET_ID,
+  MAINNET_AVAX_ASSET_ID
+} from './utils'
 
 type InitProps = {
   mnemonic: string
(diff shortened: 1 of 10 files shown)


packages/core-mobile/app/services/wallet/WalletService.tsx
```diff
@@ -693,11 +690,7 @@ class WalletService {
       accountIndex,
       avaxXPNetwork
     )
-    const assetId = isDevnet(avaxXPNetwork)
-      ? DEVNET_AVAX_ASSET_ID
-      : avaxXPNetwork.isTestnet
-      ? TESTNET_AVAX_ASSET_ID
-      : MAINNET_AVAX_ASSET_ID
+    const assetId = getAssetId(avaxXPNetwork)
```
onghwan, line 693: are we ok not to use `getAssetId(avaxXPNetwork)` here?

```
return readOnlySigner.baseTX({
      utxoSet,
      chain: 'X',
      toAddress: destinationAddress,
      amountsPerAsset: {
        [avaxXPNetwork.isTestnet
          ? TESTNET_AVAX_ASSET_ID
          : MAINNET_AVAX_ASSET_ID]: amountInNAvax
      },
      options: {
        changeAddresses: [changeAddress]
      }
    })
  }
```

onghwan (approved):
LGTM

PR #2137, opened 2024-11-27
CP-9586: Fix not able to stake on mainnet and testnet

atn4z7:
## Description

**Ticket: [CP-9586]** 

This pr follows the same solution in https://github.com/ava-labs/core-mobile/pull/2127. We can't rely on our simulated utxos to calculate correct fees. The fees will always be smaller than the actual needed fees. Hence, we will just grab whatever missing amount from the error message when submitting the delegation and deduct that from the total staking amount and retry. 

To prevent the case where the missing amount is super large, I added `p-fee-adjustment-threshold` feature flag, which is set to 0.1% currently. When the missing amount goes over 0.1%, we will reject the transaction. From my testing, the ratio seems to be less than 0.005%. 

`p-fee-adjustment-threshold`
```This flag defines the maximum allowable ratio of the additional fee adjustment (missingAmount) to the total transaction amount (amountAvailable). It is used to ensure that fee adjustments (for claiming rewards and delegation within Core Mobile) remain reasonable and prevent excessive fees from being deducted from the transaction. For example, setting this threshold to 1e-3 (scientific notation for 0.001) means the app will reject a transaction if the required fee adjustment exceeds 0.1% of the total transaction amount. This safeguard protects users from incurring disproportionately high fees.```

## Screenshots/Videos
with small amount on P-Chain
https://github.com/user-attachments/assets/7c239daa-3572-4c6e-bca4-faf664712db0

with medium amount on P-Chain
https://github.com/user-attachments/assets/8514bb5b-a145-4edf-8e5d-b38b5c5caac8

## Testing
Please make sure staking still works for both mainnet and testnet

## Checklist

[CP-9586]: https://ava-labs.atlassian.net/browse/CP-9586?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core-mobile/app/screens/earn/components/Balance.tsx
+++ b/packages/core-mobile/app/screens/earn/components/Balance.tsx
@@ -220,7 +220,9 @@ export const Balance = (): JSX.Element | null => {
             <BalanceItem
               balanceType={StakeTypeEnum.Claimable}
               iconColor={getStakePrimaryColor(StakeTypeEnum.Claimable, theme)}
-              balance={claimableInAvax?.toDisplay() ?? UNKNOWN_AMOUNT}
+              balance={
+                claimableInAvax?.toDisplay({ fixedDp: 4 }) ?? UNKNOWN_AMOUNT
+              }
               poppableItem={
                 [
                   RecoveryEvents.ImportPStart,
@@ -232,7 +234,7 @@ export const Balance = (): JSX.Element | null => {
         </Row>
       </View>
       <View>
-        {claimableInAvax?.gt(0)
+        {claimableInAvax?.gt(0.05)
           ? renderStakeAndClaimButton()
           : renderStakeButton()}
       </View>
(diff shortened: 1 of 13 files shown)

### CODE REVIEW

File: packages/core-mobile/app/screens/earn/components/Balance.tsx [hunk 1]
```diff
@@ -232,7 +234,7 @@ export const Balance = (): JSX.Element | null => {
         </Row>
       </View>
       <View>
-        {claimableInAvax?.gt(0)
+        {claimableInAvax?.gt(0.05)
```
atn4z7, line 237: the balance can be 0.000x something and we can't really claim balance in that case. adjusted the threshold here to not show the claim button in such case.

packages/core-mobile/app/services/earn/exportC.ts
```diff
@@ -53,15 +52,13 @@ export async function exportC({
 
   const cChainBalanceAvax = AvaxC.fromWei(cChainBalanceWei)
   const requiredAmountAvax = AvaxC.fromWei(requiredAmountWei)
-  const pChainFeeAvax = await calculatePChainFee(avaxXPNetwork)
-  const amountAvax = requiredAmountAvax.add(pChainFeeAvax)
```
atn4z7, line 55: pointed out by @neven-s. we don't need to include the p static fee here since we already include it in the required amount.

neven-s (approved):
Logic seems right!

PR #2142, opened 2024-12-05
QA fix staking and rpc call testing

eunjisong:
## Description
- We tested all the possible duration stake testing, but it failed due to the existing bug and always became flaky. On this PR, I randomly chose one duration, and we'll only test one stake on testnet. It could fail again if it chooses a different fee other than `normal`, but it will be addressed soon.
- This PR also contains other small flakiness around the testing suite

## Checklist


Any2suited66 (approved):
I like this.  Makes the tests much more resilient

PR #2148, opened 2024-12-10
CP-9601: Add Card component

ruijialin-avalabs:
## Description

**Ticket: [CP-9601]** 

- Add Card

## Screenshots/Videos

https://github.com/user-attachments/assets/90a59409-034d-4a51-b586-ec6644eaacc4

## Checklist


packages/k2-alpine/src/components/Card/Card.tsx
```diff
@@ -0,0 +1,43 @@
+import React, { PropsWithChildren } from 'react'
+import { SxProp } from 'dripsy'
+import { View } from '../Primitives'
+import { useTheme } from '../..'
+
+export const Card = ({
+  sx,
+  children,
+  shouldCenterAlign = false,
```
atn4z7, line 9: I think centering should be the default behavior. no need to provide a prop for is. the sx prop should be good enough.

packages/k2-alpine/src/components/Card/Card.tsx
```diff
@@ -0,0 +1,43 @@
...
+  backgroundColorOverride
```
atn4z7, line 10: I don't think we should expose this background color prop. especially when it has to do with light and dark theme. that should be an internal implementation detail only.

packages/k2-alpine/src/theme/theme.ts
```diff
@@ -9,7 +9,7 @@ export const darkTheme = makeTheme({
   },
   colors: darkModeColors,
   text,
-  isDark: true as boolean
+  isDark: true
```
atn4z7: this will fail tsc if I remember correctly

packages/k2-alpine/src/components/Card/Card.tsx
ruijialin-avalabs: I also see some card with different color, let me reach out to design about it too.

packages/k2-alpine/src/components/Card/Card.tsx
```diff
@@ -0,0 +1,43 @@
...
+} & PropsWithChildren): React.JSX.Element => {
+  const { theme } = useTheme()
+
+  const getBackgroundColor = (): string => {
+    if (backgroundColorOverride) {
+      return theme.isDark
+        ? backgroundColorOverride.dark
+        : backgroundColorOverride?.light
+    }
+    return theme.isDark ? '#3B3B3D' : '#F2F2F3'
```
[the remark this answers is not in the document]
ruijialin-avalabs: yeah, I was going to reach out to design to confirm that.

packages/k2-alpine/src/theme/theme.ts
ruijialin-avalabs: reverted.

PR #2154, opened 2024-12-12
Set max delegation fee to 20% for simple flow

ruijialin-avalabs:
## Description

- Set max delegation fee to 20% for simple flow

## Checklist

### DIFF
--- a/packages/core-mobile/app/hooks/earn/useSearchNode.ts
+++ b/packages/core-mobile/app/hooks/earn/useSearchNode.ts
@@ -42,8 +42,10 @@ export const useSearchNode = ({
       stakingAmount,
       stakingEndTime,
       minUpTime: 98,
+      maxFee: 20,
       isEndTimeOverOneYear
     })
+
     if (filteredValidators.length === 0) {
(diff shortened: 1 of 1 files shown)


atn4z7 (approved):
Looks good to me. If you could add some tests as well tomorrow, that will be great.

packages/core-mobile/app/services/earn/utils.ts
```diff
@@ -262,9 +262,23 @@ export const getRandomValidator = (
     // get the first item in the array sorted by end time
     return validators.at(0) as NodeValidator
   }
-  const endIndex = validators.length >= 5 ? 4 : validators.length - 1
+
+  // get the validators with lowest delegation fee
+  const validatorsWithLowestFee = validators.filter(
+    validator => validator.delegationFee === validators[0]?.delegationFee
```
neven-s, line 268: Can you please update docs of getRandomValidator function parameter "validators", to something like "input sorted by delegationFee to take random item form"?
At first glance it's not obvious why comparing with first item.

neven-s (approved):
Looks good! 👏

PR #2155, opened 2024-12-12
CP-9398: Sign up and Sign back in Flow

onghwan:
## Description

**Ticket: [CP-9398]** 

* implement signup and pin/bio login flow

### Signup Flow

https://github.com/user-attachments/assets/89d3e199-168a-40d5-87ab-e24e0d29a4d0

### Pin/Bio login flow

https://github.com/user-attachments/assets/65f428f1-736c-46a8-b790-9fba729e06c9

## Checklist

[CP-9398]: https://ava-labs.atlassian.net/browse/CP-9398?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/core-mobile/android/app/build.gradle
```diff
@@ -144,6 +145,9 @@ dependencies {
     implementation fileTree(dir: "libs", include: ["*.jar"])
     implementation 'com.google.firebase:firebase-messaging:24.0.1'
 
+    implementation 'com.github.bumptech.glide:glide:4.12.0'
+    kapt 'com.github.bumptech.glide:compiler:4.12.0'
+
```
onghwan, line 150: This is part of expo-image configuration.

packages/core-mobile/android/app/src/main/java/com/avaxwallet/CustomGlideModule.kt
```diff
@@ -0,0 +1,7 @@
+package com.avaxwallet
+
+import com.bumptech.glide.annotation.GlideModule
+import com.bumptech.glide.module.AppGlideModule
+
+@GlideModule
+class CustomGlideModule : AppGlideModule() {}
```
onghwan, line 7: If we don't create this module, expo-image fails to load images. ([source](https://stackoverflow.com/questions/49901629/glide-showing-error-failed-to-find-generatedappglidemodule))

packages/core-mobile/app/new/components/navigation/BlurredBackgroundView.tsx
```diff
@@ -1,16 +1,8 @@
...
 const BlurredBackgroundView = (): JSX.Element => {
-  const colorScheme = useColorScheme()
-
-  return (
-    <GlassView
-      style={{ flex: 1 }}
-      glassType={colorScheme === 'dark' ? 'dark' : 'light'}
-    />
-  )
+  return <GlassView style={{ flex: 1 }} />
```
onghwan, line 5: If you set the glassType, it applies a background color to the blur, which causes a color difference with the backgroundView. Therefore, we don’t use glassType for the blurView used as the background for the header and tab bar.

packages/core-mobile/app/new/routes/_layout.tsx
```diff
@@ -7,22 +7,110 @@ import { K2AlpineThemeProvider } from '@avalabs/k2-alpine'
...
+      // Navigate to login modal when walletState is not active
+      router.navigate('/loginWithPinOrBiometry')
+    } else if (walletState === WalletState.ACTIVE) {
+      if (canGoBackToWallet && router.canGoBack()) {
+        router.back()
+      } else {
+        router.replace('/portfolio')
+      }
+    }
+  }, [walletState, router, canGoBackToWallet, navigation])
```
onghwan, line 57: Here, the navigation logic corresponding to walletState is handled.

packages/core-mobile/app/new/routes/index.tsx
```diff
@@ -1,41 +1,22 @@
...
-      <Button type="primary" size="medium" onPress={handleSignIn}>
-        Sign in
-      </Button>
-      <Button type="primary" size="medium" onPress={handleSignUp}>
-        Sign up
-      </Button>
-    </View>
+        position: 'absolute'
+      }}
+    />
```
onghwan, line 18: I’ve set up a blank screen at the top level. The reason is that when navigation occurs due to walletState, there’s a slight delay, causing the top screen to appear very briefly (hardly noticeable on iOS and for less than 500ms on Android). It would be great to eliminate the flicker entirely, but I haven’t found a way to do that yet. In the meantime, showing the blank screen during the flicker felt the most natural.

packages/core-mobile/app/components/Avatar.tsx
```diff
@@ -104,11 +104,7 @@ const AvatarBase: FC<AvatarBaseProps> = ({
       testID="avatar__logo_avatar"
     />
   ) : (
-    <FastImage
-      // TODO: remove this workaround when we have a proper solution
-      // workaround for images not appearing
-      // https://github.com/DylanVann/react-native-fast-image/issues/974
-      fallback={true}
+    <Image
```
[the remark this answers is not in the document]
onghwan: If we can’t fix Expo’s Android performance issues, we’ll somehow need to roll back quite a few things besides `expo-image`, like `expo-router`. Maybe we need to decide whether to quickly address the performance issues or abandon Expo before moving further.

packages/core-mobile/app/new/components/SlideToConfirm.tsx
```diff
@@ -0,0 +1,108 @@
...
+  const { theme } = useTheme()
+  const slideWidth = width - 40 // Space for the slider to move
+  const unlockThreshold = slideWidth - _sliderWidth // Set unlock point near the end
+  const sliderWidth = useRef(new Animated.Value(_sliderWidth)).current
+
+  const textColorAnim = sliderWidth.interpolate({
+    inputRange: [0, unlockThreshold],
+    outputRange: theme.isDark
+      ? [theme.colors.$textPrimary, '#000000']
+      : [theme.colors.$textPrimary, '#FFFFFF'],
```
ruijialin-avalabs, line 25: I wonder if we can just use theme.colors.$surfacePrimary, so that we can simply do 

`[theme.colors.$textPrimary, heme.colors.$surfacePrimary]`

packages/core-mobile/app/new/routes/signup.tsx
```diff
@@ -0,0 +1,199 @@
...
+import { MFA } from 'seedless/types'
+import AppleSignInService from 'services/socialSignIn/apple/AppleSignInService'
+import GoogleSigninService from 'services/socialSignIn/google/GoogleSigninService'
+import { OidcProviders } from 'seedless/consts'
+import { hideLogoModal, showLogoModal } from 'new/components/LogoModal'
+import { router } from 'expo-router'
+import AnalyticsService from 'services/analytics/AnalyticsService'
+import { showSnackbar } from 'new/utils/toast'
+
+export default function Index(): JSX.Element {
```
ruijialin-avalabs, line 20: should we name it SignupScreen instead Index?

packages/core-mobile/app/new/routes/signup.tsx
```diff
@@ -0,0 +1,199 @@
...
+    //   screen: AppNavigation.Onboard.AnalyticsConsent,
+    //   params: {
+    //     nextScreen: AppNavigation.Onboard.CreateWalletStack
+    //   }
+    // })
+    // AnalyticsService.capture('RecoveryPhraseClicked')
+  }
+
+  const handleAccessExistingWallet = (): void => {
+    router.navigate('./accessWallet/')
```
ruijialin-avalabs, line 45: just curious, if there difference between `./accessWallet/` vs `./accessWallet`

packages/core-mobile/app/new/components/SlideToConfirm.tsx
onghwan: fixed.

packages/core-mobile/app/new/routes/signup.tsx
onghwan: There is no difference I think, but I removed the trailing /.

ruijialin-avalabs (approved):
lgtm.  🔥

PR #2156, opened 2024-12-13
Fix double notification on Android

neven-s:
## Description

**Ticket: [CP-9567]** 

This fixes double notification on Android devices when app is in **background** since FirebaseSDK automatically handles FCM with `notification` payload.

## Testing
- make sure that notifications are enabled in system and app settings
- make sure you are receiving any notifications and hear sound 
- use wallet on another device to send some funds on testing device (can be same wallet)
- observe that no double sound is heard or double notifications present

## Checklist

[CP-9567]: https://ava-labs.atlassian.net/browse/CP-9567?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/core-mobile/app/services/notifications/NotificationsService.ts
```diff
@@ -321,16 +321,14 @@ class NotificationsService {
         pressAction: {
           id: PressActionId.OPEN_PORTFOLIO,
           launchActivity: LAUNCH_ACTIVITY
-        }
+        },
+        channelId: channelId ?? ''
```
onghwan, line 325: Is there a difference between passing an empty string and undefined for channelId?

neven-s: yes, in my experience `notifee` will report such payload as invalid

onghwan (approved):

packages/core-mobile/app/services/notifications/NotificationsService.ts
onghwan: it's weird since `channelId` is an optional parameter.

neven-s: it's optional for ios platform, on android it's required

PR #2157, opened 2024-12-13
Qa update version test

Any2suited66:
## Description

- In the past the mobile would crash on initial startup after an update.  This test downloads a previous version of the app, installs it on the device, logs into the app, verifies the portfolio screen is displayed, updates the app with the latest external version, reuses the app state, enters the pin code, and verifies if the user lands on the portfolio screen.  
- If the app crashes the enter pin code step or verification of the portfolio screen will fail and we'll know immediately.

## Checklist


packages/core-mobile/e2e/tests/updateAppVersion/loginWithOlderVersion.sh
```diff
@@ -0,0 +1,18 @@
+set -o pipefail
+if [ "$IS_REGRESSION_RUN" = true ]; then
+
+  npm rebuild detox
+  adb uninstall com.avaxwallet
+  npx ts-node ./e2e/tests/updateAppVersion/getOldAndroidVersion.js
+
+  adb install -r "./e2e/tests/updateAppVersion/oldVersionApk/app-external-e2e-bitrise-signed.apk"
```
eunjisong, line 8: Looks like  `_install-old-android-version`, `_install-latest-android-version` is added to pixel6  and pixel6 is running both internal and external build testing. the internal smoke testing has still access to external build?

eunjisong (approved):

eunjisong (approved):

PR #2161, opened 2024-12-17
CP-9608: Authenticator flow

ruijialin-avalabs:
## Description

**Ticket: [CP-9608]** 

- add Authenticator flow

## Screenshots/Videos

https://github.com/user-attachments/assets/7e01fd01-94d9-492b-97a5-613aa98141b5

dark mode:

https://github.com/user-attachments/assets/ed606942-15b3-4b36-8dae-a4f571db7500

## Checklist

[CP-9608]: https://ava-labs.atlassian.net/browse/CP-9608?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/core-mobile/app/new/components/navigation/BackBarButton.tsx
```diff
@@ -1,19 +1,21 @@
...
 import { Platform } from 'react-native'
+import { router } from 'expo-router'
 
 const BackBarButton = (): JSX.Element => {
   const { theme } = useTheme()
 
   return (
-    <View
+    <Pressable
+      onPress={() => router.back()}
```
ruijialin-avalabs, line 11: this is needed to navigate back correctly in nested stack.

packages/core-mobile/app/new/contexts/SeedlessOidcProvider.tsx
```diff
@@ -0,0 +1,62 @@
...
+    showLogoModal()
+    const walletName = await SeedlessService.getAccountName()
+    hideLogoModal()
+
+    if (withMfa) {
+      showSnackbar('Recovery methods added successfully')
+    }
+
+    if (walletName) {
+      router.navigate('./createPin')
```
ruijialin-avalabs, line 39: todo screen in CP-9604

packages/core-mobile/app/new/contexts/SeedlessOidcProvider.tsx
```diff
@@ -0,0 +1,62 @@
...
+      return
+    }
+    router.navigate('./nameYourWallet')
```
ruijialin-avalabs, line 42: todo screen in CP-9604

packages/core-mobile/app/new/hooks/useSeedlessManageMFA.ts
```diff
@@ -0,0 +1,50 @@
+import { TotpChallenge } from '@cubist-labs/cubesigner-sdk'
+import { showSnackbar } from 'new/utils/toast'
+import useVerifyMFA from 'seedless/hooks/useVerifyMFA'
+import SeedlessService from 'seedless/services/SeedlessService'
+import Logger from 'utils/Logger'
+
+function useSeedlessManageMFA(): {
```
ruijialin-avalabs, line 7: create the same hook here with new toast component, existing useSeedlessManageMFA uses old snackbar component.

packages/core-mobile/app/new/routes/addRecoveryMethods.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+  const [selectedMethod, setSelectedMethod] = useState(
+    availableRecoveryMethods.length > 0
+      ? availableRecoveryMethods[0]?.type
+      : undefined
+  )
+
+  const handleOnPress = (): void => {
+    if (selectedMethod === RecoveryMethods.Passkey) {
+      // navigate({
```
ruijialin-avalabs, line 27: todo screen in CP-9607

packages/core-mobile/app/new/routes/addRecoveryMethods.tsx
```diff
@@ -0,0 +1,98 @@
...
+      //     title: 'How would you like to name your passkey?',
+      //     description: 'Add a Passkey name, so it’s easier to find later',
+      //     textInputPlaceholder: 'Passkey name',
+      //     fidoType: FidoType.PASS_KEY
+      //   }
+      // })
+      return
+    }
+    if (selectedMethod === RecoveryMethods.Yubikey) {
+      // navigate({
```
ruijialin-avalabs, line 39: todo screen in CP-9609

packages/core-mobile/app/new/contexts/TotpProvider.tsx
```diff
@@ -0,0 +1,132 @@
...
+import useSeedlessManageMFA from 'new/hooks/useSeedlessManageMFA'
+import { useSeedlessOidcContext } from './SeedlessOidcProvider'
+
+export interface TotpContextState {
+  handleCopyCode: () => void
+  onVerifyCode: (code: string) => Promise<Result<undefined, TotpErrors>>
+  onVerifySuccess: () => void
+  goToEnterCodeManually: () => void
+  goToScanQrCode: () => void
+  goToVerifyCode: () => void
```
onghwan, line 29: can we move these `goToXXX` functions into screens?

packages/core-mobile/app/new/contexts/SignupProvider.tsx
```diff
@@ -0,0 +1,141 @@
...
+    router.navigate('./copyCode')
+  }
+
+  const goToScanQrCode = (): void => {
+    router.navigate('./scanQrCode')
+  }
+
+  const goToVerifyCode = (): void => {
+    router.push('./verifyCode')
+  }
```
onghwan, line 77: I think we can have these routing functions in the screens, even if they might be duplicated in some of them.(they are just one line function call) By keeping the routing function calls only within the screens, I believe that when you look at a screen’s code, you can easily see which other screens it’s connected to at a glance.

packages/core-mobile/app/new/routes/(authenticator)/components/AuthenticatorSetup.tsx
```diff
@@ -0,0 +1,143 @@
...
+        </View>
+        <View sx={{ marginBottom: 36 }}>
+          <Button type="primary" size="large" onPress={onVerifyCode}>
+            Next
+          </Button>
+        </View>
+      </View>
+    </BlurredBarsContentLayout>
+  )
+}
```
onghwan, line 143: can we move these components under `(authenticator)/components` to `app/new/components`?

packages/core-mobile/app/new/routes/(authenticator)/totp/_layout.tsx
```diff
@@ -0,0 +1,16 @@
...
+export default function TotpLayout(): JSX.Element {
+  return (
+    <SignupProvider>
+      <Stack>
+        <Stack.Screen name="authenticatorSetup" />
+        <Stack.Screen name="scanQrCode" />
+        <Stack.Screen name="copyCode" />
+        <Stack.Screen name="verifyCode" />
+      </Stack>
+    </SignupProvider>
```
onghwan, line 14: Although it works well as it is now, I think `TotpLayout` should be declared in a place like the root `_layout.tsx`. Also, `signup.tsx` and `totp/_layout.tsx` should be wrapped with a single `<SignupProvider />`.

ruijialin-avalabs: I was actually trying to wrap `signup.tsx` and other totp screens in single `<SignupProvider />`.  But didn't work out well.  I will move totp to root, to use single  `<SignupProvider />` at root.

ruijialin-avalabs: moved to `totp` to root.

packages/core-mobile/app/new/contexts/SignupProvider.tsx
```diff
@@ -0,0 +1,117 @@
...
+  }, [router])
+
+  const onVerifySuccess = useCallback((): void => {
+    router.dismissAll()
+    router.back()
+    handleAccountVerified()
+    AnalyticsService.capture('SeedlessMfaVerified', {
+      type: 'Authenticator'
+    })
+  }, [router, handleAccountVerified])
```
onghwan, line 95: I think it would be good to also move the handleAccountVerified function and onVerifySuccess to the screen.

packages/core-mobile/app/new/contexts/SignupProvider.tsx
```diff
@@ -0,0 +1,117 @@
...
+  handleAccountVerified: () => Promise<void>
+  handleCopyCode: () => void
+  onVerifyCode: (code: string) => Promise<Result<undefined, TotpErrors>>
+  onVerifySuccess: () => void
+  totpKey?: string
+  totpChallenge?: TotpChallenge
+  setTotpChallenge: Dispatch<SetStateAction<TotpChallenge | undefined>>
+  oidcAuth?: OidcAuth
+  setOidcAuth: Dispatch<SetStateAction<OidcAuth | undefined>>
+  allowsUserToAddLater: boolean
```
onghwan, line 31: When navigating with expo-router, we can’t pass callbacks, but we can pass variables like booleans. So I think we don’t need to manage `allowsUserToAddLater` with context.

ruijialin-avalabs: Screenshot 2024-12-18 at 3 41 32 PM

I think searchParams only allows string or string[]

packages/core-mobile/app/new/contexts/SignupProvider.tsx
ruijialin-avalabs: handleAccountVerified is used in multiple place during the signup flows, I think it is better to just leave it in the context, since this context is very much for signup flow anyway.

onghwan: While that’s true in terms of code reusability, I think it’s more consistent to keep showLogoModal()/hideLogoModal() and routing on the screen, even if it means some duplication. Also, at the very least, I believe onVerifySuccess should be moved to the screen. The reason is that router.dismissAll() and router.back() are functions that should only be called within a specific navigation state. In other situations, different dismiss logic would be necessary.

packages/core-mobile/app/new/contexts/SignupProvider.tsx
onghwan: You’re right. However, wouldn’t it be better to pass it as a parameter, even after converting the string to a boolean?

ruijialin-avalabs: moved `allowsUserToAddLater` to search params.

onghwan (approved):
LGTM

PR #2165, opened 2024-12-18
CP-9653: Prevent rendering charts with insufficient data points

onghwan:
## Description

**Ticket: [CP-9653]** 

* Fixed a crash that occurs when rendering a chart for a token with only one data point.
* Fixed an issue where `react-native-graph` was rendering too small and the layout wasn’t adjusting on iOS.

## Testing
* For tokens like `DUA`, `BSGG`, `OH`, and `RACEX`, if chart data is not available on the Details page, we cannot render the graph. It appears that these tokens also lack chart data on CoinGecko.
* In the case of `USUAL` tokens, the chart is visible on the Details page, but in the Watchlist screen, they only have a single data point, which seems to be an issue with CoinGecko. Rendering the graph in such cases would cause a crash, so we modified it to prevent the graph from being drawn.
* In all other cases, the graph should be rendered in the Watchlist.


neven-s (approved):
neat

atn4z7 (approved):

PR #2178, opened 2025-01-07
CP-9659 Notifications update

neven-s:
## Description

**Ticket: [CP-9659]** 

This PR tackles couple issues:
- unoptimised call to register endpoint by caching deviceArn 
- unhandled background notifications on Android by processing silent (data-only) notifications

### unoptimised call to register endpoint by caching deviceArn 
- when we get `deviceArn` from backend we store it in `commonStorage` under `NOTIFICATIONS_OPTIMIZATION`
- no need to clear it since this stays same for device most of the time, and if it changes we'll get new value from backend

### unhandled background notifications on Android by processing silent (data-only) notifications
- we now pass `osType` to `register` endpoint so backend can send different FCM payload for android and ios
- `FCMService` now processes differently these types of payloads
- for silent notifications we decide android's channel id with `EVENT_TO_CH_ID` map

### Receive Balanace notification on foreground

https://github.com/user-attachments/assets/d3a3adca-ea60-4e34-8457-85298932201b

### Receive Balance notification on background

https://github.com/user-attachments/assets/11db1f70-30bf-4afc-94a0-d872aa64f6b7

### Spent Balance notification on background

https://github.com/user-attachments/assets/48bfe1cf-ac42-4694-9908-bb13f9690ce6

## Checklist

[CP-9659]: https://ava-labs.atlassian.net/browse/CP-9659?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/core-mobile/app/services/fcm/FCMService.ts
```diff
@@ -60,32 +65,70 @@ class FCMService {
...
+      await NotificationsService.displayNotification(notificationData).catch(
+        Logger.error
+      )
     })
   }
+
+  #prepareDataOnlyNotificationData = (
+    fcmData: NotificationsBalanceChangeData
+  ): DisplayNotificationParams => {
+    if (!fcmData.title) throw Error('No notification title')
```
atn4z7, line 94: will this throwing crash the app? not sure if we have a catch for it somewhere

packages/core-mobile/app/services/fcm/FCMService.ts
```diff
@@ -60,32 +65,70 @@ class FCMService {
...
+    }
+  }
+
+  #prepareNotificationData = (
+    fcm: NotificationsBalanceChange
+  ): DisplayNotificationParams => {
+    if (!fcm.notification) throw Error('No notification payload')
+    const data = this.#extractDeepLinkData(fcm.data)
+    return {
+      channelId: fcm.notification.android?.channelId,
```
atn4z7, line 110: where does `fcm.notification.android?.channelId` come from? I thought backend will not include channelId anymore?

packages/core-mobile/app/services/fcm/FCMService.ts
neven-s: this is swallowed by messaging sdk, but yeah, could be better
How about Logging it then throw?

packages/core-mobile/app/services/fcm/FCMService.ts
neven-s: this is for backward compatibility, not sure if it's guaranteed that backend code will be pushed before ours. And if they rollback we are safe with this.
Should i add comment to clear it sometime in future?

packages/core-mobile/app/services/fcm/FCMService.ts
atn4z7: it's fine then

packages/core-mobile/app/services/fcm/FCMService.ts
atn4z7: gotcha. makes sense. and yeah a comment will be good.

neven-s: [Add comment](https://github.com/ava-labs/core-mobile/pull/2178/commits/4b88ba567f7ed54b4c32a1fbcdddf66ea13d1b37)

atn4z7 (approved):
nice optimization!

PR #2201, opened 2025-01-28
CP-9701: Square Button

onghwan:
## Description

**Ticket: [CP-9701]** 

* implement SquareButton component
* fix circular dependency issues


ruijialin-avalabs (approved):
Nice!

PR #2249, opened 2025-02-07
CP-9880: add a buffer to all fees for cross chain amount

atn4z7:
## Description

**Ticket: [CP-9880]** 

Looks like we will occasionally run into situations where we are missing 10000-100000 nAVAX (0.0001 AVAX) when trying to do a delegation that involves a cross chain transfer from C-Chain. This pr adds a buffer (currently configured at 400% via the `cross-chain-fees-multiplier` feature flag) to all the fees, which are then added to the final amount to transfer. This seems to work okay since the actual fees are usually way lesser. The only drawback is users will have a little bit AVAX remaining on P-Chain after submitting the delegation.  

## Screenshots/Videos
https://github.com/user-attachments/assets/8146bed0-1fa3-4a8b-93f7-df91e3e780e0

## Testing
Please re-test all the different staking scenarios

## Checklist

[CP-9880]: https://ava-labs.atlassian.net/browse/CP-9880?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core-mobile/app/hooks/earn/useDelegation.ts
+++ b/packages/core-mobile/app/hooks/earn/useDelegation.ts
@@ -7,7 +7,7 @@ import { useCChainBaseFee } from 'hooks/useCChainBaseFee'
 import { selectActiveNetwork } from 'store/network/slice'
 import {
   selectPFeeAdjustmentThreshold,
-  selectPFeeMultiplier,
+  selectCrossChainFeesMultiplier,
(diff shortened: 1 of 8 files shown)


ruijialin-avalabs (approved):
LGTM

PR #2267, opened 2025-02-12
CP-9745: CubeSigner SDK upgrade

onghwan:
## Description

**Ticket: [CP-9745]** 

upgrade cubesigner sdk to the latest `0.4.117-0`
https://signer-docs.cubist.dev/ts-sdk/migration

## Testing
I did test the following but please test again:
- signing up with/without MFA
- logging in
- adding accounts
- exporting mnemonic phrase
- signing message/transaction
- upgrading from existing production version

## Checklist

[CP-9745]: https://ava-labs.atlassian.net/browse/CP-9745?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/core-mobile/app/seedless/errors.ts
```diff
@@ -3,9 +3,6 @@ import { ErrorBase } from 'errors/ErrorBase'
 type TotpErrorName = 'RequiresMfa' | 'WrongMfaCode' | 'UnexpectedError'
 export class TotpErrors extends ErrorBase<TotpErrorName> {}
 
-type TokenRefreshErrorName = 'RefreshFailed' | 'TokenExpired'
-export class TokenRefreshErrors extends ErrorBase<TokenRefreshErrorName> {}
```
atn4z7, line 7: removed as we don't need it

packages/core-mobile/app/seedless/services/SeedlessExportService.ts
```diff
@@ -1,19 +1,21 @@
...
+import SeedlessSession from './SeedlessSession'
 
 class SeedlessExportService {
-  sessionManager = new SeedlessSessionManager({
+  session = new SeedlessSession({
     scopes: ['export:user', 'manage:export:user', 'manage:mfa:vote'],
-    sessionStorage: new MemorySessionStorage<SignerSessionData>()
+    // passing null as session data initially
+    // user will perform OIDC auth when the export flow starts
+    sessionManager: new MemorySessionManager(null as unknown as SessionData)
```
atn4z7, line 18: the old MemorySessionStorage doesn't require passing session data to its constructor but the new one MemorySessionManager does. doing null here since we want to separate the 2 sessions (app session and export session).

packages/core-mobile/app/seedless/services/SeedlessSession.ts
```diff
@@ -121,101 +116,68 @@ class SeedlessSessionManager {
...
+      const client = await response.mfaClient()
+      if (!client) throw new Error('MFA client is missing')
+      this.signerClient = client
+    } else {
+      // when MFA is not required or MFA verification is complete
+      // we can safely access session data
+      const sessionData = response.data()
+
+      // destroy outdated client instance
+      this.signerClient = null
```
atn4z7, line 131: here we simply "destroy" the current client. next time the app needs a client, it will construct it based on the latest session data that we just save here.

packages/core-mobile/app/seedless/services/SeedlessSession.ts
```diff
@@ -121,101 +116,68 @@ class SeedlessSessionManager {
...
       return {
         success: false,
-        error: {
-          name: 'RefreshFailed',
-          message: 'Failed to load session manager from storage'
-        }
+        error: err as Error
       }
     }
-    const refreshResult = await retry({
```
atn4z7, line 159: also we don't need to do retry since internally cubist already does retry for 500 http errors.

packages/core-mobile/app/seedless/services/SeedlessSession.ts
```diff
@@ -302,31 +268,35 @@ class SeedlessSessionManager {
...
+        }
+
+        this.signerClient = client
+        this.signerClientPromise = null // Reset the promise lock
+        return client
+      })()
+    }
+
+    // return a promise to prevent multiple client creation
+    return this.signerClientPromise
```
atn4z7, line 299: I noticed we are accessing this method getSignerSession() at the same time in some places. just doing a promise lock so that we only create 1 client.

packages/core-mobile/app/seedless/store/listeners.ts
```diff
@@ -20,60 +20,42 @@ import { WalletType } from 'services/wallet/types'
...
-    }
-  }
-  GlobalEvents.onError(onErrorHandler)
+  const walletType = selectWalletType(getState())
+
+  if (walletType === WalletType.MNEMONIC) return
+
+  SeedlessService.init({
+    onSessionExpired: () => dispatch(onTokenExpired)
+  })
```
atn4z7, line 58: we can no longer listen to the token expired event globally. in the latest sdk, whenever a cubesigner client is created, we need to use it to listen to the token expired event. thus, on an app open, we are now just initializing seedless service with the logic to handle an expired session. whenever a cubesigner client is created, it will start listening to the `token-expired` event and call this "onSessionExpired" logic accordingly.

packages/core-mobile/app/seedless/services/wallet/SeedlessWallet.ts
```diff
@@ -286,7 +286,9 @@ export default class SeedlessWallet implements Wallet {
     provider: JsonRpcBatchInternal
   }): Promise<string> {
     const signer = new CsEthersSigner(
-      getEvmAddressFromPubKey(this.getPubKeyBufferC()),
+      // need to use lowercase address due to a bug in the cubist sdk
+      // TODO remove toLowerCase when the bug is fixed
```
ruijialin-avalabs, line 290: can we link to the bug in cubist sdk?

atn4z7: I have no idea actually. it's somewhere in their source code. they will probably fix it in the next sdk version.

packages/core-mobile/app/seedless/services/SeedlessSession.ts
```diff
@@ -121,101 +116,68 @@ class SeedlessSessionManager {
...
-        SEEDLESS_ORG_ID,
-        sessionResponse,
-        this.sessionStorage
-      )
+    if (response.requiresMfa()) {
+      // if MFA is required, we cannot access session data until MFA verification is complete
+      // thus, we are just storing the MFA client here (for verification operations)
+      const client = await response.mfaClient()
+      if (!client) throw new Error('MFA client is missing')
+      this.signerClient = client
```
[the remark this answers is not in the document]
ruijialin-avalabs: should we store this to a separate variable as mfaClient instead?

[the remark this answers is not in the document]
atn4z7: it's the same instance/class so I don't think it's necessary

PR #2294, opened 2025-02-19
fix swapping AVAX/wAVAX

atn4z7:
## Description

from @meeh0w 

```
my guess is something must've changed on Paraswap side, because previously when we would fetch swap rates for AVAX, we could pass srcToken=AVAX as parameter, now we need to change it to srcToken=0xeee(...) (all Es).
I'm basing my guess solely off of the fact that it's also broken for Mobile atm. and as far as I can tell, this portion of code was changed ~10months ago for mobile.
```
same fix as https://github.com/ava-labs/core-extension/pull/149

## Screenshots/Videos
**before**
https://github.com/user-attachments/assets/d013466e-b836-465e-9c1c-c8fa454fd85a

**after**
https://github.com/user-attachments/assets/b4a2dd64-6f39-4b56-b918-6b2a1e0f999c

## Testing
Please test swapping AVAX/wAVAX

## Checklist

### DIFF
--- a/packages/core-mobile/app/services/swap/SwapService.ts
+++ b/packages/core-mobile/app/services/swap/SwapService.ts
@@ -125,10 +125,13 @@ class SwapService {
           throw new Error('Account address missing')
         }
 
+        const isFromTokenNative = network.networkToken.symbol === srcToken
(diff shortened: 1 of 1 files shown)


onghwan (approved):
🫡

PR #2373, opened 2025-03-11
Fix price change indicator performance issue

onghwan:
## Description

* use plain color for text.
* use animated text only when animated param is true


ruijialin-avalabs (approved):
🙌

PR #2375, opened 2025-03-11
Fix hot reload

onghwan:
## Description

* upgrade `@react-native/babel-preset` and `@react-native/metro-config` to match the current react native version


ruijialin-avalabs (approved):
❤️

PR #2376, opened 2025-03-11
Improve json rpc error parsing

atn4z7:
## Testing
please re-test send and swap and make sure we are handing all the error cases correctly

## Checklist


packages/core-mobile/app/errors/swapError.ts
```diff
@@ -84,18 +84,15 @@ export function humanizeSwapError(err: unknown): string {
...
   if (errorString.toLowerCase().includes('network error')) {
     return 'Network error, please try again later.'
   }
 
-  if (err instanceof JsonRpcError) {
-    if (isError(err.cause, 'INSUFFICIENT_FUNDS'))
-      return 'Insufficient amount for gas. Reduce swap quantity and try again.'
-
-    return err.message
+  if (err instanceof JsonRpcError && err.data.cause instanceof JsonRpcError) {
```
ruijialin-avalabs, line 94: should this be `err instanceof JsonRpcError || err.data.cause instanceof JsonRpcError`

ruijialin-avalabs (approved):

packages/core-mobile/app/errors/swapError.ts
atn4z7: no, we are wrapping original json rpc error inside another json rpc error so we need to check both of them

PR #2401, opened 2025-03-14
QA New Gen - automate onboarding tests

eunjisong:
## Description
- **Update detox version to 20.34.5**
- method/loc files cleanup: the new gen has a new title and contents and the old file names don't make sense with the new design and aren't very clear which screen its referring. I consolidated all into onboardingPage so the methods are not spread out and it would be easier to maintain without duplicates. 
- tests: onboard existing wallet 
- test: onboard new wallet 
- test: onboard metamask wallet 
- test: update and clean warmup method 

## Checklist


packages/core-mobile/app/new/routes/signup.tsx
```diff
@@ -84,7 +84,7 @@ export default function Signup(): JSX.Element {
           type="tertiary"
           size="large"
           onPress={handleAccessExistingWallet}>
-          Access existing Wallet
+          Access existing wallet
```
eunjisong, line 87: Confirmed with Michael that we use sentence case throughout in the new design

Any2suited66 (dismissed):
Looks good.  Our first two tests in the new UI.  Nice work.

Any2suited66 (approved):

PR #2402, opened 2025-03-15
Fix auto-focus issue on login screen 

onghwan:
## Description

* Fixed an issue where the PinInput on the login screen would not automatically focus.
* Fixed an issue where the modal screen wouldn’t close after unlocking the app.


ruijialin-avalabs (approved):
👍

PR #2409, opened 2025-03-17
[Perf Improvement] Dropdown and Segmented Control

atn4z7:
## Description

various optimizations are added for screens with dropdown and segmented control. please refer to the comments/code.

## Screenshots/Videos
https://github.com/user-attachments/assets/d516b218-2a63-454d-b697-ec9776f65a1d

## Checklist

### DIFF
--- a/packages/core-mobile/app/hooks/networks/useNetworksListener.ts
+++ b/packages/core-mobile/app/hooks/networks/useNetworksListener.ts
@@ -3,7 +3,6 @@ import { ReactQueryKeys } from 'consts/reactQueryKeys'
 import { queryClient } from 'contexts/ReactQueryProvider'
 import { useEffect } from 'react'
 import { useDispatch } from 'react-redux'
-import { onAppUnlocked } from 'store/app'
 import { onNetworksFetched } from 'store/network'
 import { toggleDeveloperMode } from 'store/settings/advanced'
 
@@ -14,7 +13,7 @@ export const useNetworksListener = (): void => {
   useEffect(() => {
     return dispatch(
       addListener({
-        matcher: isAnyOf(toggleDeveloperMode, onAppUnlocked),
+        matcher: isAnyOf(toggleDeveloperMode),
         effect: async () => {
           await queryClient.invalidateQueries({
             queryKey: [ReactQueryKeys.NETWORKS]
--- a/packages/core-mobile/app/new/common/components/CollapsibleTabs.tsx
+++ b/packages/core-mobile/app/new/common/components/CollapsibleTabs.tsx
@@ -1,16 +1,12 @@
-import React, { forwardRef } from 'react'
+import React, { forwardRef, useMemo } from 'react'
 import {
   CollapsibleRef,
   TabBarProps,
   Tabs,
-  useAnimatedTabIndex,
   useCurrentTabScrollY
 } from 'react-native-collapsible-tab-view'
-import {
-  runOnJS,
-  useAnimatedReaction,
-  SharedValue
-} from 'react-native-reanimated'
+import { runOnJS, useAnimatedReaction } from 'react-native-reanimated'
+import { StyleSheet } from 'react-native'
 
 export const CollapsibleTabsContainer = forwardRef<
   CollapsibleRef,
@@ -18,37 +14,39 @@ export const CollapsibleTabsContainer = forwardRef<
     renderHeader: () => JSX.Element
     renderTabBar: (props: TabBarProps) => JSX.Element
     onIndexChange?: (index: number) => void
-    onScrollTab?: (tabIndex: SharedValue<number>) => void
     onScrollY?: (contentOffsetY: number) => void
     tabs: { tabName: string; component: JSX.Element }[]
   }
 >(
   (
-    { tabs, renderHeader, renderTabBar, onIndexChange, onScrollTab, onScrollY },
+    { tabs, renderHeader, renderTabBar, onIndexChange, onScrollY },
     ref
   ): JSX.Element => {
+    const content = useMemo(() => {
+      return tabs.map(tab => (
+        <Tabs.Tab key={tab.tabName} name={tab.tabName}>
+          <CollapsibleTabWrapper
+            component={tab.component}
+            onScrollY={onScrollY}
+          />
+        </Tabs.Tab>
+      ))
+    }, [onScrollY, tabs])
+
+    const pagerProps = useMemo(() => {
+      return { style: styles.overflow }
+    }, [])
+
     return (
       <Tabs.Container
         ref={ref}
-        headerContainerStyle={{
-          shadowOpacity: 0,
-          elevation: 0,
-          overflow: 'visible'
-        }}
-        containerStyle={{ overflow: 'visible' }}
+        headerContainerStyle={styles.tabsContainer}
+        containerStyle={styles.overflow}
         renderHeader={renderHeader}
         renderTabBar={renderTabBar}
-        pagerProps={{ style: { overflow: 'visible' } }}
+        pagerProps={pagerProps}
         onIndexChange={onIndexChange}>
-        {tabs.map(tab => (
-          <Tabs.Tab key={tab.tabName} name={tab.tabName}>
-            <CollapsibleTabWrapper
-              component={tab.component}
-              onScrollTab={onScrollTab}
-              onScrollY={onScrollY}
-            />
-          </Tabs.Tab>
-        ))}
+        {content}
       </Tabs.Container>
     )
   }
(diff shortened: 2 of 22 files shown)


packages/core-mobile/app/hooks/networks/useNetworksListener.ts
```diff
@@ -14,7 +13,7 @@ export const useNetworksListener = (): void => {
   useEffect(() => {
     return dispatch(
       addListener({
-        matcher: isAnyOf(toggleDeveloperMode, onAppUnlocked),
+        matcher: isAnyOf(toggleDeveloperMode),
```
atn4z7, line 16: this was dispatching a networksFetched action on app unlocked that triggered a balance refetch. as a result we were fetching balance twice on app unlocked, which took a lot of the js thread.

packages/core-mobile/app/new/common/components/CollapsibleTabs.tsx
```diff
@@ -75,15 +77,6 @@ const CollapsibleTabWrapper = ({
...
 
-  useAnimatedReaction(
-    () => tabIndex.value,
-    (curr, prev) => {
-      if (curr !== prev && onScrollTab) {
-        runOnJS(onScrollTab)(tabIndex)
-      }
-    }
-  )
-
```
atn4z7, line 86: @onghwan this doesn't seem to be needed so I removed it. it was doing a lot of work on the js thread.

packages/core-mobile/app/new/features/portfolio/assets/components/AssetsScreen.tsx
```diff
@@ -128,29 +129,31 @@ const AssetsScreen: FC<Props> = ({
     )
   }, [isBalanceLoading, isRefetchingBalance, refetch, isAllBalancesInaccurate])
 
+  const dataLength = data.length
+
   const header = useMemo(() => {
+    if (dataLength === 0) return
+
```
atn4z7, line 136: avoid rendering header if there is no data

packages/core-mobile/app/new/features/portfolio/assets/hooks/useAssetsFilterAndSort.ts
```diff
@@ -106,27 +106,45 @@ export const useAssetsFilterAndSort = (): {
...
+
+  return useMemo(
+    () => ({
+      filter,
+      sort,
+      view,
+      data: filteredAndSorted,
+      refetch
+    }),
+    [filter, sort, view, filteredAndSorted, refetch]
```
atn4z7, line 148: make sure we don't return new object every single time

packages/core-mobile/app/new/routes/(signedIn)/(tabs)/portfolio/index.tsx
```diff
@@ -106,39 +118,53 @@ const PortfolioHomeScreen = (): JSX.Element => {
...
       <NavigationTitleHeader
         title={activeAccount?.name ?? ''}
         subtitle={formattedBalance}
       />
     ),
+    [activeAccount?.name, formattedBalance]
+  )
+
+  const { onScroll, targetHiddenProgress } = useFadingHeaderNavigation({
+    header,
```
atn4z7, line 147: make sure we don't recreate header on every render

packages/core-mobile/app/new/routes/(signedIn)/(tabs)/portfolio/index.tsx
```diff
@@ -174,17 +200,42 @@ const PortfolioHomeScreen = (): JSX.Element => {
...
-      tabViewRef.current?.setIndex(index)
-    }
-  }
+      InteractionManager.runAfterInteractions(() => {
+        if (tabViewRef.current?.getCurrentIndex() !== index) {
+          tabViewRef.current?.setIndex(index)
+        }
+      })
+    },
+    [selectedSegmentIndex]
```
atn4z7, line 228: @onghwan I adjusted how CollapsibleTabs.Container and SegmentedControl work together 

**before:**
on segment clicked -> we tell pager view to change index -> when pager view broadcasts the index change event, we set segmentIndex value.
this delays the visual change on segment and also makes it kinda janky since a lot of things are happening in between js thread, native thread and UI thread.

**after:**
on segment clicked -> we set segmentIndex value so the segment is clicked visually right away -> we tell pager view to change index only when app is not too busy (initial app unlock, app is busy making a lot of requests).

packages/core-mobile/app/new/routes/(signedIn)/(tabs)/portfolio/index.tsx
```diff
@@ -174,17 +200,42 @@ const PortfolioHomeScreen = (): JSX.Element => {
...
-  const handleChangeTab = (index: number): void => {
-    selectedSegmentIndex.value = index
-  }
+  const handleTabChange: OnTabChange = useCallback(
+    data => {
+      if (selectedSegmentIndex.value === data.prevIndex) {
+        selectedSegmentIndex.value = data.index
+      }
+    },
+    [selectedSegmentIndex]
```
atn4z7, line 237: @onghwan also adjusted this logic so that we only set selectedSegmentIndex when necessary

packages/k2-alpine/src/components/SegmentedControl/SegmentedControl.tsx
```diff
@@ -28,86 +36,86 @@ export const SegmentedControl = ({
...
-}): JSX.Element => {
+}): JSX.Element | null => {
+  const [isReady, setIsReady] = useState(true)
   const { theme } = useTheme()
-  const [viewWidth, setViewWidth] = useState<number>(0)
-  const [textWidths, setTextWidths] = useState<number[]>(
-    Array.from({ length: items.length }, () => 0)
-  )
-  const textRatios = useMemo(() => {
-    if (!dynamicItemWidth) {
```
atn4z7, line 38: @onghwan I refactored the segmented control component to not rely on useState since these will happen on JS thread and the app is already so busy with the JS thread usually. it now uses shared values instead.

packages/k2-alpine/src/components/SegmentedControl/SegmentedControl.tsx
```diff
@@ -28,86 +36,86 @@ export const SegmentedControl = ({
...
+  const textWidths = useSharedValue(Array(items.length).fill(0))
 
-  const translationAnimation = useDerivedValue(() => {
-    if (viewWidth === 0) return 0
+  useEffect(() => {
+    const task = InteractionManager.runAfterInteractions(() => {
+      setIsReady(true)
+    })
+    return () => task.cancel()
+  }, [])
```
atn4z7, line 51: @onghwan also make it render the segmented control after app is not busy.

packages/k2-alpine/src/components/SegmentedControl/SegmentedControl.tsx
```diff
@@ -163,22 +168,37 @@ const Segment = ({
...
-    onTextWidthChange(event.nativeEvent.layout.width + 64)
-  }
+  const throttledOnPress = useMemo(
+    () =>
+      throttle(onPress, THROTTLE_DURATION, {
+        leading: true,
+        trailing: false
+      }),
+    [onPress]
+  )
```
atn4z7, line 188: added some throttle here to make sure we don't keep clicking it, which results in weird animation/state

packages/k2-alpine/src/components/SegmentedControl/SegmentedControl.tsx
```diff
@@ -137,11 +138,13 @@ export const SegmentedControl = ({
...
                 key={index}
-                ratio={textRatios[index]}
+                ratio={dynamicItemWidth ? textRatios : 1 / items.length}
                 text={item}
                 index={index}
                 selectedIndex={selectedSegmentIndex}
-                onTextWidthChange={width => handleTextWidthChange(index, width)}
+                onTextWidthChange={
+                  dynamicItemWidth ? handleTextWidthChange : undefined
+                }
```
atn4z7, line 147: also, I made it stop running onTextWidthChange on every render. that function measures component layout so it's a bit heavy. now, only when we are dealing with dynamicItemWidth then onTextWidthChange will run. similar for ratio, we are pass a constant number for non dynamic case and a shared value for dynamic case.

packages/core-mobile/app/new/common/components/CollapsibleTabs.tsx
```diff
@@ -1,70 +1,72 @@
...
+          />
+        </Tabs.Tab>
+      ))
+    }, [onScrollY, tabs])
+
+    const pagerProps = useMemo(() => {
+      return {
+        style: styles.overflow
+      }
+    }, [])
```
onghwan, line 44: Is this useMemo necessary? 😅

atn4z7: yes or else a new object will be created on every render which results in a new prop reference change and so on

onghwan (approved):

PR #2412, opened 2025-03-17
Fix to show twitter share button only when the app is installed

onghwan:
## Description
* Fix to show twitter share button only when the app is installed


ruijialin-avalabs (approved):
👍

PR #2419, opened 2025-03-19
Fix Portfolio being rendered 3 times on app unlocked

atn4z7:
## Description
please refer to the comments for details

## Screenshots/Videos/
before
https://github.com/user-attachments/assets/6cf10f9f-9225-4d19-a08f-e321b8ba97b6

after
https://github.com/user-attachments/assets/4cd6a06e-93d6-429f-8341-6fadf73be814

### DIFF
--- a/packages/core-mobile/app/new/routes/_layout.tsx
+++ b/packages/core-mobile/app/new/routes/_layout.tsx
@@ -6,10 +6,15 @@ import { K2AlpineThemeProvider } from '@avalabs/k2-alpine'
 import { Stack } from 'common/components/Stack'
 import NavigationThemeProvider from 'common/contexts/NavigationThemeProvider'
 import { GlobalToast } from 'common/utils/toast'
-import { selectIsReady, selectWalletState, WalletState } from 'store/app'
+import { selectWalletState, WalletState } from 'store/app'
 import { useSelector } from 'react-redux'
 import { useBgDetect } from 'navigation/useBgDetect'
-import { useFocusEffect, useRootNavigationState, useRouter } from 'expo-router'
+import {
+  useFocusEffect,
+  useRootNavigationState,
+  useRouter,
+  usePathname
+} from 'expo-router'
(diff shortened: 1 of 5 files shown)


packages/core-mobile/app/new/routes/_layout.tsx
```diff
@@ -56,32 +59,47 @@ export default function RootLayout(): JSX.Element | null {
...
       router.navigate('/loginWithPinOrBiometry')
-    } else if (walletState === WalletState.ACTIVE) {
-      if (canGoBackToWallet && router.canGoBack()) {
-        router.back()
-      } else {
-        router.replace('/portfolio')
-      }
     }
-  }, [walletState, router, canGoBackToWallet, navigation])
+  }, [walletState, router, navigation])
```
atn4z7, line 76: not really satisfied with this solution since we are heavily depending on the navigation routes instead of our app state. for now this should do it but we should look into refactoring this logic.

packages/core-mobile/app/new/routes/_layout.tsx
```diff
@@ -56,32 +59,47 @@ export default function RootLayout(): JSX.Element | null {
...
 
-    setShouldRenderOnlyPinScreen(walletState !== WalletState.ACTIVE)
-  }, [appIsReady, shouldRenderOnlyPinScreen, walletState])
+      if (isReturningFromInactivity) {
+        router.canGoBack() && router.back()
+      } else if (
+        pathName === '/onboarding/mnemonic/confirmation' || // onboarding completion
+        (pathName === '/loginWithPinOrBiometry' && !isSignedIn) // fresh app open
+      ) {
+        router.replace('/portfolio')
```
atn4z7, line 99: this was the main cause of the triple rendering of portfolio. we were calling `router.replace('/portfolio')` repeatedly. I added some checks here to make sure we only call it when needed.

onghwan (approved):
LGTM

PR #2441, opened 2025-03-24
CP-10064: Fix Fortuna Fuji Upgrade Bugs

atn4z7:
## Description

**Ticket: [CP-10064]** 

[CP-10064]: https://ava-labs.atlassian.net/browse/CP-10064?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/core-mobile/app/services/wallet/utils.ts
```diff
@@ -28,8 +31,21 @@ export const getAssetId = (avaxXPNetwork: Network): string => {
...
+    cChainToken.maxDecimals,
+    cChainToken.symbol
+  )
+
+  // after the Fortuna upgrade, the c-chain base fee can drop as low as 1 wei AVAX.
+  // here, we enforce a minimum base fee of 1 nano AVAX;
+  // otherwise, it rounds to 0 when converted back to nano AVAX
+  return adjustedBaseFee.toSubUnit() >= minAvax.toSubUnit()
+    ? adjustedBaseFee
+    : minAvax
```
atn4z7, line 50: this is the main fix

packages/core-mobile/app/screens/send/utils/evm/typeguard.ts
```diff
@@ -0,0 +1,16 @@
+import {
+  TokenType,
+  TokenWithBalance,
+  TokenWithBalanceEVM
+} from '@avalabs/vm-module-types'
+
+export const isSupportedToken = (
```
onghwan, line 7: Got a better name in mind? 😅 `isEVMToken`?

packages/core-mobile/app/screens/send/utils/evm/validate.ts
```diff
@@ -80,7 +81,9 @@ export const validateGasLimit = (gasLimit: bigint): void => {
   }
 }
 
-export const validateSupportedToken = (token: TokenWithBalanceEVM): void => {
+export function validateSupportedToken(
+  token: TokenWithBalance
+): asserts token is TokenWithBalanceEVM {
```
onghwan, line 86: Should we call the `isSupportedToken` function here instead?

packages/core-mobile/app/screens/send/utils/evm/typeguard.ts
atn4z7: it's not evm token actually since the type === native can also be applied to non evm tokens

packages/core-mobile/app/screens/send/utils/evm/validate.ts
atn4z7: one is an assert that will throw an error and the other is just a boolean check

onghwan (approved):
👏🏻

PR #2443, opened 2025-03-24
CP-10021: K2 Slider

onghwan:
## Description

**Ticket: [CP-10021]** 

* Implements slider component

## Screenshots/Videos
https://github.com/user-attachments/assets/58e2ce92-99e0-49ea-abac-2cefc43409e0

[CP-10021]: https://ava-labs.atlassian.net/browse/CP-10021?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/k2-alpine/package.json
+++ b/packages/k2-alpine/package.json
@@ -17,7 +17,7 @@
     "postinstall": "node_modules/.bin/patch-package"
   },
   "dependencies": {
-    "@miblanchard/react-native-slider": "^2.6.0",
+    "@miblanchard/react-native-slider": "2.6.0",
     "@shopify/react-native-skia": "1.5.0",
     "d3": "7.9.0",
     "dripsy": "4.3.7",
--- a/packages/k2-alpine/package.json
+++ b/packages/k2-alpine/package.json
@@ -17,6 +17,7 @@
     "postinstall": "node_modules/.bin/patch-package"
   },
   "dependencies": {
+    "@miblanchard/react-native-slider": "^2.6.0",
     "@shopify/react-native-skia": "1.5.0",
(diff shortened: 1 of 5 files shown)


packages/k2-alpine/package.json
```diff
@@ -17,6 +17,7 @@
     "postinstall": "node_modules/.bin/patch-package"
   },
   "dependencies": {
+    "@miblanchard/react-native-slider": "2.6.0",
```
onghwan, line 20: I initially tried using `@react-native-community/slider` since it's already a dependency due to Storybook, but its customization options aren't sufficient for our design requirements. So I decided to go with `@miblanchard/react-native-slider` instead.

packages/k2-alpine/src/components/Slider/Slider.stories.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+  return (
+    <View style={{ width: '100%', gap: 20 }}>
+      <Text variant="heading6" sx={{ alignSelf: 'center' }}>
+        Slide Value: {text}
+      </Text>
+      <Slider
+        thumbBorderColor={theme.colors.$surfacePrimary}
+        value={value}
+        onValueChange={newValue => (value.value = newValue)}
```
atn4z7, line 78: I don't think this callback is needed. Also I don't think you need to update the value here. The update should already happened by this point.

onghwan: removed.

PR #2449, opened 2025-03-26
CP-10043: Gas widget update

atn4z7:
## Description

**Ticket: [CP-10043]** 
support subtext notation in gas widget when gas has a lot of 0s

## Testing
please retest all transactions (send/swap/bridge) across networks (mainnets/testnets). when there are lots of 0s, it should display with the subtext notation.

## Checklist

[CP-10043]: https://ava-labs.atlassian.net/browse/CP-10043?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core-mobile/app/components/NetworkFeeSelector.tsx
+++ b/packages/core-mobile/app/components/NetworkFeeSelector.tsx
@@ -193,13 +193,11 @@ const NetworkFeeSelector = ({
 
   return (
     <>
-      <Space y={4} />
       <View
         sx={{
           backgroundColor: isDark ? '$neutral900' : '$neutral800',
           padding: 16,
-          borderRadius: 8,
-          marginBottom: 16
+          borderRadius: 8
         }}>
         {!networkFee?.isFixedFee && (
           <>
--- a/packages/core-mobile/app/components/NetworkFeeSelector.tsx
+++ b/packages/core-mobile/app/components/NetworkFeeSelector.tsx
@@ -242,7 +242,7 @@ const NetworkFeeSelector = ({
         <>
           <Row
             style={{ justifyContent: 'space-between', alignItems: 'center' }}>
-            <Text variant="caption" sx={{ color: '$neutral400' }}>
+            <Text variant="caption" sx={{ color: '$neutral200' }}>
               Fee Amount
             </Text>
             <View sx={{ flexDirection: 'row' }}>
--- a/packages/core-mobile/app/components/NetworkFeeSelector.tsx
+++ b/packages/core-mobile/app/components/NetworkFeeSelector.tsx
@@ -1,5 +1,4 @@
 import { Row } from 'components/Row'
-import Settings from 'assets/icons/settings.svg'
 import { Space } from 'components/Space'
 import React, {
   FC,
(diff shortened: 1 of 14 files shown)


packages/core-mobile/app/components/NetworkFeeSelector.tsx
```diff
@@ -226,45 +193,11 @@ const NetworkFeeSelector = ({
...
-              ))}
-          </Row>
-        </>
-      )}
-      <Space y={4} />
-
       <View
         sx={{
-          backgroundColor: '$neutral900',
+          backgroundColor: isDark ? '$neutral900' : '$neutral800',
```
onghwan, line 198: I thought dark/light mode had been added. let's just allow the backgroundColor prop to be overridden, with the default value of $neutral800.

packages/core-mobile/app/components/SubTextNumber.tsx
```diff
@@ -0,0 +1,62 @@
...
+  const textVariant = size === 'big' ? 'heading5' : 'subtitle2'
+  return (
+    <View style={styles.container} testID={testID}>
+      <Text variant={textVariant} sx={{ color: '$neutral50' }}>
+        {mainTextBefore}
+      </Text>
+      {subText && (
+        <Text
+          style={size === 'big' ? styles.subTextBig : styles.subTextSmall}
+          variant={textVariant}
```
onghwan, line 31: Isn’t it okay to omit `textVariant` here?

packages/core-mobile/app/hooks/useGasless.ts
```diff
@@ -0,0 +1,137 @@
...
+  signingData: SigningData
+  caip2ChainId: string
+}
+
+type Return = {
+  gaslessEnabled: boolean
+  setGaslessEnabled: React.Dispatch<React.SetStateAction<boolean>>
+  showGaslessSwitch: boolean
+  gaslessError: Alert | null
+  handleGaslessTx: (addressFrom: string) => Promise<string | undefined>
```
onghwan, line 25: Since it’s already useGasless, maybe we can omit the word “gasless” from these names?

packages/core-mobile/app/hooks/useGasless.ts
```diff
@@ -0,0 +1,137 @@
...
+type Params = {
+  maxFeePerGas: bigint | undefined
+  signingData: SigningData
+  caip2ChainId: string
+}
+
+type Return = {
+  gaslessEnabled: boolean
+  setGaslessEnabled: React.Dispatch<React.SetStateAction<boolean>>
+  showGaslessSwitch: boolean
```
onghwan, line 23:
```suggestion
  shouldShowSwitch: boolean
```

packages/core-mobile/app/hooks/useGasless.ts
```diff
@@ -0,0 +1,137 @@
...
+  gaslessError: Alert | null
+  handleGaslessTx: (addressFrom: string) => Promise<string | undefined>
+}
+
+export const useGasless = ({
+  signingData,
+  maxFeePerGas,
+  caip2ChainId
+}: Params): Return => {
```
onghwan, line 32: nice hook.

onghwan (commented):
🔥

packages/core-mobile/app/components/SubTextNumber.tsx
atn4z7: no, because we still need everything else except the fontSize from the variant style

packages/core-mobile/app/components/NetworkFeeSelector.tsx
atn4z7: this is actually just a temp solution since design didn't account for the approval UI. `isDark` (or whatever name) here is just a way for us to hardcode the correct colors and these are the only colors that we will use.

atn4z7: also this component will be gone after next sprint

packages/core-mobile/app/hooks/useGasless.ts
atn4z7: I think it's good to keep it as it will be clearer in the component that uses it (instead of just `setEnabled`, `error`,...).

packages/core-mobile/app/hooks/useGasless.ts
atn4z7: renamed to shouldShowGaslessSwitch

packages/core-mobile/app/hooks/useGasless.ts
onghwan: Yeah, but if we think it of using it like const gasless = useGasless(), then calling things like gasless.setGaslessEnabled or gasless.gaslessError feels a bit redundant from a naming perspective.

Since it’s not a major issue, I’ll go ahead and approve it.

onghwan (approved):

PR #2456, opened 2025-03-27
make react-native-quick-crypto work again and remove promise polyfill

atn4z7:
after upgrading react native and "friends", looks like the global.crypto is not being set anymore. this pr fixes it. the app should be much faster now whenever a crypto operation occurs.


ruijialin-avalabs (approved):
❤️

PR #2483, opened 2025-04-02
Fix settings bug part 2

ruijialin-avalabs:
## Description

- update testnet mode is on design to show just snackbar instead
- fix bugs in https://ava-labs.atlassian.net/browse/CP-10187?focusedCommentId=50373

## Checklist

### DIFF
--- a/packages/core-mobile/app/new/features/accountSettings/components/NotificationToggle.tsx
+++ b/packages/core-mobile/app/new/features/accountSettings/components/NotificationToggle.tsx
@@ -1,14 +1,12 @@
-import React, { useCallback, memo, useState } from 'react'
+import React, { useCallback, memo, useMemo } from 'react'
 import { useTheme, View, Text, Toggle } from '@avalabs/k2-alpine'
 import { useDispatch, useSelector } from 'react-redux'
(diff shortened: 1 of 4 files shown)


packages/core-mobile/app/new/features/accountSettings/components/NotificationToggle.tsx
```diff
@@ -21,21 +19,14 @@ const NotificationToggle = ({
...
+    [inAppEnabled, isSystemDisabled]
+  )
   const dispatch = useDispatch()
 
   const onValueChange = useCallback(
     async (isChecked: boolean): Promise<void> => {
-      setIsEnabled(isChecked)
-
-      // before we change the state, we need to check if the system settings allow us to do so
-      const { permission } = await NotificationsService.getAllPermissions(false)
```
ruijialin-avalabs, line 32: this is causing the flickering.  we actually don't need to call this here.  in the parent component, we already check if we have permission, and also check if the channel is blocked,  if it is blocked, it is passed in as prop isSystemDisabled.  which disables the toggle.

atn4z7 (approved):

PR #2496, opened 2025-04-05
Cp 10230

bogdandobritoiu:
## Description

**Ticket: [CP-10230]** 

### NFT Detail Screen

**FIXED** - **High** - After hiding all, I land on Oops screen. We should navigate back instead of showing oops screen. Do not show oops screen when everything is hidden
**FIXED** - **Low** - [design QA] slight gradient shouldn't be visible on the top & bottom of the list
**FIXED** - **Medium** -  ANDROID ONLY: when I just want to scroll the nft data section (created by standard chain…), the whole screen moves and I can’t scroll the nft data section only. always the nft full screen follows and I can’t scroll up and down the data list. However, I can scroll the data section only fine on iOS. It’s just android having the issue. please see the video below, the device on the left is iOS and the other is Android.

**FIXED** - **High** - Hide button is not responding with one tap. I had to hit twice to finally hide it. 
**Also fixed it for assets management screen. Both had double toggle for hiding bug** 
**FIXED** - **High** - When the ‘collectible hidden’ toast msg is displayed the other elements at the top of the screen cannot be tapped, i.e. the back button
**I made the whole area dismissable now but you have to press once to dismiss and once to go back or prev/next**

### NFT Image full screen

**FIXED** - Cannot scroll nft description text on android
**FIXED** - remove glow when launch
**FIXED** - make the bottom arrow button bigger [(design QA)]

### NFT grid/list View 

**FIXED** - **High** hide filter and view button when it’s empty state 
**FIXED** - manage list is not updated after being shown. they are still toggled OFF 
**FIXED** - **Low** ListView: number button bg looks darker than design
**FIXED** **High** Sort button is missing. 
**FIXED** **High** Sort Bug when we have sort button on the screen: sort a-z or sort z-a seems to be not working. When I select name z to a, I see they are not sorted correctly. I can’t verify date added sort

**Both of these have the same issue which is the way we process collectibles => tech debt**
**High** when I hit show hidden button, two bugs are visible
- show unreachable nfts placeholder appears for a moment before displaying data
- when we first load data, the screen flickers to load different networks nft and sort. 

**NFT Manage list Screen**

**FIXED** - when hide unreachable button is ON, we should gray out the toggle of the unreachable nfts
**FIXED** - no padding between manage list, searhc bar, and hide button
**FIXED** - I had to hit toggle twice to disable it sometimes

## Screenshots/Videos

https://github.com/user-attachments/assets/c8e182e2-b140-4ba3-aba7-33bab798b405

https://github.com/user-attachments/assets/e8688dfc-6a4a-41ce-98c3-5c0e9c08b4c0

Screenshot 2025-04-05 at 03 08 28

## Checklist

### DIFF
--- a/packages/core-mobile/app/store/portfolio/slice.ts
+++ b/packages/core-mobile/app/store/portfolio/slice.ts
@@ -13,8 +13,13 @@ export const portfolioSlice = createSlice({
       action: PayloadAction<{ tokenId: string }>
     ) => {
       const { tokenId } = action.payload
-      state.tokenVisibility[tokenId.toLowerCase()] =
-        !state.tokenVisibility[tokenId.toLowerCase()]
+
+      if (state.tokenVisibility?.[tokenId.toLowerCase()] != null) {
+        state.tokenVisibility[tokenId.toLowerCase()] =
+          !state.tokenVisibility[tokenId.toLowerCase()]
+      } else {
+        state.tokenVisibility[tokenId.toLowerCase()] = false
+      }
     },
     toggleCollectibleVisibility: (
       state,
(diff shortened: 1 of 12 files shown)


packages/core-mobile/app/store/portfolio/slice.ts
```diff
@@ -13,17 +13,26 @@ export const portfolioSlice = createSlice({
       action: PayloadAction<{ tokenId: string }>
     ) => {
       const { tokenId } = action.payload
-      state.tokenVisibility[tokenId.toLowerCase()] =
-        !state.tokenVisibility[tokenId.toLowerCase()]
+
+      if (state.tokenVisibility?.[tokenId.toLowerCase()] != null) {
```
atn4z7, line 17: we actually usually try not to use != null since it is kinda vague. it looks like you can just check against undefined

PR #2505, opened 2025-04-07
Seedless Bug fixes

ruijialin-avalabs:
## Description

- add `AppCompatDelegate.setDefaultNightMode(AppCompatDelegate.MODE_NIGHT_FOLLOW_SYSTEM)` to mainActivity, so that Android sets the color scheme correctly
- address this [issue](https://ava-labs.atlassian.net/browse/CP-10263?focusedCommentId=50706) by clearing appearance data on logout and set to system when app is launched
- update hide/show button to use visibility_off and visibility_on icon to toggle phrase on/off
- move useSelector hook outside of SelectionDot component which the value is not properly propagate to react-native-skia canvas.
- fixed the back arrow position in seedless readyToExport screen on android
- UI copy change to sentence case

## Checklist


packages/core-mobile/android/app/src/main/java/com/avaxwallet/MainActivity.kt
```diff
@@ -32,6 +33,7 @@ class MainActivity : ReactActivity() {
     override fun onCreate(savedInstanceState: Bundle?) {
         RNBootSplash.init(this, R.style.BootTheme)
         super.onCreate(null)
+        AppCompatDelegate.setDefaultNightMode(AppCompatDelegate.MODE_NIGHT_FOLLOW_SYSTEM)
```
atn4z7, line 36: can you add a comment in the code here?

packages/core-mobile/app/new/routes/(signedIn)/(modals)/accountSettings/seedlessExportPhrase/readyToExport.tsx
```diff
@@ -72,15 +73,30 @@ const SeedlessExportReadyScreen = (): JSX.Element => {
   }, [onCancelExportRequest])
 
   const renderCustomBackButton = useCallback(
-    () => <BackBarButton onBack={customGoBack} />,
+    () => (
+      <View
+        sx={{
+          marginLeft: Platform.OS === 'android' ? 12 : 0
```
atn4z7, line 79: do you think we will need to reuse this logic elsewhere?

ruijialin-avalabs: not sure if we need to apply this again, but if we do, we should then move it somewhere and reuse it.

PR #2518, opened 2025-04-09
add missing mfa for token expired

ruijialin-avalabs:
## Description

- add the verify MFA flow for token expired.  MFA screen would show right after social sign in.

## Screenshots/Videos

https://github.com/user-attachments/assets/03da606c-7e5e-49e4-9ab0-b47159d39990

## Checklist


packages/core-mobile/app/new/common/hooks/useInitWalletAndUnlock.ts
```diff
@@ -0,0 +1,32 @@
...
+import { useDispatch, useSelector } from 'react-redux'
+import { SEEDLESS_MNEMONIC_STUB } from 'seedless/consts'
+import { WalletType } from 'services/wallet/types'
+import { selectWalletState, selectWalletType, WalletState } from 'store/app'
+import Logger from 'utils/Logger'
+
+// This hook is used to initialize the seedless wallet and unlock it if the wallet state is inactive.
+export const useInitWalletAndUnlock = (): {
+  initWalletAndUnlock: () => Promise<void>
+} => {
```
atn4z7, line 12: nit: can we rename this to emphasize it's only about seedless?

packages/core-mobile/app/new/routes/sessionExpired/index.tsx
```diff
@@ -37,19 +35,46 @@ const SessionTimeoutScreen = (): React.JSX.Element => {
...
+              })
+              return
+            } else {
+              await SeedlessService.session.approveFido(
+                result.value.oidcToken,
+                result.value.mfaId,
+                true
+              )
+              initWalletAndUnlock()
+              router.canGoBack() && router.back()
```
atn4z7, line 67: do we need to await here before going back?

ruijialin-avalabs: I think we should await to init seedless wallet before going back again.

packages/core-mobile/app/new/common/hooks/useInitWalletAndUnlock.ts
ruijialin-avalabs: updated.

atn4z7 (approved):

PR #2546, opened 2025-04-15
Fix camera permission and zero balance to show - or hide it

ruijialin-avalabs:
## Description

**Ticket: []** 

- fix camera permission on android
- set zero balance to - in portoflio screen, and hide the amount for un-fetched accounts
- check if availableAvax or totalStakedInAvax is zero before doing division

## Screenshots/Videos

keyboardAwareScrollView effect

https://github.com/user-attachments/assets/22c71336-0f2e-404d-a6b2-9840a0b56e4c

## Checklist

### DIFF
--- a/packages/core-mobile/android/app/src/main/AndroidManifest.xml
+++ b/packages/core-mobile/android/app/src/main/AndroidManifest.xml
@@ -31,7 +31,7 @@
             android:label="@string/app_name"
             android:launchMode="singleTop"
             android:screenOrientation="portrait"
-            android:windowSoftInputMode="adjustResize"
+            android:windowSoftInputMode="adjustPan"
             android:theme="@style/BootTheme"
             >
             <intent-filter android:autoVerify="true">
--- a/packages/core-mobile/android/app/src/main/AndroidManifest.xml
+++ b/packages/core-mobile/android/app/src/main/AndroidManifest.xml
@@ -31,7 +31,7 @@
             android:label="@string/app_name"
             android:launchMode="singleTop"
             android:screenOrientation="portrait"
-            android:windowSoftInputMode="adjustPan"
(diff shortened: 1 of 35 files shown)


packages/core-mobile/android/app/src/main/AndroidManifest.xml
```diff
@@ -31,7 +31,7 @@
             android:label="@string/app_name"
             android:launchMode="singleTop"
             android:screenOrientation="portrait"
-            android:windowSoftInputMode="adjustResize"
+            android:windowSoftInputMode="adjustPan"
```
ruijialin-avalabs, line 34: update to adjustPan to avoid keyboard always pushing all content up.  this will only pan the input field so that use sees what they type in.

packages/core-mobile/app/new/common/components/QrCodeScanner.tsx
```diff
@@ -48,69 +48,46 @@ export const QrCodeScanner = ({
...
+  }, [checkIosPermission])
+
+  const checkAndroidPermission = useCallback(async () => {
+    // on android, permission.canAskAgain returns false when user select "ask every time",
+    // so we separate the logic for android
+    if (Platform.OS === 'android') {
+      const status = await check(PERMISSIONS.ANDROID.CAMERA).catch(Logger.error)
+      if (status === PermissionStatus.GRANTED) {
+        setIsAndroidPermissionGranted(true)
+        return
```
atn4z7, line 73: do you think this logic can be reused for other permission?

ruijialin-avalabs: we can revisit this after.

atn4z7 (approved):
nice add on the keyboard stuff.

PR #2592, opened 2025-04-24
Hide set as my avatar button for video 

ruijialin-avalabs:
## Description

**Ticket: []** 

## Screenshots/Videos

https://github.com/user-attachments/assets/bd1c663a-594d-480e-b14b-6403196700fb

## Checklist


bogdandobritoiu:
we should actually hide the button instead of displaying that toast (design confirmed we should hide the button if the collectible is either unprocessable or a video)

packages/core-mobile/app/new/features/portfolio/collectibles/components/CollectibleDetailsContent.tsx
```diff
@@ -49,9 +49,16 @@ export const CollectibleDetailsContent = ({
...
+  const isVideo =
+    collectible?.imageData?.type.includes('video') ||
+    collectible?.imageData?.type.includes('mp4')
 
+  const handleSaveAvatar = (): void => {
+    if (isVideo) {
+      showSnackbar('Video avatars are not supported')
+      return
+    }
+    if (!collectible?.imageData?.image) return
```
onghwan, line 61: Can a collectible be a video and still have an imageData.image (like a thumbnail)? If so, wouldn’t it be enough to just check whether imageData.image exists?

ruijialin-avalabs: based on the data return, I don't think it will return both image and video, since imageData contains only one type.  but let's just hide the button.

atn4z7 (approved):

PR #2618, opened 2025-04-29
Send bugs

ruijialin-avalabs:
## Description

**Ticket: []** 

- sort primary networks to top on portfolio screen and send select token screen
- fix android add contact textInput field height
- remove sort animation on portfolio screen
- fix empty recipient for send
- hide wallet Info (including show private key button)
- add done return key to input on send screen
- fix crash on setting max amount on send screen
- fix issue with editing amount in decimal places on TokenUnitInput (it was setting max digits to 7, if the entered amount is more than 7, editing the amount will reset, updated to just use token.decimals)
- hide placeholder avatar if user scans qrcode and enter/paste an address since we don't have avatar for an address

## Checklist

### DIFF
--- a/packages/core-mobile/app/new/features/portfolio/assets/hooks/useAssetsFilterAndSort.ts
+++ b/packages/core-mobile/app/new/features/portfolio/assets/hooks/useAssetsFilterAndSort.ts
@@ -18,6 +18,7 @@ import {
   isTokenWithBalancePVM
 } from '@avalabs/avalanche-module'
 import { useErc20ContractTokens } from 'common/hooks/useErc20ContractTokens'
+import { sortedTokensWithBalance } from 'common/utils/sortTokensWithBalance'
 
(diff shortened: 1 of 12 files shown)


packages/core-mobile/app/new/features/portfolio/assets/hooks/useAssetsFilterAndSort.ts
```diff
@@ -108,7 +109,8 @@ export const useAssetsFilterAndSort = (): {
 
   const filteredAndSorted = useMemo(() => {
     const filtered = getFiltered()
-    return getSorted(filtered)
+    const sorted = getSorted(filtered)
+    return sortedTokensWithBalance(sorted)
```
atn4z7, line 113: should we just do sortedTokensWithBalance here since it will override the output of getSorted?

atn4z7 (approved):

atn4z7 (approved):

PR #2620, opened 2025-04-29
Edge 2 edge + FlatlistScreen and ScrollViewScreen

bogdandobritoiu:
## Description

**Ticket: [CP-10330]** 

Motivation
Given the big amount of issues on each screen related to button placements/tap gestures/safeareas/scrolling while keyboard is open, it was very hard to fix copy paste the solution for all of these edge cases. So, I've created two template screens that handle all of these gestures in a controlled manner and replaces the code on 95% of our apps screens.

This change removed a lot of duplicate code that would have been hard to maintain on each screen.

During these changes I have also adapted the navigation bars and buttons as per design and have been included in these components.

There is not much to say on how to reuse these two components ScrollviewScreenTemplate and FlatlistScreenTemplate except the fact that they have two props

1. isModal => fixes modal bottom buttons (necessary to be added in all modal screens)
2. hasParent => stacked screen that requires to have this prop otherwise the navigation buttons are broken and do not display properly
3. renderHeader => add a fixed header underneath the title to remain sticky while scrolling

Changes
- added edge2edge on android => removed expo-status-bar
- I've went through most of the screens, except the main portfolio screens which I didn't touch (because of the segmented control and CollapsibleTabs which are very different than the rest of the screens)
- Refactored the navigation header and the buttons to account for modal handle
- Refactored screenOptions for modals to have showTransparent as true by default
- Added a sticky component for scrolling inside FlatListScreenTemplate

## Checklist


packages/core-mobile/app/new/common/components/ScrollViewScreenTemplate.tsx
```diff
@@ -0,0 +1,162 @@
...
+  useAnimatedStyle,
+  useSharedValue
+} from 'react-native-reanimated'
+import { useSafeAreaInsets } from 'react-native-safe-area-context'
+import { BlurViewWithFallback } from './BlurViewWithFallback'
+import { KeyboardAvoidingView } from './KeyboardAvoidingView'
+import { LinearGradientBottomWrapper } from './LinearGradientBottomWrapper'
+import ScreenHeader from './ScreenHeader'
+
+interface FlatListScreenTemplateProps extends KeyboardAwareScrollViewProps {
```
onghwan, line 25:
```suggestion
interface ScrollViewScreenTemplate extends KeyboardAwareScrollViewProps {
```

packages/core-mobile/app/new/common/components/ScrollViewScreenTemplate.tsx
```diff
@@ -0,0 +1,162 @@
...
+  title?: string
+  subtitle?: string
+  children: React.ReactNode
+  hasParent?: boolean
+  isModal?: boolean
+  disabled?: boolean
```
onghwan, line 31:
```suggestion
  shouldAvoidKeyboard?: boolean
```
I think it’s better to base it on the enabled rather than the disabled.

packages/core-mobile/package.json
```diff
@@ -152,6 +150,7 @@
     "react-native-collapsible-tab-view": "8.0.0",
     "react-native-config": "1.5.3",
     "react-native-device-info": "13.0.0",
+    "react-native-edge-to-edge": "^1.6.0",
```
atn4z7, line 153:
```suggestion
    "react-native-edge-to-edge": "1.6.0",
```

packages/core-mobile/app/new/features/send/components/SendToken.tsx
```diff
@@ -153,145 +144,141 @@ export const SendToken = ({ onSend }: { onSend: () => void }): JSX.Element => {
...
-          <Text variant="heading2" style={{ marginBottom: 12 }}>
-            {`${'How much would\nyou like to send?'}`}
-          </Text>
-        </Animated.View>
-        {/* Send to */}
-        <Card
+    <ScrollScreen
+      isModal
+      title={`${'How much would\nyou like to send?'}`}
+      navigationTitle={`${'How much would you like to send?'}`}
```
atn4z7, line 168:
```suggestion
      navigationTitle={'How much would you like to send?'}
```

PR #2661, opened 2025-05-02
Fix for ALL android buttons

bogdandobritoiu:
## Description

Features
- Fix android bottom padding on all screens including 3 button navigation
- Added a ScrollView instead of KeyboardAwareScrollView for screens without keyboard interactions
- Fixed Account settings scroll issue on Android
- Fixed Swap/Stake/Bridge bottom screen padding

## Checklist


packages/core-mobile/app/new/common/components/ScrollScreen.tsx
```diff
@@ -206,7 +205,86 @@ export const ScrollScreen = ({
...
+              {renderFooter?.()}
+            </View>
+          </LinearGradientBottomWrapper>
+        ) : null}
+      </KeyboardAvoidingView>
+    )
+  }
+
+  return (
+    <KeyboardAvoidingView enabled={false} style={{ flex: 1 }}>
```
atn4z7, line 262: should we just use View here then since enabled is false?

PR #2737, opened 2025-05-12
Browser refactor + other fixes

bogdandobritoiu:
## Description

Browser
- closing a tab also navigated back
- Favorites and History had bad zIndex for Android 
- toggle between favorites and history animation
- browser history switch to ListScreen
- fixed the double opacity animation triggered when alert was opening on top of focused keyboard
- history, moved the clear all button in the navigation headerRight

Other
- TokenDetail list animations fix
- ListScreen can now be used on screens that have a tabbar and are not modals (new prop hasTabBar)
- Fixed SearchBar tap area to include the left icon and full height
- ConnectedWebsites, moved the disconnect all button at the top (design confirmed)
- Added useBottomTabBarHeight as a hook for throwing Error that it's not in the context (needed for ListScreen, just for accessing in case it isn't inside a tabbar route tree)
- dismiss keyboard when navigating back

## Checklist


packages/core-mobile/app/new/common/hooks/useBottomTabBarHeight.ts
```diff
@@ -0,0 +1,9 @@
+import { useBottomTabBarHeight as useRNTabBarHeight } from 'react-native-bottom-tabs'
+
+export function useBottomTabBarHeight(fallback = 0): number {
```
atn4z7, line 3: can you add a comment explaining when the fallback will happen/be needed?

packages/core-mobile/app/new/common/components/ListScreen.tsx
```diff
@@ -96,14 +104,19 @@ export const ListScreen = <T,>({
...
+  const keyboard = useKeyboardState()
+
+  const tabBarHeight = useBottomTabBarHeight()
 
   const { onScroll, scrollY, targetHiddenProgress } = useFadingHeaderNavigation(
     {
       header: <NavigationTitleHeader title={navigationTitle ?? title ?? ''} />,
       targetLayout: headerLayout,
       shouldHeaderHaveGrabber: isModal,
+      hideHeaderBackground: true,
```
atn4z7, line 116: why do want to hideHeaderBackground for ListScreen?

packages/core-mobile/app/new/routes/(signedIn)/(tabs)/browser/index.tsx
```diff
@@ -93,16 +93,11 @@ const Browser = (): React.ReactNode => {
   })
 
   useFocusEffect(() => {
-    if (!KeyboardController.isVisible()) {
+    if (Number(Platform.Version) > 29) {
```
atn4z7, line 96: can you add a comment here for this check?

bogdandobritoiu: Removed the check here

packages/core-mobile/app/new/common/components/ListScreen.tsx
bogdandobritoiu: I've updated it to check if renderHeader is present, if so then most likely it has a searchbar and that means the ListScreen will control the background and separator instead so we need to remove them from the navigation header

atn4z7 (approved):

PR #2750, opened 2025-05-14
Fix bottom sheet flickering on Android

atn4z7:
## Description

dismissing a modal via gesture for some reason results in some flickering on Android. for now, looks like the best solution is just to use 0 duration transition on Android to prevent the flickering. the drawback is that it also affects the transition when pressing physical back button.

## Screenshots/Videos
**before**
https://github.com/user-attachments/assets/bbeb7d85-e0e4-436c-8b1d-44b8f375510a

**after**
https://github.com/user-attachments/assets/ab56d2e6-07ad-42e0-bcac-b5f8cc393570

## Checklist

### DIFF
--- a/packages/core-mobile/app/new/common/hooks/useModalScreenOptions.tsx
+++ b/packages/core-mobile/app/new/common/hooks/useModalScreenOptions.tsx
@@ -2,7 +2,8 @@ import {
   CardStyleInterpolators,
   StackCardInterpolatedStyle,
   StackCardInterpolationProps,
-  StackNavigationOptions
+  StackNavigationOptions,
+  TransitionSpecs
 } from '@react-navigation/stack'
 import {
   MODAL_BORDER_RADIUS,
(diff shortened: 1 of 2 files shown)


onghwan (dismissed):
🔥

packages/core-mobile/app/new/common/hooks/useModalScreenOptions.tsx
```diff
@@ -79,6 +89,13 @@ export function useModalScreenOptions(): {
             marginTop: 0,
             paddingTop: insets.top
           },
+          transitionSpec: {
+            open: TransitionSpecs.BottomSheetSlideInSpec,
+            close: {
+              animation: 'timing',
+              config: { duration: 0 }
+            }
+          },
```
onghwan, line 98: nit: can we declare `androidModalTransitionSpec` constant and reuse it?

onghwan (approved):

PR #2835, opened 2025-05-28
Add back invalidateQueries logic for getNetwork

ruijialin-avalabs:
## Description

**Ticket: []** 

- add back invalidateQueries for `ReactQueryKeys.NETWORKS ` query to update network on app unlocked

## Checklist

### DIFF
--- a/packages/core-mobile/app/new/features/onboarding/components/SetWalletName.tsx
+++ b/packages/core-mobile/app/new/features/onboarding/components/SetWalletName.tsx
@@ -28,7 +28,7 @@ export const SetWalletName = ({
     <ScrollScreen
       shouldAvoidKeyboard
       showNavigationHeaderTitle={false}
-      title="How would you like to name your wallet?"
+      title="Add a name for your wallet"
       renderFooter={renderFooter}
       contentContainerStyle={{ padding: 16 }}>
       <View


packages/core-mobile/app/hooks/networks/useNetworksListener.ts
```diff
@@ -1,12 +1,29 @@
...
+import { toggleDeveloperMode } from 'store/settings/advanced'
 
 export const useNetworksListener = (queryClient: QueryClient): void => {
   const dispatch = useDispatch()
 
+  // @ts-ignore
+  useEffect(() => {
+    return dispatch(
+      addListener({
+        matcher: isAnyOf(toggleDeveloperMode, onAppUnlocked),
```
atn4z7, line 17: I think just onAppUnlocked is good enough. also can you add a comment explaining the reason we are doing this (cause we totally forgot haha)?

ruijialin-avalabs: let's keep the toggleDeveloperMode just in case, what if we need to update something for testmode network.

ruijialin-avalabs: added the comment

PR #2838, opened 2025-05-29
Show Accounts based on Active status

bogdandobritoiu:
## Description

**Ticket: [CP-10749]** 

Features
- Added zustand store for recentAccountsIndexes with persistence 
- Ordering of recentAccountIndexes by last activeAccount.index pressed
- Delete the recentAccountIndexes when account is deleted

## Screenshots/Videos

https://github.com/user-attachments/assets/514a8646-b7e2-43d3-bbba-97f2b4e428e0

## Checklist

### DIFF
--- a/packages/core-mobile/app/new/features/accountSettings/components/AcccountList.tsx
+++ b/packages/core-mobile/app/new/features/accountSettings/components/AcccountList.tsx
@@ -1,17 +1,18 @@
-import { AnimatedPressable, useTheme, View, Text } from '@avalabs/k2-alpine'
+import { AnimatedPressable, Text, useTheme, View } from '@avalabs/k2-alpine'
+import { getItemEnteringAnimation } from 'common/utils/animations'
 import { useRouter } from 'expo-router'
-import React, { useCallback, useMemo, useRef } from 'react'
+import React, { useCallback, useEffect, useMemo, useRef } from 'react'
+import { FlatList } from 'react-native-gesture-handler'
+import Animated, { LinearTransition } from 'react-native-reanimated'
 import { useDispatch, useSelector } from 'react-redux'
+import AnalyticsService from 'services/analytics/AnalyticsService'
 import {
   Account,
   selectAccounts,
   selectActiveAccount,
   setActiveAccountIndex
 } from 'store/account'
-import Animated, { LinearTransition } from 'react-native-reanimated'
-import { getItemEnteringAnimation } from 'common/utils/animations'
-import AnalyticsService from 'services/analytics/AnalyticsService'
-import { FlatList } from 'react-native-gesture-handler'
+import { useRecentAccounts } from '../store'
 import { AccountItem } from './AccountItem'
 
 const CARD_PADDING = 12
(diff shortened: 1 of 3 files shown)


packages/core-mobile/app/new/features/accountSettings/store.ts
```diff
@@ -1,6 +1,40 @@
 import { AvatarType } from '@avalabs/k2-alpine'
 import { createZustandStore } from 'common/utils/createZustandStore'
+import { create } from 'zustand'
+import { persist, createJSONStorage } from 'zustand/middleware'
+import AsyncStorage from '@react-native-async-storage/async-storage'
```
atn4z7, line 5: let's use mmkv. we don't really use async storage anymore.

packages/core-mobile/app/new/features/accountSettings/components/AcccountList.tsx
```diff
@@ -26,18 +27,25 @@ export const AccountList = (): React.JSX.Element => {
...
-  const accountsToDisplay = useMemo(() => {
-    return accounts.slice(0, 5)
-  }, [accounts])
+  const recentAccounts = useMemo(() => {
+    return recentAccountIndexes.map(index => accountCollection[index])
+  }, [accountCollection, recentAccountIndexes])
+
+  useEffect(() => {
+    if (activeAccount?.index != null) {
+      addRecentAccount(activeAccount.index)
```
atn4z7, line 43: I don't think we should do this here since this will only trigger when this component is mounted and we do change active account in other places as well (in some background logic for example). you should do this in a listener in account slice and listen to setActiveAccountIndex action instead.

packages/core-mobile/app/utils/mmkv/storages.ts
```diff
@@ -13,3 +15,16 @@ export const commonStorage = new MMKV({
 export const queryStorage = new MMKV({
   id: `query`
 })
+
+export const zustandMMKVStorage = {
```
atn4z7, line 19:
```suggestion
export const zustandMMKVStorage: StateStorage = {
```

packages/core-mobile/app/utils/mmkv/storages.ts
```diff
@@ -1,9 +1,11 @@
 import { MMKV } from 'react-native-mmkv'
 import { StorageKey } from 'resources/Constants'
+import { StorageValue } from 'zustand/middleware'
 
 export const commonStorageKeys = [
   StorageKey.POSTHOG_SUSPENDED,
-  StorageKey.SECURE_ACCESS_SET
+  StorageKey.SECURE_ACCESS_SET,
+  StorageKey.RECENT_ACCOUNTS
 ]
```
atn4z7, line 9: let's create zustandStorageKeys

packages/core-mobile/app/utils/mmkv/storages.ts
bogdandobritoiu: had to use const zustandMMKVStorage: PersistStorage<StateStorage>

atn4z7 (approved):

PR #2887, opened 2025-06-06
CP-10753: Fetch Solana Balances

B0Y3R-AVA:
… now

## Description

**Ticket: [CP-10753](https://ava-labs.atlassian.net/browse/CP-10753)** 

- Added support for Solana token types (SPL and native SOL)
  - Updated balance service to handle Solana-specific token types
  - Added proper token type detection for Solana VM networks
  - Added temporary hardcoded Solana address for testing

- Enhanced Network Handling
  - Added Solana networks to primary chain IDs (mainnet, testnet, devnet)
  - Updated network caching to respect Solana support flag
  - Improved CAIP2 chain ID handling for Solana networks
  - Added proper error handling for unsupported networks

- Code Improvements
  - Added network VM type checks for token type selection
  - Updated network mapping to include CAIP ID
  - Simplified Solana chain ID resolution logic
  - Added proper type definitions for Solana-specific functionality

## Screenshots/Videos

https://github.com/user-attachments/assets/dafaf45d-18ae-4630-bded-4f639df4c3c0

## Testing
- Verify balance fetching works for both SPL tokens and native SOL
- Test network switching between Solana and other networks
- Confirm proper handling of Solana addresses
- Check network caching behavior with Solana support enabled/disabled

## Checklist

[CP-10753]: https://ava-labs.atlassian.net/browse/CP-10753?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core-mobile/app/store/balance/listeners.ts
+++ b/packages/core-mobile/app/store/balance/listeners.ts
@@ -42,8 +42,6 @@ import SentryWrapper from 'services/sentry/SentryWrapper'
 import { ReactQueryKeys } from 'consts/reactQueryKeys'
 import { queryClient } from 'contexts/ReactQueryProvider'
 import { selectIsSolanaSupportBlocked } from 'store/posthog'
-import { TokenType } from '@avalabs/vm-module-types'
-import { isSolanaChainId } from 'utils/network/isSolanaNetwork'
 import { Balances, LocalTokenWithBalance, QueryStatus } from './types'
 import {
   fetchBalanceForAccount,


packages/core-mobile/app/store/balance/listeners.ts
```diff
@@ -345,6 +347,20 @@ const fetchBalanceForNetworks = async (
...
       }
+      if (isSolanaChainId(chainId)) {
+        return [
+          ...tokenBalance,
+          {
+            ...token,
+            localId:
+              token.type === TokenType.SPL
+                ? token.address
+                : `SOL_${token.symbol}`,
```
atn4z7, line 358: should we be stricter here regarding the token type?  `SOL_${token.symbol}` will apply to any token type that is not SPL. also, could there be 2 tokens with the same symbol here?

packages/core-mobile/app/store/balance/utils.ts
```diff
@@ -7,13 +7,20 @@ import { isTokenMalicious } from 'utils/isTokenMalicious'
...
 
 export function getLocalTokenId(
   token: TokenWithBalance | NetworkContractToken
 ): string {
+  if (token.type === TokenType.SPL) {
+    return `spl-${token.address}`
+  }
+  if (token.type === TokenType.NATIVE && token.symbol === 'SOL') {
+    return 'native-sol'
+  }
```
atn4z7, line 23: should we combine this logic with this one? also, let's use `_` instead of `-`. I think that is a bit safer. 
```   
    localId:  token.type === TokenType.SPL
                ? token.address
                : `SOL_${token.symbol}`,
```

packages/core-mobile/app/utils/caip2ChainIds.ts
```diff
@@ -137,38 +137,83 @@ export const getEvmCaip2ChainId = (chainId: number): string => {
...
+    return ChainId.SOLANA_DEVNET_ID
+  } else if (caip2ChainId === SolanaCaip2ChainId.TESTNET) {
+    return ChainId.SOLANA_TESTNET_ID
+  }
+
+  // For custom formats, return mainnet ID if it's a valid Solana address format
+  const [namespace, address] = caip2ChainId.split(':')
+  if (namespace === BlockchainNamespace.SOLANA && address) {
+    return ChainId.SOLANA_MAINNET_ID
+  }
```
atn4z7, line 166: just curious, when does this happen?

packages/core-mobile/app/utils/caip2ChainIds.ts
```diff
@@ -137,38 +137,83 @@ export const getEvmCaip2ChainId = (chainId: number): string => {
...
+  ChainId.AVALANCHE_TEST_P,
+  ChainId.AVALANCHE_X,
+  ChainId.AVALANCHE_TEST_X
+]
+
+const BITCOIN_CHAIN_IDS = [ChainId.BITCOIN, ChainId.BITCOIN_TESTNET]
+
+const SOLANA_CHAIN_IDS = [
+  ChainId.SOLANA_MAINNET_ID,
+  ChainId.SOLANA_DEVNET_ID,
```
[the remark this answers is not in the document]
B0Y3R-AVA: strangely we might not need test net for solana, but we do use dev net (we use it in the extension as well)

packages/core-mobile/app/utils/caip2ChainIds.ts
B0Y3R-AVA: had added it in as a fallback case when debugging some issues on balances. after removing it and doing some testing i dont believe we need this at all anymore

packages/core-mobile/app/store/balance/listeners.ts
B0Y3R-AVA: looking deeper into this, i'm not sure we need it at all.

we should be able to just use `getLocalTokenId` in the balances `listeners.ts::fetchBalanceForNetworks` and the usual fallback case should work fine for `getLocalTokenId`

atn4z7 (approved):

PR #2904, opened 2025-06-10
CP-10881: Fix unable to view token detail via price alert push notification 

atn4z7:
## Description

**Ticket: [CP-10881]** 

there are actually 2 issues that cause the infinite loading screen
1/ the track token details screen requires both a coingecko id and market type SEARCH to function correctly. we were passing only coingecko id from the deeplink
2/ we weren't handling undefined token in `getTokenAddress`, which makes `useTokenDetails` hook throw an error when it tries to calll `setTokenInfo`

this pr also fixes these issues:
- select token screen (in swap/send/bridge) doesn't handle long token name correctly
- select network screen in bridge shows "also supporting" evm networks

## Screenshots/Videos
https://github.com/user-attachments/assets/5f296108-35b2-4b43-940c-c1ade2751e8a

## Testing
please retest price alert push notification as well as checking all the select token/network screens

## Checklist

[CP-10881]: https://ava-labs.atlassian.net/browse/CP-10881?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core-mobile/app/contexts/DeeplinkContext/utils/handleDeeplink.ts
+++ b/packages/core-mobile/app/contexts/DeeplinkContext/utils/handleDeeplink.ts
@@ -8,6 +8,7 @@ import { showSnackbar } from 'new/common/utils/toast'
 import { router } from 'expo-router'
 import { History } from 'store/browser'
 import { navigateFromDeeplinkUrl } from 'utils/navigateFromDeeplink'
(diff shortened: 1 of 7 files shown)


ruijialin-avalabs (approved):
🔥

PR #2939, opened 2025-06-20
Derive missing Solana session keys for existing Seedless users

onghwan:
## Description

* Derive the missing session keys needed to generate Solana addresses for existing seedless users.


packages/core-mobile/app/store/account/listeners.ts
```diff
@@ -184,20 +189,31 @@ const migrateSolanaAddressesIfNeeded = async (
...
-
-  if (!hasSolanaAddressesMigrated) {
-    const accounts = selectAccounts(state)
-    const entries = Object.values(accounts)
-    if (entries.some(account => !account.addressSVM)) {
-      // reload only when there are accounts without Solana addresses
-      reloadAccounts(_action, listenerApi)
+  const accounts = selectAccounts(state)
+  const entries = Object.values(accounts)
+  if (entries.some(account => !account.addressSVM)) {
```
onghwan, line 196: Let me drop the `ViewOnceKey` logic here, since `refreshSeedlessSessionKeys` can also be called in initAccounts.
`wallet.deriveMissingKeys` can fail if users have a poor internet connection, so I think it’s safer to try deriving addresses for existing users when not all accounts have an addressSVM.

packages/core-mobile/app/seedless/services/wallet/SeedlessWallet.ts
```diff
@@ -168,6 +163,18 @@ export default class SeedlessWallet implements Wallet {
...
+  public async deriveMissingKeys(): Promise<void> {
+    const identityProof = await this.#client.apiClient.identityProve()
+    const mnemonicId = await this.getMnemonicId()
+
+    await CoreSeedlessAPIService.deriveMissingKeys({
+      identityProof,
+      mnemonicId
+    })
+
+    SeedlessService.invalidateSessionKeysCache()
```
atn4z7, line 175: I think this invalidateSessionKeysCache should belong to the refresh call

packages/core-mobile/app/store/account/listeners.ts
atn4z7: I'm cool with this but don't we only set ViewOnceKey.MIGRATE_SOLANA_ADDRESS to true once we have successfully migrated?

packages/core-mobile/app/seedless/services/SeedlessService.ts
```diff
@@ -122,6 +124,14 @@ class SeedlessService {
   invalidateSessionKeysCache(): void {
     this.sessionKeysListCache = undefined
   }
+
+  async refreshSessionKeys(): Promise<void> {
+    const allKeys = await this.getSessionKeysList()
+
+    const pubKeys = transformKeyInfosToPubKeys(allKeys)
+    Logger.info('saving public keys')
+    await SeedlessPubKeysStorage.save(pubKeys)
```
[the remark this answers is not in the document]
onghwan: good point. fixed.

packages/core-mobile/app/seedless/services/wallet/SeedlessWallet.ts
onghwan: good point. changed `invalidateSessionKeysCache` to private and fixed to call `refreshSessionKeys` instead of calling `invalidateSessionKeysCache`

packages/core-mobile/app/seedless/services/wallet/SeedlessWallet.ts
```diff
@@ -35,7 +35,6 @@ import { stripChainAddress } from 'store/account/utils'
 import { AddressPublicKey, Curve, isEvmPublicKey } from 'utils/publicKeys'
 import { findPublicKey } from 'utils/publicKeys'
 import { SeedlessPubKeysStorage } from 'seedless/services/storage/SeedlessPubKeysStorage'
-import { transformKeyInfosToPubKeys } from 'seedless/services/wallet/transformKeyInfosToPubkeys'
```
atn4z7, line 38: should we move this util to Seedless service folder now?

onghwan: moved.

atn4z7 (approved):

PR #2941, opened 2025-06-20
Import wallets UI

neven-s:
**Ticket: []** 

#### 1. Wallet Type Refactoring
Removed global `selectWalletType` selector from the `app` store
Replaced with wallet-specific type handling using `useActiveWallet` hook and wallet objects
Updated multiple components to get wallet type from the active wallet rather than global state
#### 2. Account Creation Improvements
Modified `addAccount` thunk to accept `walletType` as a parameter instead of reading from global state
Updated account creation logic to be wallet-specific, using `selectAccountsByWalletId` instead of all accounts
Added wallet type parameter to account creation calls in `useWallet.ts `and `importWallet.tsx`
#### 3. New Private Key Viewing Feature
Added two new screens:
`verifyPinForPrivateKey.tsx` - PIN/biometric verification before showing private key
`viewPrivateKey.tsx` - Display private key with copy functionality
Made `getSigningKey` method public in `MnemonicWallet.ts` to support private key extraction
Updated navigation routes to include the new private key screens

## Checklist


packages/core-mobile/app/new/routes/(signedIn)/(modals)/accountSettings/account.tsx
```diff
@@ -82,8 +85,13 @@ const AccountScreen = (): JSX.Element => {
...
 
   const renderFooter = useCallback(() => {
-    return <AccountButtons accountId={account?.id ?? ''} />
-  }, [account?.id])
+    return (
+      <AccountButtons
+        accountId={account?.id ?? ''}
+        walletType={wallet?.type ?? WalletType.UNSET}
+      />
+    )
```
atn4z7, line 93: can we just do an early null return for this whole component when account or wallet is not found/invalid?

packages/core-mobile/app/new/routes/(signedIn)/(modals)/accountSettings/verifyPinForPrivateKey.tsx
```diff
@@ -0,0 +1,69 @@
...
+const VerifyPinForPrivateKeyScreen = (): JSX.Element => {
+  const { replace } = useRouter()
+  const { accountIndex } = useLocalSearchParams<{ accountIndex: string }>()
+  const activeNetwork = useSelector(selectActiveNetwork)
+  const activeWallet = useActiveWallet()
+
+  const getPrivateKeyFromMnemonic = async (
+    mnemonic: string
+  ): Promise<string> => {
+    const wallet: MnemonicWallet = new MnemonicWallet(mnemonic)
```
atn4z7, line 22: let's use WalletFactory.createWallet here instead

atn4z7, line 22: this logic should live inside WalletService

packages/core-mobile/app/new/routes/(signedIn)/(modals)/accountSettings/account.tsx
neven-s: done in https://github.com/ava-labs/core-mobile/commit/420e26e5ce2c8d1265895b0e03f3c7f95278cebc

packages/core-mobile/app/new/routes/(signedIn)/(modals)/accountSettings/verifyPinForPrivateKey.tsx
neven-s: moved to https://github.com/ava-labs/core-mobile/commit/97efd6f3e71ed27ada2bec9d28dd88d866ff4380

PR #2960, opened 2025-06-24
CP-10837: Call refreshPublicKeys in the initAccounts listener

onghwan:
## Description

**Ticket: [CP-10837]** 

* Fixed it to call `refreshPublicKeys` in the `initAccounts` listener to ensure the Seedless wallet has public keys before creating accounts. We don’t need to call `refreshPublicKeys` every time the active Seedless wallet changes, since we support only one Seedless wallet and its keys should already be available from storage.

[CP-10837]: https://ava-labs.atlassian.net/browse/CP-10837?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


neven-s (approved):
Nice!

PR #2990, opened 2025-06-27
CP-11154 Fix double bio prompt

neven-s:
## Description

**Ticket: [CP-11154]** 

verifyBiometric now takes into account that if biometric migration happened and is successful that means that user has already provided valid bio data.

## Checklist

[CP-11154]: https://ava-labs.atlassian.net/browse/CP-11154?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core-mobile/app/new/common/hooks/usePinOrBiometryLogin.ts
+++ b/packages/core-mobile/app/new/common/hooks/usePinOrBiometryLogin.ts
@@ -46,6 +46,7 @@ export function usePinOrBiometryLogin({
     reset: resetRateLimiter,
     remainingSeconds
   } = useRateLimiter()
+  const [isVerifyingBiometric, setIsVerifyingBiometric] = useState(false)
 
   // get formatted time based on time ticker and rest interval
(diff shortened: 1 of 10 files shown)


packages/core-mobile/app/new/common/hooks/usePinOrBiometryLogin.ts
```diff
@@ -156,12 +156,27 @@ export function usePinOrBiometryLogin({
...
+            //already prompted user for bio, assume verified
+            setVerified(true)
+            resetRateLimiter()
+            return new NothingToLoad()
+          }
+          if (result.success) {
+            throw new Error(
+              'Invalid state: migration status is not RunBiometricMigration'
+            )
+          }
```
atn4z7, line 175: do we actually want to throw an error for any status here? what if the status is no migration needed?

PR #3014, opened 2025-07-03
CP-10893 Token Activity Screen - Load the screen first then populate the history

bogdandobritoiu:
## Description

**Ticket: [CP-10893]** 

Motivation
We are displaying a fullscreen spinner when the screen opens. For a faster UX experience we should load the data we already have and then add a spinner for the rest. After the two requests for tokenInfo and chartData finish we can display them separately and with an animation.

**TokenDetailsScreen**
- pass through params the entire token data for name/symbol/price/percentage when opening the screen
- animate the chart when loading finishes and data is initialised
- animate the about section and the rest of the information after it with a delay between them
- animate the bottom button
- display token name if it's not the same as the token symbol (AAVE has the same name, BTC has Bitcoin as name)
- fix share button and favorites button to match design + press area

**TokenShareScreen**
- pass through params the entire token data for name/symbol/price/percentage when opening the screen
- remove spinner for the whole screen

## Screenshots/Videos

Before

https://github.com/user-attachments/assets/96b44d22-ec7b-4736-a29d-3048a29941e3

After

https://github.com/user-attachments/assets/15c8701c-8499-474e-8739-043753ee0017

## Checklist


packages/core-mobile/app/new/features/track/screens/TrackTokenDetailScreen.tsx
```diff
@@ -94,12 +98,38 @@ const TrackTokenDetailScreen = (): JSX.Element => {
...
+  )
+
+  const handleSwap = useCallback(
+    (initialTokenIdTo?: string): void => {
+      navigateToSwap(AVAX_TOKEN_ID, initialTokenIdTo)
+    },
+    [navigateToSwap]
+  )
+
+  const { actions } = useTrackTokenActions({
```
bogdandobritoiu, line 123: Had to delete the component TokenDetailsFooter and move it to a hook to be able to return null in renderFooter. I needed the actions.length to hide footer.

ScrollScreen's renderFooter needs to return null and not an empty react element (object) to be able to hide it's UI. Which made this screen cut at the bottom when there were no actions.

packages/core-mobile/app/new/routes/(signedIn)/(modals)/accountSettings/securityAndPrivacy.tsx
```diff
@@ -97,7 +97,7 @@ const SecurityAndPrivacyScreen = (): JSX.Element => {
   const pinAndBiometricData = useMemo(() => {
     const data: GroupListItem[] = [
       {
-        title: 'Require PIN code immediately',
+        title: 'Require PIN immediately',
```
bogdandobritoiu, line 100: small copy change requested from design

packages/core-mobile/app/new/features/track/components/ShareChart.tsx
```diff
@@ -9,97 +9,105 @@ import {
...
+  const range = useMemo(() => {
+    return {
+      diffValue: ranges.diffValue
+        ? ranges.diffValue
+        : token.priceChange24h ?? 0,
+      percentChange: ranges.percentChange
+        ? ranges.percentChange
+        : token.priceChangePercentage24h ?? 0
+    }
+  }, [ranges, token.priceChange24h, token.priceChangePercentage24h])
```
bogdandobritoiu, line 58: Loads prices based off route params data and then updates them when tokenInfo and ranges are available

packages/core-mobile/app/new/features/track/market/components/MarketTokensScreen.tsx
```diff
@@ -22,7 +22,7 @@ const MarketTokensScreen = ({
   charts: Charts
   sort: DropdownSelection
   view: DropdownSelection
-  goToMarketDetail: (tokenId: string, marketType: MarketType) => void
+  goToMarketDetail: (token: MarketToken) => void
```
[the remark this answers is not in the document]
atn4z7: I believe if you split up token info, you should be able to achieve the same result.

PR #3015, opened 2025-07-03
CP-11338: Fix to prompt bio auth before setting wallet name

onghwan:
## Description

**Ticket: [CP-11338]** 

* Fix the onboarding flow to prompt for biometric authentication before naming the wallet, returning the user to the mnemonic setup step if Touch ID unlock (when enabled) fails.
* Fix an onboarding issue where navigating back from the wallet naming step to PIN setup could create multiple empty wallets by using activeWalletId when available.

[CP-11338]: https://ava-labs.atlassian.net/browse/CP-11338?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core-mobile/app/hooks/useWallet.test.ts
+++ b/packages/core-mobile/app/hooks/useWallet.test.ts
@@ -106,7 +106,7 @@ describe('useWallet', () => {
 
       expect(BiometricsSDK.storeWalletSecret).toHaveBeenCalledWith(
         mockWalletId,
-        mockMnemonic
+        mnemonic
       )
       expect(response).toBe(mockWalletId)
     })


packages/core-mobile/app/new/routes/onboarding/mnemonic/createPin.tsx
```diff
@@ -22,22 +33,36 @@ export default function CreatePin(): JSX.Element {
       }
       AnalyticsService.capture('OnboardingPasswordSet')
       onPinCreated({
+        walletId: activeWalletId ?? uuid(),
```
onghwan, line 36: There was a bug where going back to the PIN setting from the wallet name step during onboarding could result in creating multiple wallets. This could eventually allow users to end up with multiple empty wallets. I fixed this by using activeWalletId when it’s available.

ruijialin-avalabs (approved):

PR #3030, opened 2025-07-08
CP-11396: Fix swap percentage option bug

onghwan:
## Description

**Ticket: [CP-11396]** 

* Fix bug where selecting swap options was unresponsive
* Fix bug where swap option could be set larger than available balance.

[CP-11396]: https://ava-labs.atlassian.net/browse/CP-11396?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core-mobile/app/new/common/components/TokenInputWidget.tsx
+++ b/packages/core-mobile/app/new/common/components/TokenInputWidget.tsx
@@ -22,6 +22,7 @@ import Animated, {
   FadeOut,
   LinearTransition
 } from 'react-native-reanimated'
+import { roundToUnitOrSignificantDigit } from 'common/utils/roundToUnitOrSignificantDigit'
 import { LogoWithNetwork } from './LogoWithNetwork'
 
 export const TokenInputWidget = ({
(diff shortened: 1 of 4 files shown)


packages/core-mobile/app/new/common/components/TokenInputWidget.tsx
```diff
@@ -141,20 +144,6 @@ export const TokenInputWidget = ({
...
-    if (selectedButton && token?.decimals && amount) {
-      const unit = new TokenUnit(amount, token?.decimals, token?.symbol)
-      const display = unit.toDisplay({ asNumber: true })
-      const multiplier = 10 ** token.decimals
-
-      return BigInt(Math.round(display * multiplier))
-    }
-
-    return amount
-  }, [amount, percentageButtons, token?.decimals, token?.symbol])
```
onghwan, line 156: If a value other than amount (e.g., `inputValue`) is used for the `TextInput`, the value displayed in the field may differ from the actual swap amount. Therefore, if you want to enforce decimal precision when using the percentage buttons, you should modify amount directly.

PR #3034, opened 2025-07-08
CP-11327: Better error messages for Solana transaction failures

B0Y3R-AVA:
## Description

**Ticket: [CP-11327](https://ava-labs.atlassian.net/browse/CP-11327)** 

- Added handling for solana specific error codes that come through in messages of rpcErrors from VM module 
- List of error codes handled found [here](https://www.quicknode.com/docs/solana/error-references)
- Unit tests for new error handling logic 

## Checklist

### DIFF
--- a/packages/core-mobile/app/utils/getJsonRpcErrorMessage/getJsonRpcErrorMessage.ts
+++ b/packages/core-mobile/app/utils/getJsonRpcErrorMessage/getJsonRpcErrorMessage.ts
@@ -19,6 +19,57 @@ export const getJsonRpcErrorMessage = (error: unknown): string => {
   return 'Unexpected error'
 }
 
+function getSolanaErrorMessage(errorMessage: string): string | null {
+  if (errorMessage.includes('-32002')) {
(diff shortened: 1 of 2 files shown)


packages/core-mobile/app/utils/getJsonRpcErrorMessage/getJsonRpcErrorMessage.ts
```diff
@@ -34,6 +85,13 @@ export const parseJsonRpcError = (
 
   // Handle detailed error message based on cause
   if (cause?.error?.message) {
+    // Try to get Solana specific error message
+    const solanaError = getSolanaErrorMessage(cause.error.message)
+    if (solanaError) {
+      return solanaError
+    }
```
atn4z7, line 92: we should probably start thinking about how to handle rpc error message better since what we have right now is not really scalable. maybe we can include the chain id + action (send/swap/etc) and narrow down the logic accordingly.

atn4z7 (approved):

PR #3035, opened 2025-07-09
Boyer/cp 10880

B0Y3R-AVA:
## Description

**Ticket: []** 

## Checklist


packages/core-mobile/app/seedless/services/wallet/SeedlessWallet.ts
```diff
@@ -509,4 +515,21 @@ export default class SeedlessWallet implements Wallet {
...
+     * properly by the dApps. I think it's because the dApps provide a UTF-8 message,
+     * but for it to be accepted by Solana Ledger app, we need to serialize it,
+     * add a message header etc., and I think Ledger then signs the whole thing, which
+     * makes it impossible to verify the signature with the original message.
+     */
+    throw new Error(
+      'Signing off-chain messages is only supported with seedphrase wallets at the moment'
+    )
+  }
 }
```
onghwan, line 535: It’d be great to have `signEvmMessage`, similar to how you refactored it in MnemonicWallet.

packages/core-mobile/app/services/wallet/MnemonicWallet.ts
```diff
@@ -269,8 +231,14 @@ export class MnemonicWallet implements Wallet {
   }: {
     accountIndex: number
     network: Network
-    provider: JsonRpcBatchInternal
+    provider: JsonRpcBatchInternal | Avalanche.JsonRpcProvider | SolanaProvider
```
onghwan, line 234: What’s the reason for this change? From what I see, `getSigningKey` is only used with the EVM provider.

packages/core-mobile/app/services/wallet/PrivateKeyWallet.ts
```diff
@@ -388,4 +410,13 @@ export class PrivateKeyWallet implements Wallet {
...
+
+  private async signSolanaMessage(
+    data: string,
+    _accountIndex: number
+  ): Promise<string> {
+    if (typeof data !== 'string') throw new Error('data must be string')
+    const signer = await this.getSvmSigner()
+    return await signer.signMessage(data)
+  }
 }
```
onghwan, line 422: Same here. Having `signEvmMessage` would be great.

onghwan (commented):
🎉

packages/core-mobile/app/services/walletconnectv2/solanaRequestUtils.ts
```diff
@@ -0,0 +1,112 @@
...
+  const { method, params } = requestEvent.params.request
+
+  // Solana dApps use different parameter formats than our internal VM module:
+  // 1. Different dApps (Jupiter, Orca) structure their parameters differently
+  // 2. Encoding differences: dApps use base58, our VM uses base64
+  // 3. Parameter shape: dApps use {pubkey, message} format, VM expects [{account, serializedMessage}]
+
+  switch (method) {
+    // Handle Solana message signing
+    case 'solana_signMessage': {
```
atn4z7, line 86: could we use the rpc method enums here?

packages/core-mobile/app/services/walletconnectv2/WalletConnectService.ts
```diff
@@ -73,8 +74,18 @@ class WalletConnectService implements WalletConnectServiceInterface {
...
-      requestSession &&
+      if (requestSession) {
+        /**
+         * Solana dApps use different parameter formats than our internal
+         * VM module so we need to transform them, @see solanaRequestUtils.ts
+         * for more details
+         */
+        if (requestEvent.params.request.method.includes('solana_')) {
+          transformSolanaParams(requestEvent, requestSession)
+        }
```
atn4z7, line 85: I don't think we should do this transform here since wallet connect service doesn't/shouldn't know about the vm module stuff. it would be great if this is done in the rpc listeners instead before the request is passed to vm module.

packages/core-mobile/app/services/walletconnectv2/solanaRequestUtils.ts
```diff
@@ -0,0 +1,112 @@
...
+        params,
+        requestSession
+      )
+      if (transformedParams) {
+        requestEvent.params.request.params = transformedParams
+      }
+      break
+    }
+  }
+}
```
atn4z7, line 112: nice utils and tests!!

packages/core-mobile/app/services/wallet/MnemonicWallet.ts
```diff
@@ -269,8 +231,14 @@ export class MnemonicWallet implements Wallet {
   }: {
     accountIndex: number
     network: Network
-    provider: JsonRpcBatchInternal
+    provider: JsonRpcBatchInternal | Avalanche.JsonRpcProvider
```
onghwan, line 234:
```suggestion
    provider: JsonRpcBatchInternal
```
You can say:

⸻
Do we still need `Avalanche.JsonRpcProvider` here? If not, I don’t think the `JsonRpcBatchInternal` type guard below is necessary either.

atn4z7 (approved):

PR #3045, opened 2025-07-13
CP-11322: Show Solana SPL token activity on Token Details 

B0Y3R-AVA:
…fore mapping to vm

## Description

**Ticket: [CP-11322](https://ava-labs.atlassian.net/browse/CP-11322)** 

## Screenshots/Videos
Screenshot 2025-07-13 at 5 33 21 PM

Screenshot 2025-07-13 at 5 33 40 PM

https://github.com/user-attachments/assets/c173cdcc-6c84-4b10-94c8-92a220c4dd9b

## Testing
Manually Tested with XT
- Imported private key from Mobile Wallet into XT so I could confirm matching activity on both, as seen above in the videos. 

## Checklist

[CP-11322]: https://ava-labs.atlassian.net/browse/CP-11322?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core-mobile/app/hooks/networks/useTokenList.ts
+++ b/packages/core-mobile/app/hooks/networks/useTokenList.ts
@@ -1,0 +1,32 @@
+import { UseQueryResult, useQuery } from '@tanstack/react-query'
+import Config from 'react-native-config'
+import Logger from 'utils/Logger'
+
+const QUERY_KEY = ['tokenList', 'solana']
+
+const HOUR_IN_MS = 1000 * 60 * 60
+const DAY_IN_MS = 1000 * 60 * 60 * 24
+
+export const useTokenList = (
+  chainId: string
+): UseQueryResult<string[], Error> => {
+  return useQuery({
+    queryKey: [...QUERY_KEY, chainId],
+    queryFn: async () => {
+      try {
+        const response = await fetch(
+          `${Config.PROXY_URL}/tokenlist?includeSolana`
+        )
+        const data = await response.json()
+        return data[chainId]?.tokens || []
+      } catch (error) {
+        Logger.error('Failed to fetch SPL token metadata', error)
+        return []
+      }
+    },
+    // Cache for 1 hour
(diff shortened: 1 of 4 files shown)


packages/core-mobile/app/hooks/networks/useTokenList.ts
```diff
@@ -0,0 +1,32 @@
...
+        const response = await fetch(
+          `${Config.PROXY_URL}/tokenlist?includeSolana`
+        )
+        const data = await response.json()
+        return data[chainId]?.tokens || []
+      } catch (error) {
+        Logger.error('Failed to fetch SPL token metadata', error)
+        return []
+      }
+    },
```
atn4z7, line 26: the query function here should return the raw response. you can use the select function to transform the response after (if you are using it in a react component). that way we only cache 1 response for all chain ids.

packages/core-mobile/app/services/activity/ActivityService.tsx
```diff
@@ -30,6 +40,58 @@ export class ActivityService {
...
+        tokens: cachedData
+      }
+    }
+
+    // If no cached data, fetch and cache
+    try {
+      const tokenListResponse = await fetch(
+        `${Config.PROXY_URL}/tokenlist?includeSolana`
+      )
+      const tokenData = await tokenListResponse.json()
```
atn4z7, line 77: looks like the query is never executed since fetch is still being used. you can do something like this:
```
export const getCachedTokenList = async (chainId: string) => {
  return queryClient.fetchQuery({
    queryKey: ['tokenList', 'solana'],
    queryFn: fetchTokenListWithSolana,
    staleTime,
    gcTime,
    select: (data) => data[chainId]?.tokens ?? []
  })
}

getCachedTokenList()
```

atn4z7 (approved):

PR #3094, opened 2025-07-23
Fix for child press event on AnimatedPressable

bogdandobritoiu:
## Description

I created a new hook `usePreventParentPress` to control which event is fired. 

I've also removed unused hooks useKeyboardOpen useKeyboardHeight because we use react-native-keyboard-controller for that.

## Screenshots/Videos

https://github.com/user-attachments/assets/15b6c19c-9819-4002-9e8e-f070e3049581

## Checklist


packages/core-mobile/app/new/features/accountSettings/components/AccountItem.tsx
```diff
@@ -50,6 +51,13 @@ export const AccountItem = memo(
     } = useTheme()
     const { formatCurrency } = useFormatCurrency()
 
+    const { handleParentPress, createChildPressHandler } =
+      usePreventParentPress(() => onSelectAccount(account))
+
+    const handleChildPress = createChildPressHandler(() =>
+      gotoAccountDetails(account.id)
+    )
+
```
atn4z7, line 60: I think it's better to also have a createParentPressHandler instead of passing the parent press logic as a hook param

atn4z7 (approved):

PR #3131, opened 2025-08-07
CP-11751 - GA - Unable to display `swapped` when swapping ERC20 to AVAX

bogdandobritoiu:
## Description

**Ticket: [CP-11751]** 

When a tx method is Unknown we have to check if from and to is equal to the tx.token[0] from and to in case we only have a tx.token.length of 1. 

This would be considered a swap tx. Previously this check was displaying "sent" instead of "swapped for"

## Screenshots/Videos
Before
Simulator Screenshot - iPhone 16 Pro - 2025-08-07 at 16 12 52

After
Simulator Screenshot - iPhone 16 Pro - 2025-08-07 at 16 12 44

## Testing

Swap from any token to AVAX. 

## Checklist

### DIFF
--- a/packages/core-mobile/app/new/features/activity/utils.ts
+++ b/packages/core-mobile/app/new/features/activity/utils.ts
@@ -129,3 +129,13 @@ export function isNftTransaction(tx: TokenActivityTransaction): boolean {
 export function isSupportedNftChainId(chainId: number): boolean {
   return isAvalancheCChainId(chainId) || isEthereumChainId(chainId)
 }
+
+// To differentiate between swap and sent/received
+// we need to check if the from and to addresses are the same
+// as the from and to addresses of the token
+export function isUnknownSwap(tx: TokenActivityTransaction): boolean {
+  return (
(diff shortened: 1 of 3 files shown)


packages/core-mobile/app/new/features/activity/utils.ts
```diff
@@ -129,3 +129,13 @@ export function isNftTransaction(tx: TokenActivityTransaction): boolean {
 export function isSupportedNftChainId(chainId: number): boolean {
   return isAvalancheCChainId(chainId) || isEthereumChainId(chainId)
 }
+
+// To differentiate between swap and sent/received
+// we need to check if the from and to addresses are the same
+// as the from and to addresses of the token
+export function isUnknownSwap(tx: TokenActivityTransaction): boolean {
```
onghwan, line 136:
```suggestion
export function isSwapTransaction(tx: TokenActivityTransaction): boolean {
```
It might not be necessary to include “unknown” in the function name.

onghwan (approved):

PR #3132, opened 2025-08-07
CP-11712: Add unique user id for troubleshooting

onghwan:
## Description

**Ticket: [CP-11712]** 

* Add unique user ID to help us troubleshoot more efficiently

## Screenshots/Videos
* iOS
https://github.com/user-attachments/assets/a0191d15-781a-4406-ac20-ed0b8b4e1b73

* Android
https://github.com/user-attachments/assets/6a0d680a-c8e6-42da-bfaf-de5944504a6e

## Testing
* If you’d like to check the unique user ID in the `FullScreenWarning` component, you should trigger an error that falls back to the Sentry error boundary. The easiest way is to throw an error from within a React component during render. For example:
```
const Component = (): JSX.Element => {
  throw new Error("Test");
};
```
* To fall back to `TopLevelErrorFallback` in the local build, follow the steps below.

index.js
```
if (__DEV__) {
  // ...
  
  DevDebuggingConfig.LOGBOX_DISABLED && LogBox.ignoreAllLogs(true)

  // Add these three lines
  ErrorUtils.setGlobalHandler((error, isFatal) => {
    DefaultErrorReporter(error, isFatal)
  })

   // ...
}
```

DevDebuggingConfig.ts
```
const DevDebuggingConfig = {
  WDYR: false,
  STORYBOOK_ENABLED: false,
  LOGBOX_DISABLED: true, // <= set to true
  SHOW_DEMO_NFTS: false,
  API_MOCKING: false,
  SENTRY_SPOTLIGHT: true,  // <= set to true
  METRO_DEV_MENU: false
}
```

[CP-11712]: https://ava-labs.atlassian.net/browse/CP-11712?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core-mobile/app/services/user/UserService.ts
+++ b/packages/core-mobile/app/services/user/UserService.ts
@@ -1,0 +1,23 @@
+import { MMKV } from 'react-native-mmkv'
+import { uuid } from 'utils/uuid'
+
+const storage = new MMKV({
+  id: `user`
+})
+
+class UserService {
+  static getUniqueID(): string {
+    const UNIQUE_ID_KEY = 'USER_SERVICE_UNIQUE_ID'
+
+    let id = storage.getString(UNIQUE_ID_KEY)
+
+    if (!id) {
+      id = uuid()
+      storage.set(UNIQUE_ID_KEY, id)
+    }
+
+    return id
+  }
+}
+
+export default UserService
--- a/packages/core-mobile/app/services/user/UserService.ts
+++ b/packages/core-mobile/app/services/user/UserService.ts
@@ -1,19 +1,14 @@
-import { MMKV } from 'react-native-mmkv'
+import { StorageKey } from 'resources/Constants'
+import { commonStorage } from 'utils/mmkv'
 import { uuid } from 'utils/uuid'
-
-const storage = new MMKV({
-  id: `user`
-})
 
 class UserService {
   static getUniqueID(): string {
-    const UNIQUE_ID_KEY = 'USER_SERVICE_UNIQUE_ID'
-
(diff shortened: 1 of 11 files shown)


packages/core-mobile/app/services/user/UserService.ts
```diff
@@ -0,0 +1,23 @@
...
+    const UNIQUE_ID_KEY = 'USER_SERVICE_UNIQUE_ID'
+
+    let id = storage.getString(UNIQUE_ID_KEY)
+
+    if (!id) {
+      id = uuid()
+      storage.set(UNIQUE_ID_KEY, id)
+    }
+
+    return id
```
onghwan, line 19: Unlike what we do in PostHog, I didn’t store this unique ID in Redux, because an error could occur before rehydration.

B0Y3R-AVA (approved):
Nice Work!

packages/core-mobile/app/services/user/UserService.ts
```diff
@@ -0,0 +1,23 @@
+import { MMKV } from 'react-native-mmkv'
+import { uuid } from 'utils/uuid'
+
+const storage = new MMKV({
+  id: `user`
+})
+
+class UserService {
```
atn4z7, line 8: let's also clear this storage on logout

onghwan: fixed to use `commonStorage`, which already gets cleared on logout.

atn4z7 (approved):

PR #3145, opened 2025-08-12
CP-11169: don't show balance-transferred notifications while foregrounded

onghwan:
## Description

**Ticket: [CP-11169]** 

* Fix to stop showing `balance-transferred` type notifications when the app is in the foreground

[CP-11169]: https://ava-labs.atlassian.net/browse/CP-11169?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core-mobile/app/services/fcm/FCMService.ts
+++ b/packages/core-mobile/app/services/fcm/FCMService.ts
@@ -65,8 +65,11 @@ class FCMService {
         return
       }
 
-      if (result.data.data.event === BalanceChangeEvents.BALANCES_SPENT) {
-        // skip showing notification if user just spent balance in app
+      if (
+        result.data.data.event === BalanceChangeEvents.BALANCES_SPENT ||
+        result.data.data.event === BalanceChangeEvents.BALANCES_TRANSFERRED
+      ) {
+        // skip showing notification if user just spent balance in app, or transferred balance between user's own accounts
         return
       }
 


packages/core-mobile/app/services/fcm/FCMService.ts
```diff
@@ -65,8 +65,11 @@ class FCMService {
         return
       }
 
-      if (result.data.data.event === BalanceChangeEvents.BALANCES_SPENT) {
-        // skip showing notification if user just spent balance in app
+      if (
+        result.data.data.event === BalanceChangeEvents.BALANCES_SPENT ||
+        result.data.data.event === BalanceChangeEvents.BALANCES_TRANSFERRED
+      ) {
+        // skip showing notification if user just spent balance in app, or transferred balance between user's own accounts
```
onghwan, line 72: 1. We already have logic in place to ignore `BALANCES_SPENT` notifications when the app is in the foreground. This was developed about 11 months ago during the initial notification implementation.
2. About six months ago, the `BALANCES_TRANSFERRED` notification type was added to the `core-notification-sender-service`. This notification is sent when a user transfers tokens between their own accounts. Presumably, this was to prevent users from receiving both `BALANCES_SPENT` and `BALANCES_RECEIVED` notifications for the same transfer.
3. Therefore, for the same reason as `BALANCES_SPENT`, we can also ignore `BALANCES_TRANSFERRED` notifications when in the foreground.

PR #3162, opened 2025-08-19
Stake - stop polling if operation result is `Dropped` on retry function

bogdandobritoiu:
## Description

Stop polling if result for getAtomicTxStatus is `Dropped` for `importC` and `exportC`

### DIFF
--- a/packages/core-mobile/app/utils/js/retry.test.ts
+++ b/packages/core-mobile/app/utils/js/retry.test.ts
@@ -69,7 +69,7 @@ describe('failure case', () => {
     expect(getAndIncrementCounter).toHaveBeenCalledTimes(2)
   })
 
-  it('should stop polling when isStopping is true', async () => {
+  it('should stop retrying when shouldStop is true', async () => {
     const getAndIncrementCounter = jest.fn().mockImplementation(() => {
       return 0
     })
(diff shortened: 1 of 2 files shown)


packages/core-mobile/app/utils/js/retry.ts
```diff
@@ -14,7 +14,7 @@ type RetryParams<T> = {
  * Retries an operation with defined backoff policy.
  *
  * @param operation - The operation to retry.
- * @param isSuccess - The predicate to check if the operation succeeded.
+ * @param shouldStop - The predicate to check if the operation should stop polling.
```
onghwan, line 17:
```suggestion
 * @param shouldStop - The predicate to check if the operation should stop retrying.
```

packages/core-mobile/app/utils/js/retry.test.ts
```diff
@@ -68,4 +68,17 @@ describe('failure case', () => {
 
     expect(getAndIncrementCounter).toHaveBeenCalledTimes(2)
   })
+
+  it('should stop polling when isStopping is true', async () => {
```
onghwan, line 72:
```suggestion
  it('should stop retrying when isStopping is true', async () => {
```
nit: let's keep the wording consistent.

onghwan (approved):

PR #3220, opened 2025-09-17
CP-12093 - Do not show keyboard during the swap

bogdandobritoiu:
## Description

**Ticket: [CP-12093](https://ava-labs.atlassian.net/browse/CP-12093)**

Please provide:

- A summary of the changes
- Motivation and context for this change
- Any dependencies introduced
- (If breaking) migration steps and instructions

Android and iOS native screens have different keyboard interactions when navigating:
- iOS has multiple  the keyboard and hides it you open another formsheet, 
- Android has a single keyboard on all screens so it will be on top of the next formsheet opened. 

**Because of this behavior we are forced to control the keyboard dismissal only for Android, IOS handles this as expected.**

Another interaction that is different is when clicking outside of the Scrollview when keyboard is up. It won't close the it on Android, iOS does this as expected.

**Because of this behavior the renderFooter which is placed outside of the scrollview won't close the keyboard when pressed.**

**So in order to control this for Android, when we press the Next button it will first close the keyboard and then navigate/show error.**
In case we get an error I think it's best for the user to see the error instead of scrolling because the keyboard is up.

## Screenshots/Videos

Before
https://github.com/user-attachments/assets/eaee0427-d4eb-449d-9166-181116ab721d

After
https://github.com/user-attachments/assets/3046d061-8ebf-4453-90bb-7ffd2b2f5114

## Testing

- Provide steps to test the happy path of your feature
- Provide steps to test edge cases and error states
- Trigger a build on bitrise and reference it here

Build number
iOS - 6116
Android - 6117

Any Android device

1. Open swap, press next and verify if keyboard got closed before navigating
2. Open swap, press AVAX token to select, verify keyboard is closed when navigating
3. Should happen the same on Send/Bridge/Buy/Withdraw

## Checklist

[CP-12093]: https://ava-labs.atlassian.net/browse/CP-12093?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/core-mobile/app/new/features/send/components/SendToken.tsx
```diff
@@ -176,6 +177,17 @@ export const SendToken = ({ onSend }: { onSend: () => void }): JSX.Element => {
     return loadAvatar(recipient?.avatar)
   }, [recipient?.avatar])
 
+  const onSubmit = useCallback(() => {
+    // (Android) native screens need to dismiss the keyboard before navigating
+    // the footer is outside of the scrollview that controls keyboardShouldPersistTaps
+    // so on Android we need to dismiss it before navigating
+    if (Platform.OS === 'android' && Keyboard.isVisible()) {
+      Keyboard.dismiss()
+    }
```
onghwan, line 186: can we do this on unmount?

onghwan: If that’s what you meant, it might be better to call `Keyboard.dismiss()` before calling navigation functions.(navigate, back, etc)

And could you also make a util function like this and use it in all the call sites?
```
export const dismissKeyboardIfNeeded = (): void => {
  if (Platform.OS === 'android' && Keyboard.isVisible()) {
    Keyboard.dismiss()
  }
}
```

packages/core-mobile/app/new/common/screens/SelectTokenScreen.tsx
```diff
@@ -96,6 +97,13 @@ export const SelectTokenScreen = <T extends object>({
     }
   }, [networkChainId, networks])
 
+  useEffect(() => {
+    // (Android) native screens need to dismiss the keyboard before navigating
+    // the footer is outside of the scrollview that controls keyboardShouldPersistTaps
+    // so on Android we need to dismiss it before navigating
```
onghwan, line 103: can we move this comment inside the `dismissKeyboardIfNeeded` function? Since the function is already checking `Platform.OS` I think it makes sense to move it there.

onghwan (approved):
Thanks for the fix!

B0Y3R-AVA (approved):

PR #3767, opened 2026-04-20
CP-14069: Fix non-dismissible Ledger scan error alert

ruijialin-avalabs:
## Description

iOS: 8216
Android: 8217

**Ticket: [CP-14069]**

`LedgerService` was calling `Alert.alert` directly with no buttons array.
On Android, an alert with no buttons renders without any dismiss control,
forcing the user to kill and reopen the app. The scan timeout path also
had an OK button but no `onPress` action, leaving state uncleaned.

**Root cause:** A service class showed UI alerts it had no ability to
navigate away from or clean up after.

**Fix:**
- Added `scanErrorListeners` to `LedgerService` (mirrors the existing
  `deviceListeners` pattern) with `addScanErrorListener`,
  `removeScanErrorListener`, and `notifyScanError`
- Replaced both `Alert.alert` calls in `LedgerService` with
  `notifyScanError` — the service now emits the error, not the UI
- Removed `Alert` import from `LedgerService`
- `DeviceConnectionScreen` subscribes via `useEffect`; the alert OK
  button calls `resetSetup()` to clean up scan state on dismiss

## Screenshots/Videos

N/A — logic-only change.

## Testing

- Start Ledger onboarding with no Ledger device nearby
- Wait for the scan timeout (~30 seconds) — confirm the "Scan failed"
  alert appears with an OK button that dismisses and resets state
- Force a BLE scan error (e.g. toggle Bluetooth mid-scan) — confirm the
  "Scan Error" alert is dismissible and cleans up state
- Verify the above on both iOS and Android (Android was the platform
  where the non-dismissible alert occurred)

- Scan for Ledger with no device in range → alert dismisses cleanly,
  user can retry or cancel without killing the app

## Checklist

[CP-14069]: https://ava-labs.atlassian.net/browse/CP-14069?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core-mobile/app/services/ledger/LedgerService.ts
+++ b/packages/core-mobile/app/services/ledger/LedgerService.ts
@@ -99,9 +99,6 @@ class LedgerService {
   private scanSubscription: { unsubscribe: () => void } | null = null
   private scanInterval: ReturnType<typeof setInterval> | null = null
   private deviceListeners: Set<(devices: LedgerDevice[]) => void> = new Set()
-  private scanErrorListeners: Set<
-    (error: { title: string; message: string }) => void
-  > = new Set()
   private currentDevices: LedgerDevice[] = []
   private isScanning = false
 
@@ -269,7 +266,10 @@ class LedgerService {
   }
 
   // Handle scan errors (matching original implementation)
-  private handleScanError(error: Error): void {
+  private handleScanError(
+    error: Error,
+    onScanError: (error: { title: string; message: string }) => void
+  ): void {
     Logger.error('Scan error:', error)
     this.stopDeviceScanning()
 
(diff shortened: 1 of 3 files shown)


packages/core-mobile/app/services/ledger/LedgerService.ts
```diff
@@ -99,6 +99,9 @@ class LedgerService {
   private scanSubscription: { unsubscribe: () => void } | null = null
   private scanInterval: ReturnType<typeof setInterval> | null = null
   private deviceListeners: Set<(devices: LedgerDevice[]) => void> = new Set()
+  private scanErrorListeners: Set<
+    (error: { title: string; message: string }) => void
+  > = new Set()
```
atn4z7: will we actually have a lot of listeners? wonder if we could just add a callback param to `startDeviceScanning`

ruijia1in: we shouldn't have more listeners.

ruijia1in: updated to use callback

atn4z7 (dismissed):

atn4z7 (dismissed):

B0Y3R-AVA (dismissed):

atn4z7 (approved):

B0Y3R-AVA (approved):

PR #3768, opened 2026-04-20
CP-14074: Only derive missing seedless session keys when seedless wallet accounts are missing addressSVM

ruijialin-avalabs:
## Description

**Ticket: [CP-14074]**

- Adds a guard in `migrateSolanaAddressesIfNeeded` so that `deriveMissingSeedlessSessionKeys` is only called when the seedless wallet's own accounts are actually missing `addressSVM`.
- Previously, the function would call `deriveMissingSeedlessSessionKeys` whenever *any* account across all wallets was missing a Solana address — even if those accounts belonged to a non-seedless wallet. This narrows the check to only accounts belonging to the seedless wallet.

## Screenshots/Videos

N/A — logic-only change, no UI impact.

## Testing

- Have a seedless wallet and a non-seedless wallet (e.g. Ledger or mnemonic) in the app
- Ensure the non-seedless wallet has an account missing `addressSVM`
- Confirm that `deriveMissingSeedlessSessionKeys` is **not** called (no unnecessary key derivation triggered)
- Ensure the seedless wallet still derives missing session keys when its own accounts are missing `addressSVM`
- Trigger a build on Bitrise and reference it here

iOS: 8217
Android: 8218

## Checklist

[CP-14074]: https://ava-labs.atlassian.net/browse/CP-14074?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core-mobile/app/store/account/listeners.ts
+++ b/packages/core-mobile/app/store/account/listeners.ts
@@ -297,17 +297,17 @@ const migrateSolanaAddressesIfNeeded = async (
   const isSolanaSupportBlocked = selectIsSolanaSupportBlocked(state)
   const accounts = selectAccounts(state)
   const entries = Object.values(accounts)
+  const seedlessWallet = selectSeedlessWallet(state)
+  const hasAccountsWithoutSVM = entries.some(account => !account.addressSVM)
+  const hasSeedlessAccountsWithoutSVM = entries.some(
+    account =>
+      account.addressSVM === undefined &&
+      account.walletId === seedlessWallet?.id
+  )
 
   // Only migrate Solana addresses if Solana support is enabled
-  if (!isSolanaSupportBlocked && entries.some(account => !account.addressSVM)) {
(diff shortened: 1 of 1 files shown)


B0Y3R-AVA (dismissed):

packages/core-mobile/app/store/account/listeners.ts
```diff
@@ -301,7 +301,13 @@ const migrateSolanaAddressesIfNeeded = async (
   // Only migrate Solana addresses if Solana support is enabled
   if (!isSolanaSupportBlocked && entries.some(account => !account.addressSVM)) {
```
atn4z7: should we just adjust entries to be accounts of seedless wallet instead? feel like checking two different account lists is a bit redundant

atn4z7 (approved):

B0Y3R-AVA (approved):
dev tested

PR #3782, opened 2026-04-23
CP-14082: Move chain ID fields to plaintext analytics properties

ruijialin-avalabs:
## Description

**Ticket: [CP-14082]**

- Add typed `PlaintextProperties<E>` conditional type to `captureWithEncryption` so plaintext fields are validated per event name at compile time
- Add `AnalyticsPlaintextEvents` type map in `analytics.ts` defining the plaintext shape for each event
- Guard `applyTempChainIdConversion` against double-converting values already in CAIP-2 format

Screenshot 2026-04-23 at 10 14 44 AM

Screenshot 2026-04-23 at 10 15 26 AM

## Testing
iOS: 8280
Android: 8281

- [ ] Verify `SendTransactionSucceeded` events include `chainId` as a top-level plaintext property (CAIP-2 format)
- [ ] Verify `SwapConfirmed` events include `sourceChainId` and `targetChainId` as top-level plaintext properties
- [ ] Verify encrypted `data` blob no longer contains chain ID fields for these events
- [ ] Verify existing analytics events without plaintext properties still work correctly
- [ ] TypeScript compiles cleanly (`yarn tsc`)

## Checklist

[CP-14082]: https://ava-labs.atlassian.net/browse/CP-14082

### DIFF
--- a/packages/core-mobile/app/services/analytics/AnalyticsService.ts
+++ b/packages/core-mobile/app/services/analytics/AnalyticsService.ts
@@ -7,7 +7,8 @@ import { AnalyticsServiceNoop } from 'services/analytics/AnalyticsServiceNoop'
 import {
   AnalyticsEventName,
   AnalyticsServiceInterface,
-  CaptureEventProperties
+  CaptureEventProperties,
+  PlaintextProperties
 } from './types'
 
 if (!Config.ANALYTICS_ENCRYPTION_KEY) {
@@ -47,7 +48,8 @@ class AnalyticsService implements AnalyticsServiceInterface {
 
   async captureWithEncryption<E extends AnalyticsEventName>(
     eventName: E,
-    properties: AnalyticsEvents[E]
+    properties: AnalyticsEvents[E],
+    plaintextProperties?: PlaintextProperties<E>
   ): Promise<void> {
     if (!this.isEnabled) {
       return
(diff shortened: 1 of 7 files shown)


packages/core-mobile/app/services/analytics/AnalyticsService.ts
```diff
@@ -64,7 +66,8 @@ class AnalyticsService implements AnalyticsServiceInterface {
       return PostHogService.capture(eventName, {
         data: encrypted,
         enc,
-        keyID: keyID
+        keyID: keyID,
+        ...plaintextProperties
```
atn4z7: can you confirm with Steve if this is the format he wants? or does he want `data : { ...plaintextProperties, encrypted: string }`?

packages/core-mobile/app/types/analytics.ts
```diff
@@ -206,7 +204,7 @@ export type AnalyticsEvents = {
       addressSVM: string
     }[]
   }
-  SendTransactionSucceeded: { txHash: string; chainId: number }
+  SendTransactionSucceeded: { txHash: string }
```
atn4z7: could we document these types another way so that we can see the whole shape?

packages/core-mobile/app/services/analytics/AnalyticsService.ts
ruijia1in: confirmed with Steve.

PR #3784, opened 2026-04-23
CP-13944: Separate plaintext chain IDs from encrypted analytics payloads

onghwan:
## Description

**Ticket: [CP-13944]**

- Introduce a dedicated `AnalyticsEncryptedEvents` type (enforced via `WithEncrypted<T>` helper) that forces every encrypted event to carry an `encrypted: { ... }` object alongside sibling plaintext properties (e.g. `caip2ChainId`, `caip2SourceChainId`, `caip2TargetChainId`).
- Narrow `AnalyticsService.captureWithEncryption` so the compiler only accepts events declared in `AnalyticsEncryptedEvents`; the method now destructures `{ encrypted, ...plaintext }`, encrypts only the `encrypted` object, and merges the plaintext siblings as top‑level PostHog properties. Spread order guarantees the encryption envelope (`data`/`enc`/`keyID`) cannot be overwritten by accidentally‑named plaintext keys.
- Migrate all 15 `captureWithEncryption` call sites (send, swap, bridge, earn/nest egg, account, WalletConnect, approval controller) to the new nested shape, and update associated Jest expectations (`ApprovalController.test.ts`, `walletConnect.test.ts`) to assert the nested `encrypted` payload and top‑level CAIP‑2 chain ID siblings.
- Motivation: previously `captureWithEncryption` was typed against `AnalyticsEvents`, which meant any event (including fully‑plaintext ones) could be passed through encryption, and there was no type‑level distinction between data that must be encrypted and data that may flow to the PostHog dashboard as plaintext for filtering. The new split prevents accidental leakage and makes the contract explicit at the type layer.
- Also pulls in `CP-14082`: `applyTempChainIdConversion` now guards with `isCaip2Format` so values that are already CAIP‑2 encoded are not re‑converted.
- No new dependencies. No runtime behavior changes to dashboards beyond the added plaintext CAIP‑2 chain IDs.

## Testing
iOS: 8284
Android: 8285

- [x] Verify `SendTransactionSucceeded` events include `chainId` as a top-level plaintext property (CAIP-2 format)
- [x] Verify `SwapConfirmed` events include `sourceChainId` and `targetChainId` as top-level plaintext properties
- [x] Verify encrypted `data` blob no longer contains chain ID fields for these events
- [ ] Verify existing analytics events without plaintext properties still work correctly
- [ ] TypeScript compiles cleanly (`yarn tsc`)

## Checklist
- [X] I have added testing steps
- [X] I have added/updated necessary unit tests

[CP-13944]: https://ava-labs.atlassian.net/browse/CP-13944?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core-mobile/app/new/features/bridge/hooks/useBridgeTransfer.ts
+++ b/packages/core-mobile/app/new/features/bridge/hooks/useBridgeTransfer.ts
@@ -73,10 +73,12 @@ export const useBridgeTransfer = ({
     })
 
     AnalyticsService.captureWithEncryption('BridgeTransactionStarted', {
-      chainId: sourceNetwork.chainId,
-      sourceTxHash: pendingTransfer.sourceTxHash,
-      fromAddress: pendingTransfer.fromAddress,
-      toAddress: pendingTransfer.toAddress
+      encrypted: {
+        chainId: sourceNetwork.chainId,
+        sourceTxHash: pendingTransfer.sourceTxHash,
+        fromAddress: pendingTransfer.fromAddress,
+        toAddress: pendingTransfer.toAddress
+      }
     })
 
     dispatch(setPendingTransfer(pendingTransfer))
--- a/packages/core-mobile/app/new/features/bridge/hooks/useBridgeTransfer.ts
+++ b/packages/core-mobile/app/new/features/bridge/hooks/useBridgeTransfer.ts
@@ -72,7 +72,7 @@ export const useBridgeTransfer = ({
       targetChainId: targetNetwork.chainId
     })
 
-    AnalyticsService.captureWithEncryption('BridgeTransactionStarted', {
+    AnalyticsService.capture('BridgeTransactionStarted', {
       encrypted: {
         chainId: sourceNetwork.chainId,
         sourceTxHash: pendingTransfer.sourceTxHash,
--- a/packages/core-mobile/app/new/features/collectibleSend/hooks/useSendTransactionCallbacks.ts
+++ b/packages/core-mobile/app/new/features/collectibleSend/hooks/useSendTransactionCallbacks.ts
@@ -34,7 +34,7 @@ export const useSendTransactionCallbacks = (): {
             txHash
           },
           {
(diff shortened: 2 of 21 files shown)


packages/core-mobile/app/new/features/bridge/hooks/useBridgeTransfer.ts
```diff
@@ -73,10 +73,12 @@ export const useBridgeTransfer = ({
     })
 
     AnalyticsService.captureWithEncryption('BridgeTransactionStarted', {
-      chainId: sourceNetwork.chainId,
-      sourceTxHash: pendingTransfer.sourceTxHash,
-      fromAddress: pendingTransfer.fromAddress,
-      toAddress: pendingTransfer.toAddress
```
atn4z7: seems like we should just remove this `captureWithEncryption` and have just a `capture` function. the params will be `encrypted` and `plaintext`. then the even type can be:

```
SendTransactionSucceeded: {
   encrypted: { txHash: string; chainId: number }
   plaintext: { caip2ChainId: string }
  }
```
what do you think?

onghwan: I like the idea of having a single capture function.

https://github.com/ava-labs/core-mobile/pull/3784/changes/b3ad6a9c6804f154d3c5d3e16ae5face214ea7c6

I still wanted to send the plain properties as-is, so I didn't label them with a "plaintext" property.

atn4z7: could we document all even types in one place? right now looks like you are doing them in 2 separate sections.

onghwan: fixed

atn4z7 (approved):

ruijialin-avalabs (approved):

PR #3794, opened 2026-04-23
CP-14060: Leverage Gauge Component

bogdandobritoiu:
## Description

**Ticket: [CP-14060](https://ava-labs.atlassian.net/browse/CP-14060)**

 ## Summary

Adds a new `LeverageGauge` component to `@avalabs/k2-alpine` - a Skia-rendered ruler/wheel for picking a numeric value with gestures, haptics, per-digit animations, and optional manual text input.

- **Ruler wheel** - pan gesture with iOS-style velocity, decay coast, and UIKit rubber-band formula past min/max. Over-damped spring returns to edge with inward-velocity zeroing to avoid "bounce deeper then return" glitches.
- **Animated number display** - each digit rendered as its own Skia text slot so only the characters that change animate. Decimal digits fade progressively toward neighbor values during motion; integers/`.` stay solid. × glyph slides with directional timing (instant on digit add, cubic-ease on digit drop).
- **Manual input** - optional `TextInput` overlay reveals on press; supports `integersOnly` (number-pad keyboard + digit-only filter) and arbitrary decimals. Cursor is pinned to end on focus (no select-all). Text updates on UI thread via `useAnimatedProps` so fast wheel swipes update visible value with zero JS dispatch lag.
- **Haptics** - light on minor ticks, medium on integers, heavy at edges. Suppressed during programmatic moves (preset presses) and passive sync. 
- **Dynamic width** -  the wheel measures its parent via `onLayout` and fills available horizontal space; no hardcoded widths.
- **Readiness-driven fade-in** - Canvas opacity is held at 0 until `useFont` resolves + two `requestAnimationFrame`s confirm React commit + first native paint, then fades in over 300ms. Replaces earlier timing-based approximations.
- **App-wide Skia warmup** - `SkiaPreload` is mounted inside `K2AlpineThemeProvider` so any consumer of k2-alpine gets typeface + glyph-atlas preloading for free at app start.

**Props**
- Range: `min`, `max`, `step`, `presets`, `decimals`, `integersOnly`
- Behavior: `enableManualInput`, `formatValue`, `subtitle`
- Feedback: `onHapticTick`, `onCommit`
- Physics: `velocityPower` (release flick multiplier), `coastDeceleration`

**Tests**
- 88 unit tests in `helpers.test.ts` covering `clamp`, `snapToStep`, `resolvePreset`,
`isMajorTick`, `validateRange`, `formatNumber`, `sanitizeTypedText`, `commitDraftText`, `shouldSyncExternalValue`.

## Screenshots/Videos

https://github.com/user-attachments/assets/b40a4aa0-a10a-4e90-a66f-aacae2600a5e

## Testing

iOS - 8305
Android - 8306

Integrated into `EventDetailsScreen` from `Trade` tab with both an integer gauge ("Add leverage") and a decimal gauge ("Add leverage (decimals)").

- [ ] Scrub the wheel - ticks bloom near the center, haptics fire at each step, edges feel heavy.
- [ ] Flick quickly - coast decelerates and snaps to the nearest step without looping between two values.
- [ ] Drag past min/max - rubber-band resistance engages and spring returns cleanly (no over-bounce past the edge).
- [ ] Tap Min / Max presets - wheel animates to the preset (≈900ms ease-out) instead of jumping.
- [ ] Tap the number (manual input mode) - keyboard opens, caret sits at the end of the existing value (no select-all).
- [ ] Type a value - wheel animates to match as you type; blur commits the value.
- [ ] Swipe the wheel while the input is focused - visible text updates live on UI thread; caret stays pinned to end.

## Checklist

[CP-14060]: https://ava-labs.atlassian.net/browse/CP-14060?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/k2-alpine/src/components/LeverageGauge/LeverageGauge.tsx
```diff
@@ -0,0 +1,270 @@
...
+      currentValue.value = withTiming(
+        snapped,
+        { duration: 900, easing: Easing.out(Easing.cubic) },
+        finished => {
+          isActive.value = false
+          isProgrammatic.value = false
+          if (finished) {
+            scheduleOnRN(stableOnChange, snapped)
+            scheduleOnRN(stableOnCommit, snapped)
+          }
```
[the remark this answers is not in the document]
bogdandobritoiu: Good catch, fixed. The callback now early-returns on `!finished` so an interrupting gesture keeps ownership of `isActive`/`isProgrammatic`.

packages/k2-alpine/src/components/LeverageGauge/LeverageGauge.tsx
```diff
@@ -0,0 +1,270 @@
...
+        // animations driven by actual paint-readiness, not timing guesses.
+        entering={FadeIn.duration(200)}
+        style={{ alignSelf: 'stretch', gap: 4 }}
+        testID={testID}
+        accessible
+        accessibilityRole="adjustable"
+        accessibilityValue={{
+          min: Math.round(vMin),
+          max: Math.round(vMax),
+          now: Math.round(value),
```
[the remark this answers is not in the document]
bogdandobritoiu: Tried this and it crashes — the native a11y bridge converts `min`/`max`/`now` to `int64` and throws `Loss of precision during arithmetic conversion: (long long) 4.2`. The precise value is already carried in `text: formatValue(value)`, which VoiceOver/TalkBack prefer anyway, so keeping the rounded ints for the numeric fields.

packages/k2-alpine/src/components/LeverageGauge/LeverageGauge.tsx
```diff
@@ -0,0 +1,270 @@
...
+    [min, max, effectiveStep]
+  )
+
+  // Decimal places to show. integersOnly forces 0; explicit prop wins next;
+  // otherwise inferred from step.
+  const vDecimals = useMemo(() => {
+    if (integersOnly) return 0
+    if (typeof decimals === 'number') return Math.max(0, Math.floor(decimals))
+    return vStep < 1 ? Math.ceil(-Math.log10(vStep)) : 0
+  }, [integersOnly, decimals, vStep])
```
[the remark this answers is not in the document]
bogdandobritoiu: Fixed. `vDecimals` now uses a new `getStepDecimals` helper shared with `snapToStep`, so step=0.25 correctly infers 2 decimals.

packages/k2-alpine/src/components/LeverageGauge/LeverageGauge.tsx
```diff
@@ -0,0 +1,270 @@
...
+        />
+      </View>
+    )
+  }
+
+  return (
+    <GestureHandlerRootView style={{ flexShrink: 1 }}>
+      <Animated.View
+        // Short fade on the chrome (card, Min/Max buttons, subtitle). The
+        // Skia canvases inside (wheel + number) have their own opacity
```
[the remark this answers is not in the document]
bogdandobritoiu: Agreed on the inconsistency — leaving as follow-up so this PR stays scoped to the component itself. Will open a separate cleanup to align the k2-alpine gesture pattern (drop the embedded root here and document the requirement).

packages/k2-alpine/src/components/LeverageGauge/LeverageWheel.tsx
```diff
@@ -0,0 +1,518 @@
...
+        currentValue.value = withTiming(
+          snapped,
+          { duration: 900, easing: Easing.out(Easing.cubic) },
+          finished => {
+            // If cancelled by a new gesture (finished=false, or a drag is
+            // currently in progress), do NOT touch isActive / commit —
+            // otherwise we hand control back to the sync effect mid-gesture.
+            if (!finished || isDragging.value) return
+            isActive.value = false
+            scheduleOnRN(onChange, snapped)
```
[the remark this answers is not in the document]
bogdandobritoiu: Fixed. Removed the pre-emit; `settleTo` now calls `onChange` once on completion, with `onCommit` right after.

packages/k2-alpine/src/components/LeverageGauge/LeverageWheel.tsx
```diff
@@ -0,0 +1,518 @@
...
+    return ticks
+      .filter(t => t.isMajor)
+      .map(t => {
+        const integer = Math.round(t.value)
+        const text = String(integer)
+        const textWidth = labelFont.measureText(text).width
+        return { value: integer, text, x: t.x - textWidth / 2 }
+      })
+  }, [ticks, labelFont])
+
```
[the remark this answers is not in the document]
bogdandobritoiu: Fair point. In practice we only expect small ranges (1–40 for leverage), so keeping the full tick set for now. If we add consumers with larger ranges, adaptive step / virtualization is the right move — tracking as follow-up.

packages/k2-alpine/src/components/LeverageGauge/helpers.ts
```diff
@@ -0,0 +1,137 @@
...
+  if (v > max) return max
+  return v
+}
+
+export const snapToStep = (v: number, min: number, step: number): number => {
+  'worklet'
+  const offset = v - min
+  const snapped = Math.round(offset / step) * step
+  // Avoid floating-point drift for common step sizes
+  const decimals = step < 1 ? Math.ceil(-Math.log10(step)) : 0
```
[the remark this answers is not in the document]
bogdandobritoiu: Fixed. Added `getStepDecimals` that scales until integer instead of using `-log10(step)`, so non-power-of-10 steps (0.25, 0.125) round correctly. Added regression tests for 0.25 and 0.125.

packages/k2-alpine/src/components/LeverageGauge/types.ts
```diff
@@ -0,0 +1,52 @@
...
+   */
+  integersOnly?: boolean
+
+  // Optional — behavior
+  enableManualInput?: boolean
+  formatValue?: (v: number) => string
+  subtitle?: string
+
+  // Optional — feedback
+  onHapticTick?: boolean
```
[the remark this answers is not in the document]
bogdandobritoiu: Agreed the `on*` prefix is misleading for a boolean. Holding off on the rename until we decide whether to make it a callback (per-tick haptic customization), since either outcome changes the prop. Will revisit in a follow-up.

packages/k2-alpine/src/hooks/usePreloadSkiaFonts.tsx
```diff
@@ -0,0 +1,60 @@
...
+      {bigFont && (
+        <SkText text={WARMUP_CHARS} x={0} y={0} font={bigFont} color="black" />
+      )}
+      {smallFont && (
+        <SkText
+          text={WARMUP_CHARS}
+          x={0}
+          y={0}
+          font={smallFont}
+          color="black"
```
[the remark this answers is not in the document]
bogdandobritoiu: Good catch. Fixed by sizing the warmup Canvas to 800×100 and positioning it off-screen (`top: -200`) so Skia actually rasterizes all glyphs into the atlas instead of clipping most of them away.

packages/k2-alpine/src/components/LeverageGauge/LeverageGauge.tsx
```diff
@@ -0,0 +1,283 @@
...
+      isProgrammatic.value = true
+      currentValue.value = withTiming(
+        snapped,
+        { duration: 900, easing: Easing.out(Easing.cubic) },
+        finished => {
+          // If cancelled (e.g. by a new gesture), leave the flags alone —
+          // the interrupting gesture is now responsible for managing them.
+          // Clearing them here would re-enable prop sync / haptics
+          // mid-gesture.
+          if (!finished) return
```
atn4z7, line 131: shouldn't we reset isProgrammatic in this case?

packages/k2-alpine/src/components/LeverageGauge/LeverageWheel.tsx
```diff
@@ -0,0 +1,517 @@
...
+  // True only during finger-down drag. During decay/settle it's false, which
+  // lets the step-crossing reaction skip onChange and avoid flooding the JS
+  // queue with stale values that could fire after the settle completes.
+  const isDragging = useSharedValue(false)
+
+  const panGesture = Gesture.Pan()
+    .onStart(() => {
+      isActive.value = true
+      isDragging.value = true
+      gestureStartValue.value = currentValue.value
```
atn4z7, line 167: what about isProgrammatic? should we set it to false here?

packages/k2-alpine/src/hooks/usePreloadSkiaFonts.tsx
```diff
@@ -0,0 +1,70 @@
...
+ *
+ * We render invisible SkText covering all digit glyphs at both sizes so
+ * the rasterized atlases are populated before any real component mounts.
+ * Mount once near the root of the app.
+ */
+const BIG_FONT_SIZE = 60
+const SMALL_FONT_SIZE = 11
+// Characters any LeverageGauge-style Skia component might render — digits,
+// decimal point, and the "×" multiplier glyph.
+const WARMUP_CHARS = '0123456789.x'
```
atn4z7: looks like this x (lowercase Latin letter X) is different from the x (multiplication sign) rendered in the actual component so skia warmup probably misses it.

packages/k2-alpine/src/components/LeverageGauge/LeverageGauge.tsx
```diff
@@ -0,0 +1,283 @@
...
+        continue
+      }
+      if (p >= vMin && p <= vMax) out.push(p)
+      else dropped++
+    }
+    if (dropped > 0) {
+      // eslint-disable-next-line no-console
+      console.warn(
+        `[LeverageGauge] Dropped ${dropped} preset(s) outside [${vMin}, ${vMax}].`
+      )
```
atn4z7, line 99: let's use logger.warn

packages/k2-alpine/src/components/LeverageGauge/LeverageWheel.tsx
```diff
@@ -0,0 +1,517 @@
...
+  /** True during preset/programmatic animations — suppresses haptics. */
+  isProgrammatic: SharedValue<boolean>
+  min: number
+  max: number
+  step: number
+  /** When true, only major (integer) ticks animate; sub-steps stay static. */
+  integersOnly: boolean
+  onChange: (value: number) => void
+  onCommit: (value: number) => void
+  onHapticTick: boolean
```
atn4z7: probably better to name enableHaptics or hapticsEnabled since this is not a callback

packages/k2-alpine/src/components/LeverageGauge/LeverageDisplay.tsx
```diff
@@ -0,0 +1,562 @@
...
+    opacity: canvasOpacity.value
+  }))
+  useEffect(() => {
+    if (!font) return
+    const cancelIds: { raf2?: number } = {}
+    const raf1 = requestAnimationFrame(() => {
+      cancelIds.raf2 = requestAnimationFrame(() => {
+        canvasOpacity.value = withTiming(1, { duration: 300 })
+      })
+    })
```
atn4z7, line 371: just curious why we need 2 requestAnimationFrame here. in startEdit, we only have 1.

packages/k2-alpine/src/theme/ThemeProvider.tsx
```diff
@@ -1,13 +1,15 @@
...
 import { ColorSchemeName } from 'react-native'
+import { SkiaPreload } from '../hooks/usePreloadSkiaFonts'
 import { darkTheme, lightTheme } from './theme'
 
 export const K2AlpineThemeProvider: FC<
   PropsWithChildren & { colorScheme: ColorSchemeName }
 > = ({ colorScheme, children }) => {
   return (
     <DripsyProvider theme={colorScheme === 'dark' ? darkTheme : lightTheme}>
+      <SkiaPreload />
```
atn4z7: I have a concern about putting this here. This means the app will always pay the warmup cost even if users don't use the LeverageGauge component. Should we load `SkiaPreload` in the navigation/layout stack for this flow instead?

packages/k2-alpine/src/components/LeverageGauge/LeverageGauge.tsx
bogdandobritoiu: k2-alpine doesn't depend on core-mobile, and the `Logger` util lives in `core-mobile/app/utils/Logger.ts`. There are already two other `console.warn` calls in this folder (`helpers.ts`) — keeping them consistent for now. Happy to revisit if we add a shared logger to the design system later.

packages/k2-alpine/src/theme/ThemeProvider.tsx
bogdandobritoiu: Good call — moved `SkiaPreload` out of `K2AlpineThemeProvider` and into the trade tab's `_layout.tsx` so apps that never enter trade don't pay the warmup cost. The trade tab is mounted before the user can drill into a market detail, so the warmup window is preserved.

packages/k2-alpine/src/components/LeverageGauge/LeverageDisplay.tsx
bogdandobritoiu: Different purposes — the two RAFs here gate the Skia canvas opacity fade-in:
- 1st RAF: wait for React to commit
- 2nd RAF: wait for the native Canvas to tick a frame

With only one RAF, the fade-in starts before Skia has actually painted, and you see a flash of the canvas appearing mid-fade. Both RAFs are needed for the readiness gate.

`startEdit` only needs one RAF because it's just waiting for React to mount the `TextInput` so we can call `.focus()` on it — no Skia paint involved.

packages/k2-alpine/src/theme/ThemeProvider.tsx
atn4z7: I was thinking about putting it in the layout component for the trade flow (not the actual screen) so it should still warm up everything correctly. gating it behind a prop is cool as long as it can be done dynamically. I'm down to keep as is if you think it doesn't affect performance yet.

packages/k2-alpine/src/components/LeverageGauge/LeverageGauge.tsx
bogdandobritoiu: Fixed.

packages/k2-alpine/src/components/LeverageGauge/LeverageWheel.tsx
bogdandobritoiu: Done. Set `isProgrammatic.value = false` here.

packages/k2-alpine/src/hooks/usePreloadSkiaFonts.tsx
bogdandobritoiu: Good catch

packages/k2-alpine/src/components/LeverageGauge/LeverageWheel.tsx
bogdandobritoiu: Renamed to `hapticsEnabled`.

packages/k2-alpine/src/theme/ThemeProvider.tsx
bogdandobritoiu: I loaded it in the trade layout

atn4z7 (approved):

onghwan (approved):

PR #3795, opened 2026-04-24
CP-14129: stop showing notifications for deleted wallets

B0Y3R-AVA:
## Summary
- Re-subscribes balance-change notifications on `removeAccount` so the backend webhook drops addresses for deleted wallets. The `/v1/push/balance-changes/subscribe` endpoint is a full override (confirmed by Ferenc and the service docs), so re-posting the current address list is the correct "unsubscribe" — the dedicated unsubscribe endpoint is device-wide and wrong for this case.
- Re-runs the same subscribe on `onAppUnlocked` so users who already deleted a wallet *before* installing this version get auto-reconciled the next time they open the app, rather than having to re-add and remove the wallet.
- Shows the wallet/account a balance-change item belongs to in the Notification Center, so users with multiple imported wallets can tell them apart (Gergely's ask).

https://github.com/user-attachments/assets/3622abd0-aa44-41e2-ac68-9aea92862465

## Why
Akash reported this morning that the Notification Center was showing transactions he never made ("USDC approved", etc.). Amanda screen-shared with him and repro'd the bug: import a wallet, delete it, have someone else transact on that address → the tx still appears in Akash's NC.

Root cause on the mobile side: `removeAccount` / `removeWallet` Redux dispatches never re-triggered `subscribeBalanceChangeNotifications`, so the backend kept the deleted address subscribed. Parallel fix landing on extension (Bohdan, linked thread below) uses the same approach since the extension already resubscribes on `ACCOUNTS_UPDATED`.

Expected behavior, confirmed by Gergely + Ferenc ([thread](https://avalancheavax.slack.com/archives/C0440E1H8DS/p1777040332726119)):

iOS: 8351
Android: 8352

1. Add wallet 1 & 2
2. Transact on wallet 2 → TX1 delivered
3. NC shows TX1
4. Remove wallet 2
5. Someone transacts on wallet 2 → TX2
6. **TX1 should remain in the NC until dismissed** (already delivered history)
7. **TX2 should not arrive**

Step 7 is handled by the listener change (this PR). Step 6 means we intentionally do **not** filter already-delivered notifications client-side — Ferenc confirmed the backend keeps history and Gergely confirmed that's the desired UX.

Jira:
- https://ava-labs.atlassian.net/browse/CP-14129 — Amanda's mobile NC ticket
- https://ava-labs.atlassian.net/browse/CP-14130 — cross-platform client-side bug (Bohdan, extension fix in parallel)

## Test plan
- [ ] Import a second wallet into a prod build
- [ ] Have someone transact on that wallet → confirm notification appears in NC attributed to the correct wallet/account
- [ ] Delete the imported wallet
- [ ] Have them transact again → confirm **no new notification** appears for the deleted address
- [ ] Confirm the **previously-delivered** notification for the deleted wallet is **still visible** in the NC until dismissed
- [ ] **Legacy-user case:** on a build *before* this PR, import a wallet, delete it, then upgrade to this build. Open the app and unlock with PIN. Have someone transact on the deleted address → confirm no new notification arrives (the unlock-time re-sync should drop it from the backend webhook).
- [ ] Single-wallet setup: items show just the account name (e.g. "Account 2 · to 0x...")
- [ ] Two-wallet setup: items show wallet + account (e.g. "Wallet B · Account 2 · to 0x...")
- [ ] Confirm price-alert and news notifications still render unchanged

### DIFF
--- a/packages/core-mobile/app/new/features/notifications/components/BalanceChangeItem.tsx
+++ b/packages/core-mobile/app/new/features/notifications/components/BalanceChangeItem.tsx
@@ -1,13 +1,9 @@
 import React, { FC } from 'react'
-import { useSelector } from 'react-redux'
-import { selectAccounts } from 'store/account'
-import { selectWallets } from 'store/wallet/slice'
 import {
   AppNotification,
   isBalanceChangeNotification,
   BalanceChangeEventSchema
 } from '../types'
-import { getAccountLabel } from '../utils'
 import NotificationListItem from './NotificationListItem'
 import NotificationIcon from './NotificationIcon'
 
@@ -15,6 +11,12 @@ type BalanceChangeItemProps = {
   notification: AppNotification
   showSeparator: boolean
   accessoryType: 'chevron' | 'none'
+  /**
+   * Optional "{wallet} · {account}" label resolved by the parent screen.
+   * Prepended to the subtitle so multi-wallet users can tell which wallet a
+   * balance-change notification belongs to.
+   */
+  accountLabel?: string | null
   testID?: string
 }
 
@@ -74,11 +76,9 @@ const BalanceChangeItem: FC<BalanceChangeItemProps> = ({
   notification,
   showSeparator,
   accessoryType,
+  accountLabel,
   testID
 }) => {
-  const accounts = useSelector(selectAccounts)
-  const wallets = useSelector(selectWallets)
-  const accountLabel = getAccountLabel(notification, accounts, wallets)
   const baseSubtitle = getSubtitle(notification)
   const subtitle = accountLabel
     ? `${accountLabel} · ${baseSubtitle}`
--- a/packages/core-mobile/app/new/features/notifications/components/BalanceChangeItem.tsx
+++ b/packages/core-mobile/app/new/features/notifications/components/BalanceChangeItem.tsx
@@ -1,9 +1,13 @@
 import React, { FC } from 'react'
+import { useSelector } from 'react-redux'
+import { selectAccounts } from 'store/account'
+import { selectWallets } from 'store/wallet/slice'
 import {
   AppNotification,
   isBalanceChangeNotification,
   BalanceChangeEventSchema
 } from '../types'
+import { getAccountLabel } from '../utils'
 import NotificationListItem from './NotificationListItem'
 import NotificationIcon from './NotificationIcon'
 
@@ -72,10 +76,18 @@ const BalanceChangeItem: FC<BalanceChangeItemProps> = ({
   accessoryType,
   testID
 }) => {
+  const accounts = useSelector(selectAccounts)
+  const wallets = useSelector(selectWallets)
+  const accountLabel = getAccountLabel(notification, accounts, wallets)
+  const baseSubtitle = getSubtitle(notification)
+  const subtitle = accountLabel
+    ? `${accountLabel} · ${baseSubtitle}`
+    : baseSubtitle
+
   return (
     <NotificationListItem
       title={getTitle(notification)}
-      subtitle={getSubtitle(notification)}
+      subtitle={subtitle}
       icon={<NotificationIcon notification={notification} />}
       timestamp={notification.timestamp}
       showSeparator={showSeparator}
--- a/packages/core-mobile/app/new/features/notifications/utils.ts
+++ b/packages/core-mobile/app/new/features/notifications/utils.ts
@@ -3,7 +3,6 @@ import { AccountCollection } from 'store/account/types'
 import { Wallet } from 'store/wallet/types'
 import {
   AppNotification,
-  isBalanceChangeNotification,
   NotificationCategory,
   NotificationTab,
   NotificationType,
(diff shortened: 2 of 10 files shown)


packages/core-mobile/app/new/features/notifications/utils.ts
```diff
@@ -99,6 +102,37 @@ export function mapTypeToCategory(
...
+  accounts: AccountCollection,
+  wallets: { [id: string]: Wallet }
+): string | null {
+  if (!isBalanceChangeNotification(notification)) return null
+  const addr = notification.data?.accountAddress?.toLowerCase()
+  if (!addr) return null
+
+  const account = Object.values(accounts).find(
+    a => a.addressC.toLowerCase() === addr
+  )
```
atn4z7: should we prioritize active account over the others?

packages/core-mobile/app/store/notifications/listeners/listeners.ts
```diff
@@ -89,9 +94,11 @@ export const addNotificationsListeners = (
   startListening({
     matcher: isAnyOf(
       onRehydrationComplete,
+      onAppUnlocked,
```
atn4z7: I don't think we need both onRehydrationComplete and onAppUnlocked. if we are doing onAppUnlocked now, let's remove onRehydrationComplete.

packages/core-mobile/app/new/features/notifications/utils.ts
B0Y3R-AVA: Each EVM `addressC` is unique per BIP44 derivation index, so `find` only ever returns the single account that owns the notification address — there is no ambiguity to break. Active account prioritization wouldn`t change which row matches, just shave a microscopic scan cost in a list that has at most a handful of accounts. Happy to memoize the lookup if we want to be extra defensive on perf.

packages/core-mobile/app/store/notifications/listeners/listeners.ts
B0Y3R-AVA: Done in e4b807b — dropped `onRehydrationComplete` from this matcher. While I was here I also added `cancelActiveListeners()` + a small `delay()` to coalesce the burst of `removeAccount` dispatches that `removeWallet` emits, so we only fire one subscribe with the final address set (addresses Copilot`s race-condition note).

packages/core-mobile/app/new/features/notifications/utils.ts
atn4z7: but I believe we could have the same address in different wallets (mnemonic, ledger, private key,...)

packages/core-mobile/app/store/notifications/listeners/listeners.ts
atn4z7: should we only do this for `removeAccount` or it's fine to do it for all actions?

packages/core-mobile/app/new/features/notifications/components/BalanceChangeItem.tsx
```diff
@@ -72,10 +76,18 @@ const BalanceChangeItem: FC<BalanceChangeItemProps> = ({
   accessoryType,
   testID
 }) => {
+  const accounts = useSelector(selectAccounts)
+  const wallets = useSelector(selectWallets)
+  const accountLabel = getAccountLabel(notification, accounts, wallets)
+  const baseSubtitle = getSubtitle(notification)
+  const subtitle = accountLabel
+    ? `${accountLabel} · ${baseSubtitle}`
+    : baseSubtitle
```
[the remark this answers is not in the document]
B0Y3R-AVA: Done in 2700687 — `BalanceChangeItem` is now dumb (no Redux subs), and `NotificationsScreen` builds an `address → label` Map once via `useMemo` and passes the resolved label down as a prop. Replaced `getAccountLabel` with `buildAccountLabelMap` in `utils.ts` and updated the unit tests accordingly.

packages/core-mobile/app/new/features/notifications/utils.ts
B0Y3R-AVA: Good catch — you`re right, a private-key wallet (or another mnemonic) can legitimately produce an account whose `addressC` already exists in another wallet. Fixed in 3f97129: `buildAccountLabelMap` now takes `activeAccountId` and processes the active account first, so when an address is owned by accounts in multiple wallets the active wallet`s label wins. Added a unit test covering this case.

packages/core-mobile/app/store/notifications/listeners/listeners.ts
B0Y3R-AVA: Fine for all actions IMO — the coalescing is idempotent and just bounds how many subscribe POSTs we make per burst. The flow on every fired action: cancel any in-flight effect, wait 100ms (and now also abort the superseded fetch via `listenerApi.signal`), then post the *current* address set. Whether the burst was triggered by `removeAccount`, `setAccounts`, `onAppUnlocked`, etc. doesn`t change correctness — and on quiet activity it just collapses to a single subscribe per ~100ms window.

packages/core-mobile/app/store/notifications/listeners/listeners.ts
```diff
@@ -88,15 +93,21 @@ export const addNotificationsListeners = (
...
       onFcmTokenChange,
       turnOnAllNotifications,
       onNotificationsTurnedOnForBalanceChange
     ),
     effect: async (_, listenerApi) => {
+      // removeWallet dispatches removeAccount once per account synchronously,
+      // so coalesce bursts to avoid an out-of-order subscribe re-adding an
+      // address we just removed (the backend reconciles to the posted set).
+      listenerApi.cancelActiveListeners()
+      await listenerApi.delay(100)
```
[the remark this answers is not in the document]
B0Y3R-AVA: Done in 3f97129 — wrapped `delay()` in try/catch with an early return on cancel, matching the pattern in `swap/store/listeners.ts`.

packages/core-mobile/app/store/notifications/listeners/listeners.ts
```diff
@@ -88,15 +93,21 @@ export const addNotificationsListeners = (
...
       await subscribeBalanceChangeNotifications(listenerApi).catch(reason => {
         Logger.error(
```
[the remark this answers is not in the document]
B0Y3R-AVA: Threaded `listenerApi.signal` down through `subscribeBalanceChangeNotifications` → `subscribeForBalanceChange` → `appCheckPostJson` in 3f97129, so a superseded subscribe is now aborted at the network layer rather than just at the listener boundary. `appCheckPostJson` got an optional `{ signal }` options bag — backward compatible with all existing callers.

packages/core-mobile/app/new/features/notifications/utils.ts
```diff
@@ -99,6 +102,37 @@ export function mapTypeToCategory(
...
+  if (!account) return null
+
+  const walletCount = Object.keys(wallets).length
+  if (walletCount > 1) {
+    const wallet = wallets[account.walletId]
+    if (wallet) return `${wallet.name} · ${account.name}`
+  }
```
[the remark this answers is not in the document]
B0Y3R-AVA: Already addressed earlier — `getAccountLabel` was replaced with `buildAccountLabelMap` (returns `Map<lowerAddr, label>`), and `NotificationsScreen` builds the map once via `useMemo` and passes the resolved label down as a prop. List rows no longer subscribe to Redux or scan the accounts collection per render.

packages/core-mobile/app/store/notifications/listeners/listeners.ts
atn4z7: sounds good. could you adjust the comment to reflect that?

B0Y3R-AVA: Reworked the comment in 15602b8 to make the general intent explicit — we coalesce bursts for *any* matched action, with `removeWallet` cascading `removeAccount`s as the most concrete motivating case but not the only one. Also called out that `cancelActiveListeners()` + the threaded `signal` cover both the listener boundary and the in-flight fetch.

atn4z7 (dismissed):

onghwan (approved):
dev-tested

atn4z7 (approved):

PR #3796, opened 2026-04-24
QA - Unfreeze android internal e2e build for automation

eunjisong:
## Description
- Android e2e testing was blocked because the app was stuck on the splash screen and not moving forward. It's unblocked now with this PR and we can test android internal e2e on bitrise. 
- iOS external e2e testing on AWS device farm setup is added. It's not fully functioning yet since I don't have access to the device farm yet but it's a starting point 
- Remove some smoke tag from some tests so that the smoke run can be fast and compact 
- I am expecting some failures on android test from now on since it's just functioning. I'll add any maintenance code in the future PR.


packages/react-native-nitro-avalabs-crypto/android/build.gradle
```diff
@@ -486,6 +486,7 @@ android {
   namespace "com.margelo.nitro.nitroavalabscrypto"
 
   compileSdkVersion getExtOrIntegerDefault("compileSdkVersion")
+  ndkVersion "27.1.12297006"
```
atn4z7: could you explain this change? is it needed to fix the Android being stuck issue?

scripts/bitrise/yarnSetup.sh
```diff
@@ -10,17 +10,19 @@ corepack enable
...
     # on ubuntu, yarn setup command will fail
     # as patch-package doesn't have write access to the node_modules folder
     # thus, we need to set write permission manually and rerun yarn
-    # Bitrise issue link https://support.bitrise.io/hc/en-us/requests/39436?page=1
+    # Bitrise issue link https://support.bitrise.io/hc/en-us/requests/39436?pc=1
     sudo chown root .
 
+    export APP_ENV=ci
     yarn install --immutable && yarn setup
 fi
```
atn4z7: could you explain this change as well? also do we need it twice?

eunjisong: I've updated the code.

packages/react-native-nitro-avalabs-crypto/android/build.gradle
eunjisong: I've removed it, working without it 👍 thank you!

atn4z7 (approved):

ruijialin-avalabs (approved):

PR #3807, opened 2026-04-29
CP-14164: force native crypto resolution + wallet/seed caching for address derivation

ruijialin-avalabs:
## Summary

Bitrise: https://app.bitrise.io/app/7d7ca5af7066e290/pipelines/180f678a-8176-481f-81ca-e355872d0348
iOS: 84156
Android: 8417

- Force all `@noble/hashes` and `@noble/curves` imports to resolve to patched native crypto copies, eliminating all unpatched duplicates from the bundle
- Cache wallet instances, BIP39 seeds, and BIP32/SLIP-10 master nodes to eliminate redundant keychain reads and cryptographic derivations
- Evict cached secrets from JS heap on app lock and logout
- Fix concurrency race condition in wallet creation and correctness bug in native `pointAddScalar` infinity check
- **76x speedup** for address derivation: 20 accounts from ~111s → ~1.5s

**Ticket: [CP-14062](https://ava-labs.atlassian.net/browse/CP-14062)**

## Problem

Multiple unpatched copies of `@noble/hashes` (9 copies, only 1 patched) and `@noble/curves` (5 copies, only 1 patched) were installed due to transitive dependencies from `ethereum-cryptography`, `ox`, `viem`, and others. The patches that redirect crypto operations to native implementations (`react-native-quick-crypto` for hashes, `react-native-nitro-avalabs-crypto` for curves) only applied to the top-level copies.

Additionally, `WalletFactory.createWallet()` was called for every `getPublicKeyFor` cache miss, triggering expensive `BiometricsSDK.loadWalletSecret()` (Keychain read + AES decrypt) and `mnemonicToSeed()` (2048 PBKDF2-HMAC-SHA512 iterations) on every call. BIP32 and SLIP-10 master nodes were also re-derived from the seed on every key derivation.

### 1. Force native crypto resolution (Metro + Yarn)

**`metro.config.js`** — Added `resolveRequest` overrides that force ALL `@noble/hashes` and `@noble/curves` imports to resolve to the single patched top-level copy. Extracted shared resolution logic into `resolveNoblePackage()` with try/catch fallback so Metro doesn't crash if a subpath is missing.

**`package.json`** (root) — Single global `resolutions` entry for each: `@noble/curves` → patched 1.9.7, `@noble/hashes` → 1.8.0. Replaces what had grown into 14 version-specific `@noble/curves` selectors and 11 per-parent `@noble/hashes` selectors. Direct deps also added in `packages/core-mobile/package.json` for explicitness.

**`@noble-curves` patch** — Merged the previously-stacked weierstrass + edwards patches into a single file, dropping the brittle inner-patch-hash selector that went stale on every patch edit. Removed unused `react-native`/`browser` polyfill blocks (`zlib`, `http`, `fs`, `stream`, etc.) and an accidental `.DS_Store` from the patch — `@noble/curves` is pure ECC and imports none of those.

**Result:** Lockfile collapses to **1 copy each** of `@noble/curves` (patched 1.9.7) and `@noble/hashes` (1.8.0), down from 5+ and 9+ unpatched duplicates. Verified via bundle sourcemap — zero nested/unpatched copies in the iOS bundle.

### 2. Wallet instance + seed + master node caching

**`WalletFactory.ts`** — `getOrCreateWallet()` returns cached instances for read-only operations with in-flight Promise deduplication to prevent concurrent callers from triggering duplicate keychain reads.

**`MnemonicWallet.ts`** — Caches three layers of derived state as private fields:
- `#seedCache` — `mnemonicToSeed()` result (2048 PBKDF2 rounds), computed once per instance
- `#bip32Node` — BIP32 master node from seed, reused across SECP256K1 key derivations
- `#slip10Node` — SLIP-10 master node from seed, reused across ED25519 key derivations and Solana signing

**`WalletService.tsx`** — Read-only methods (`getPublicKey`, `getPublicKeyFor`, `getRawXpubXP`) use `getOrCreateWallet()`. Signing operations (`sign`, `signMessage`) use `createWallet()` for fresh instances.

**`WalletDerivedDataCache.ts`** — Refactored to nested Maps (`walletId → subkey → value`) for O(1) per-wallet clearing. Added in-flight Promise tracking for concurrent wallet creation deduplication.

### 3. Security: secret eviction + cache invalidation

**`thunks.ts`** — `clearWallet(walletId)` called on individual wallet removal.

### 4. Secp256k1 patch fixes

**`@bitcoinerlab-secp256k1` patch** —
- Fixed `Uint8Array → ArrayBuffer` conversion for Nitro bridge via `toNitroBuf()` helper
- Fixed `_pointAddScalar` infinity check: normalize native `ArrayBuffer` result to `Uint8Array` before `.length` check (ArrayBuffer has `.byteLength`, not `.length`)
- Fixed hunk header count (33 → 36)
- Removed accidental `.DS_Store` diff and unrelated `package.json` react-native/browser field mappings

## Benchmark Results

| Metric | Before | After | Speedup |
|---|---|---|---|
| Total (20 indices) | 110,945ms | 1,456ms | **76x** |
| Avg per index | 5,547ms | 73ms | **76x** |
| Median per index | 6,519ms | 79ms | **83x** |

## Testing

- use `rebel.... ` wallet or any wallet with 20+ account, the background account discovery should be a lot faster compare to latest main
- in general manual add account, create new mnemonic wallet should be quicker, and signing transaction should be quicker, since we are caching the wallet, caching the seed.  vs before, we create wallet instance every time, and load secret from storage

[CP-14062]: https://ava-labs.atlassian.net/browse/CP-14062?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


onghwan (dismissed):

packages/core-mobile/app/services/wallet/WalletFactory.ts
```diff
@@ -12,11 +12,37 @@ import { WalletDerivedDataCache } from './WalletDerivedDataCache'
 
 class WalletFactory {
   private derivedDataCache = new WalletDerivedDataCache()
+  private walletInstanceCache = new Map<string, Wallet>()
```
atn4z7: should we just add this to WalletDerivedDataCache instead?

packages/core-mobile/app/store/wallet/thunks.ts
```diff
@@ -185,6 +186,10 @@ export const removeWallet = createAsyncThunk<void, string, ThunkApi>(
 
     thunkApi.dispatch(_removeWallet(walletId))
 
+    // Clear cached wallet instance and derived data for the removed wallet
+    WalletFactory.clearWalletInstance(walletId)
+    WalletFactory.cache.clearWallet(walletId)
```
atn4z7: if not, we might need to rename the cache property here since it has become a bit confusing now that we have different caches.

packages/core-mobile/app/services/wallet/WalletFactory.ts
ruijia1in: moved.

packages/core-mobile/app/services/wallet/WalletFactory.ts
```diff
@@ -17,6 +17,23 @@ class WalletFactory {
     return this.derivedDataCache
   }
 
+  async getOrCreateWallet({
```
atn4z7, line 19: nit: I feel like we don't need to rename this to getOrCreateWallet. it's kinda an implementation detail. we are doing the same already with stuff like pubkey.

atn4z7 (dismissed):

packages/core-mobile/app/services/wallet/WalletFactory.ts
ruijia1in: will address this as well.  I need to fix the patch files anyway.

ruijia1in: actually I think it is better to name it getOrCreateWallet, since we are now caching the wallet instance.

atn4z7: I see your point, but we're already doing the same with `getPublicKey` for example. and just like with React Query hooks/functions, we don't need to reflect caching in the name since that's more of an implementation detail than part of the API.

atn4z7 (dismissed):

atn4z7 (dismissed):
great work!!

atn4z7 (approved):

atn4z7 (dismissed):

.yarn/patches/@bitcoinerlab-secp256k1-npm-1.2.0-1098d4b329.patch
```diff
@@ -158,48 +150,8 @@ index 4a914633a13067af05838ab582a9dd41b53306ba..68df133248e9a1d067735ad1915f64af
...
-+    "_stream_readable": "readable-stream/readable",
-+    "_stream_writable": "readable-stream/writable",
-+    "_stream_duplex": "readable-stream/duplex",
-+    "_stream_passthrough": "readable-stream/passthrough",
-+    "stream": "stream-browserify",
-+    "vm": "vm-browserify"
-+  },
-+  "browser": {
-+    "zlib": "browserify-zlib",
-+    "http": "@tradle/react-native-http",
```
atn4z7, line 192: this stuff should be removed from the patch since it is auto generated after yarn install

packages/core-mobile/app/store/app/listeners.ts
```diff
@@ -141,6 +139,8 @@ const lockApp = async (
   if (isTimeManipulated || secondsPassed >= TIME_TO_LOCK_IN_SECONDS) {
     dispatch(setIsLocked(true))
     dispatch(onAppLocked())
+    // Clear cached wallet instances to evict secrets (mnemonics, seeds, private keys) from heap
+    walletDerivedDataCache.clearAll()
```
atn4z7: I don't think this is necessary since we are doing the same with the mnemonic

.yarn/patches/@bitcoinerlab-secp256k1-npm-1.2.0-1098d4b329.patch
ruijia1in: actually this is to remove it from the patch.

packages/core-mobile/app/store/app/listeners.ts
ruijia1in: removed.

packages/core-mobile/app/services/wallet/MnemonicWallet.ts
```diff
@@ -60,11 +62,37 @@ export function assertMnemonicWallet(
...
+  }
+
+  private async getSlip10Node(): Promise<HDKey> {
+    if (!this.#slip10Node) {
+      this.#slip10Node = slip10.fromMasterSeed(
+        Uint8Array.from(await this.getSeed())
+      )
+    }
+    return this.#slip10Node
+  }
```
atn4z7, line 94: do we need these to be async?

packages/core-mobile/app/store/wallet/thunks.ts
```diff
@@ -185,6 +186,9 @@ export const removeWallet = createAsyncThunk<void, string, ThunkApi>(
 
     thunkApi.dispatch(_removeWallet(walletId))
 
+    // Clear all cached data (wallet instance, public keys, xpubs) for the removed wallet
+    WalletFactory.cache.clearWallet(walletId)
```
atn4z7, line 190: we shouldn't call cache directly outside wallet code since that's an internal implementation detail. could you add a clear method to WalletFactory and use that instead?

package.json
```diff
@@ -52,23 +50,11 @@
...
-    "@noble/curves@~1.2.0": "patch:@noble/curves@patch%3A@noble/curves@npm%253A1.9.7%23./.yarn/patches/@noble-curves-npm-1.9.7-2b9efc8ab4.patch%3A%3Aversion=1.9.7&hash=9b53f1&locator=mobile-monorepo%2540workspace%253A.#./.yarn/patches/@noble-curves-patch-8bddd6a892.patch",
-    "@noble/curves@~1.1.0": "patch:@noble/curves@patch%3A@noble/curves@npm%253A1.9.7%23./.yarn/patches/@noble-curves-npm-1.9.7-2b9efc8ab4.patch%3A%3Aversion=1.9.7&hash=9b53f1&locator=mobile-monorepo%2540workspace%253A.#./.yarn/patches/@noble-curves-patch-8bddd6a892.patch",
-    "@noble/curves@^1.8.0": "patch:@noble/curves@patch%3A@noble/curves@npm%253A1.9.7%23./.yarn/patches/@noble-curves-npm-1.9.7-2b9efc8ab4.patch%3A%3Aversion=1.9.7&hash=9b53f1&locator=mobile-monorepo%2540workspace%253A.#./.yarn/patches/@noble-curves-patch-8bddd6a892.patch",
+    "@noble/curves": "patch:@noble/curves@npm%3A1.9.7#./.yarn/patches/@noble-curves-npm-1.9.7-2b9efc8ab4.patch",
```
atn4z7, line 57: nice cleanup!!

atn4z7 (approved):

onghwan (approved):
dev-tested

PR #3814, opened 2026-05-05
CP-14178: fix P/X-chain signing for imported private-key wallets

B0Y3R-AVA:
## Summary

`PrivateKeyWallet.getAvaSigner` was passing a hex private key into `Avalanche.SimpleSigner`, which only accepts a BIP-39 mnemonic. Every P/X-chain action (claim, delegate, send, import/export) for imported-key wallets threw `Invalid mnemonic phrase.`

Switched the X/P signer to `Avalanche.StaticSigner` with raw key buffers, mirroring the Core extension's pattern in `core-extension` `packages/service-worker/src/services/wallet/WalletService.ts`. Added regression tests for both `signAvalancheTransaction` and `signAvalancheMessage` on the imported-key path.

Real user reported in Slack 2026-05-04 with 52.21 AVAX stuck on P-chain after their stake matured.

Ticket: [CP-14178](https://ava-labs.atlassian.net/browse/CP-14178)

https://github.com/user-attachments/assets/54c1fa50-9a19-4fa9-8fb0-65b7449df868

## Changes

- `PrivateKeyWallet.getAvaSigner(provider)` now returns `Avalanche.StaticSigner` constructed with raw key buffers.
- `getSigner` PVM/AVM case forwards the provider through.
- `signAvalancheTransaction` `instanceof` check widened from `SimpleSigner` to `StaticSigner`.
- `signAvalancheMessage` fetches an Avalanche XP provider via `NetworkService.getAvalancheProviderXP(network.isTestnet)` since `signMessage` only receives an EVM provider.
- New `PrivateKeyWallet.test.ts` regression tests for both signing paths.

`MnemonicWallet`, `SeedlessWallet`, `LedgerWallet`, and `KeystoneWallet` are untouched.

## Dev testing 
iOS: 8411
Android: 8412

1. Install the QA build.
2. Onboard a fresh wallet via **Import private key** using a key that holds some P-chain AVAX.
3. On P-chain, send a small amount to another address. Tx should sign and broadcast (previously threw `Invalid mnemonic phrase`).
4. On P-chain, stake/delegate to any node. Signing should succeed.
5. Wipe the wallet and re-onboard with a 24-word mnemonic that has P-chain funds.
6. Repeat the P-chain send and stake to confirm the mnemonic flow still works (regression check on `MnemonicWallet`).

Watch for any `Invalid mnemonic phrase` or `invalid signer` toast or red error during steps 3, 4, and 6. There should be none.

## Test plan

- [x] `yarn test app/services/wallet/PrivateKeyWallet.test.ts` — 2/2 passing
- [x] `yarn test app/services/wallet` — 9/9 suites, 208/208 passing
- [x] `yarn lint` — 0 errors in changed files
- [x] `yarn tsc` — no new errors
- [x] Manual smoke (Android, internalDebug): imported private key — stake and send on P-chain both work
- [x] Manual regression smoke: mnemonic wallet — stake and send on P-chain both still work

[CP-14178]: https://ava-labs.atlassian.net/browse/CP-14178?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core-mobile/app/services/wallet/PrivateKeyWallet.ts
+++ b/packages/core-mobile/app/services/wallet/PrivateKeyWallet.ts
@@ -72,6 +72,10 @@ export class PrivateKeyWallet implements Wallet {
     provider: Avalanche.JsonRpcProvider
   ): Promise<Avalanche.StaticSigner> {
     const keyBuffer = Buffer.from(strip0x(this.privateKey), 'hex')
+    // An imported private key is a single secp256k1 keypair with no BIP-44
+    // derivation; the same key produces both the X/P (bech32) and C-chain
+    // (EVM) addresses, so we pass it to StaticSigner's X/P and C slots
+    // identically. Mirrors Core Extension's WalletService PrivateKey path.
     return new Avalanche.StaticSigner(keyBuffer, keyBuffer, provider)
   }
 
--- a/packages/core-mobile/app/services/wallet/PrivateKeyWallet.ts
+++ b/packages/core-mobile/app/services/wallet/PrivateKeyWallet.ts
@@ -68,9 +68,10 @@ export class PrivateKeyWallet implements Wallet {
   }
 
   private async getAvaSigner(
-    accountIndex: number
-  ): Promise<Avalanche.SimpleSigner> {
-    return new Avalanche.SimpleSigner(this.privateKey, accountIndex)
+    provider: Avalanche.JsonRpcProvider
+  ): Promise<Avalanche.StaticSigner> {
+    const keyBuffer = Buffer.from(strip0x(this.privateKey), 'hex')
+    return new Avalanche.StaticSigner(keyBuffer, keyBuffer, provider)
   }
 
   private async getSvmSigner(): Promise<SolanaSigner> {
@@ -79,7 +80,7 @@ export class PrivateKeyWallet implements Wallet {
   }
 
   private async getSigner({
-    accountIndex,
+    accountIndex: _accountIndex,
     network,
     provider
   }: {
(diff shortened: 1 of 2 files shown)


packages/core-mobile/app/services/wallet/PrivateKeyWallet.ts
```diff
@@ -68,9 +69,10 @@ export class PrivateKeyWallet implements Wallet {
   }
 
   private async getAvaSigner(
-    accountIndex: number
-  ): Promise<Avalanche.SimpleSigner> {
-    return new Avalanche.SimpleSigner(this.privateKey, accountIndex)
+    provider: Avalanche.JsonRpcProvider
+  ): Promise<Avalanche.StaticSigner> {
+    const keyBuffer = Buffer.from(strip0x(this.privateKey), 'hex')
+    return new Avalanche.StaticSigner(keyBuffer, keyBuffer, provider)
```
onghwan, line 79: nit: could we add a quick inline comment explaining why it's safe to pass `keyBuffer` as both the first and second parameters?

B0Y3R-AVA: Good call. Addressed in d2206d943 with a 4-line comment explaining that an imported private key is a single secp256k1 keypair (no BIP-44 derivation) so the same buffer goes into both the X/P and C slots, matching Core Extension's pattern.

atn4z7 (approved):

onghwan (approved):

PR #3822, opened 2026-05-07
Tag high-volume fetch errors with source for Sentry bucket splitting

B0Y3R-AVA:
### What this does
adds a source tag to 6 high-volume fetch errors (PostHog, Glacier, ProfileApi,
GasStation) so the TypeError: Network request failed Sentry bucket — currently 5,716 users in
one anonymous pile — splits into "which subsystem is actually failing".

What it doesn't do: fix anything, change UX, or reduce failure counts.

The bet: one week after deploy, the Sentry breakdown tells us whether 4WW is mostly harmless PostHog analytics retries (most likely — we can stop worrying about it) or real Glacier/ProfileApi failures (we go fix those next)
  
Phase 1 instrumentation PR. Goal: split the project-wide `TypeError: Network request failed` Sentry bucket (`CORE-REACT-NATIVE-4WW` — 5,716 users / 325k events / chronic since 2025-02-07) into per-source signals so we can act on it.

Today the issue is a black box: every fetch in the app routes through the same `whatwg-fetch:566` polyfill, so Sentry fingerprints all network failures into one giant bucket regardless of whether the failed call was PostHog analytics, a Glacier balance fetch, the gas station, profile-api, or anything else. This PR tags the highest-volume `Logger.error` callsites with a `source` tag. After deploy, Sentry filters / dashboards on `source:glacier`, `source:profile-api`, etc. surface the dominant contributors so we can target real fixes in follow-up PRs.

This PR ships zero behavior change. Errors that previously went to Sentry continue to go to Sentry. The only difference is added scope tags.

## What changed

**API extension:**
- `SentryService.captureException(message, value, tags?)` — added optional `tags?: Record<string, string>`. When provided, wraps the call in `Sentry.withScope` and `scope.setTags(tags)`. When omitted, the hot path is unchanged (no withScope overhead, no fingerprint risk).
- `Logger.error(message, value?, tags?)` — added optional 3rd arg, forwarded to `SentryService.captureException`.
- `SentryTag` extended with `PostHog`, `Glacier`, `ProfileApi`, `GasStation`. (`FusionSdk` and `AccountService` already existed.)

**Six callsites tagged** (chosen by Sentry volume + clear single source-of-failure):
- `app/services/posthog/PostHogService.ts` (capture + identifyUser) → `PostHog`
- `app/services/network/NetworkService.ts` (fetchNetworks + fetchDeBankNetworks) → `Glacier`
- `app/services/earn/EarnService.ts` (getTransformedStakesForAllAccounts) → `Glacier`
- `app/hooks/useGasless.ts` (handleGaslessTx) → `GasStation`
- `app/services/wallet/WalletService.tsx` (getAddressesForExtendedPublicKey) → `ProfileApi`

**Critical detail:** at four of these callsites the existing code interpolated the underlying error into the message string and passed `value=undefined`. That would have caused `SentryService.captureException` to synthesize a fresh `Error` and fingerprint at the SentryService line, defeating the entire goal. They now pass the underlying error as `value` so Sentry preserves the original `TypeError: Network request failed` (or whatever) type and stack — the new tagged events join the existing `4WW` issue's fingerprint with a `source` tag, instead of creating sibling issues.

## What was deliberately scoped out

- **`Deeplink` and `AppCheck` tags** — not added in this PR. Deeplink's "no url" error is a payload validation problem, not a fetch. AppCheck has its own dedicated Sentry issue groups already and tagging it requires a separate sweep of `AppCheckService` callsites. Future PR.
- **`BalanceService` Glacier callsites** — likely a much larger 4WW contributor than EarnService, but adding them needs careful auditing of 4 separate sites. Future PR.
- **`GaslessService.waitForConfirmation`** — looked tempting but it's a chain RPC call (`provider.waitForTransaction`), not a gas-station call. Tagging it `GasStation` would mislead investigation. Skipped.
- **Tighter `tags` type** — `Record<string, string>` is loose; could narrow to `{ source: SentryTagValue }`. Acceptable for this PR since `SentryTag.X` provides the convention; can tighten later.
- **`Logger.error` unit test** — couldn't get the `jest.mock` / `jest.spyOn` patterns to intercept the singleton's call to `SentryService.captureException` in this codebase's test harness. The pass-through is a one-line forward; the SentryService tests cover the receiving end. Acceptable trade-off; if it causes pain later we can revisit.

## Tests

- `app/services/sentry/SentryService.test.ts` — 5 new tests covering `captureException` with tags (Error+tags, non-Error+tags, no-tags-no-withScope, no-tags-no-setTags, isAvailable=false-with-tags-still-no-ops). 16/16 SentryService tests pass.
- `app/services/network/NetworkService.test.ts` — updated existing fetch-error test to assert tag is passed and original error object is preserved (mocked rejections changed from `'string error'` to `new TypeError('Network request failed')` to match production behavior).
- 264/264 tests pass across services/sentry, posthog, network, earn (EarnService), wallet, gasless directories. tsc clean. lint clean.

## Risk

- **Sentry fingerprint impact:** the four newly-passes-Error sites should join `4WW` (good — this is the goal). The two PostHog sites and EarnService were already passing the error correctly, so their existing Sentry issues (if any) gain a tag but keep the same fingerprint. No issue should regress in fingerprinting.
- **Backward compatibility:** Logger.error 3rd arg is optional. Every existing 2-arg call is unchanged.
- **Search & alerts:** anyone with saved Sentry searches or alert rules pinned to specific issue groups should re-check after this lands. New tagged-but-same-fingerprint events join existing groups; the issue title should remain the same.

## Dev testing

**Build numbers:** _to be filled in once Bitrise has built this branch._

Reviewer steps:

- [x] **Sanity:** open the app, normal sign-in, navigate around. No crashes, no new errors.
- [x] **Sentry verification (after a build deploys):** trigger a deliberate fetch failure (e.g. enable airplane mode, then refresh balances). Find the corresponding event in Sentry; confirm the new tag appears under "Tags" with `source: glacier`.
- [x] **Sentry filter test:** in Sentry, filter by `source:glacier` over the last 24h after deploy. New events from this branch should appear with the tag.
- [x] **Existing PostHog dashboards:** confirm dashboards reading existing analytics are unaffected (this PR doesn't change analytics events, only Sentry tagging).
- [x] **CI:** unit tests, lint, tsc all pass on this branch.

### DIFF
--- a/packages/core-mobile/app/services/network/NetworkService.ts
+++ b/packages/core-mobile/app/services/network/NetworkService.ts
@@ -21,7 +21,7 @@ import Logger from 'utils/Logger'
 import { DebankNetwork } from 'services/network/types'
 import { addIdToPromise, settleAllIdPromises } from '@avalabs/evm-module'
 import { Module } from '@avalabs/vm-module-types'
-import { SpanName } from 'services/sentry/types'
+import { SentryTag, SpanName } from 'services/sentry/types'
 import { mapToVmNetwork } from 'vmModule/utils/mapToVmNetwork'
 import { ChainInfo } from '@avalabs/glacier-sdk'
 import GlacierService from 'services/glacier/GlacierService'
@@ -39,12 +39,16 @@ class NetworkService {
     const networks = await this.fetchNetworks({
       includeSolana
     }).catch(reason => {
-      Logger.error(`[NetworkService][fetchNetworks]${reason}`)
+      Logger.error('[NetworkService][fetchNetworks] failed', reason, {
+        source: SentryTag.Glacier
+      })
       return {} as Networks
     })
 
     const deBankNetworks = await this.fetchDeBankNetworks().catch(reason => {
-      Logger.error(`[NetworkService][fetchDeBankNetworks]${reason}`)
+      Logger.error('[NetworkService][fetchDeBankNetworks] failed', reason, {
+        source: SentryTag.Glacier
+      })
       return {} as Networks
     })
 
--- a/packages/core-mobile/app/services/network/NetworkService.ts
+++ b/packages/core-mobile/app/services/network/NetworkService.ts
@@ -40,14 +40,14 @@ class NetworkService {
       includeSolana
     }).catch(reason => {
       Logger.error('[NetworkService][fetchNetworks] failed', reason, {
-        source: SentryTag.Glacier
+        source: SentryTag.Proxy
       })
       return {} as Networks
     })
 
     const deBankNetworks = await this.fetchDeBankNetworks().catch(reason => {
       Logger.error('[NetworkService][fetchDeBankNetworks] failed', reason, {
-        source: SentryTag.Glacier
+        source: SentryTag.Proxy
       })
       return {} as Networks
     })
--- a/packages/core-mobile/app/hooks/useGasless.ts
+++ b/packages/core-mobile/app/hooks/useGasless.ts
@@ -9,6 +9,7 @@ import { useNetworks } from 'hooks/networks/useNetworks'
 import { getChainIdFromCaip2 } from 'utils/caip2ChainIds'
 import GaslessService from 'services/gasless/GaslessService'
 import Logger from 'utils/Logger'
+import { SentryTag } from 'services/sentry/types'
 import NetworkService from 'services/network/NetworkService'
 import { JsonRpcBatchInternal } from '@avalabs/core-wallets-sdk'
 import { resolve } from '@avalabs/core-utils-sdk'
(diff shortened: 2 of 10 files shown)


packages/core-mobile/app/services/network/NetworkService.ts
```diff
@@ -39,12 +39,16 @@ class NetworkService {
...
+      Logger.error('[NetworkService][fetchNetworks] failed', reason, {
+        source: SentryTag.Glacier
+      })
       return {} as Networks
     })
 
     const deBankNetworks = await this.fetchDeBankNetworks().catch(reason => {
-      Logger.error(`[NetworkService][fetchDeBankNetworks]${reason}`)
+      Logger.error('[NetworkService][fetchDeBankNetworks] failed', reason, {
+        source: SentryTag.Glacier
```
onghwan: maybe we should add a new tag(i.e. `SentryTag.Proxy` or `SentryTag.NetworkService`) for `fetchNetworks` and `fetchDeBankNetworks` here. I don't think the data from Glacier.

B0Y3R-AVA: fixed

onghwan (approved):

bogdandobritoiu (approved):
dev tested

PR #3824, opened 2026-05-08
split useFeeEstimation revert-error fingerprint by on-chain selector

B0Y3R-AVA:
## Summary
- Split CORE-REACT-NATIVE-A8W (currently 311 users / 2,206 events / 90d, multi-revert umbrella) into per-revert-signature Sentry fingerprints by passing `error.details.data` (the on-chain 4-byte selector) as the fingerprint discriminator.
- Extends `SentryService.captureMessage` to accept an optional `fingerprint` arg.
- When `details.data` is missing, falls back to Sentry's `{{ default }}` token so events are still grouped by stack (otherwise calling `setFingerprint` would replace default grouping and over-collapse).
- Pure observability change. No user-facing behavior change.

## Why
A8W today collapses all `EstimateNativeFeeError`-with-details captures into a single fingerprint (Sentry titles it `Sentry.withScope$argument_0` because there's no thrown error to title with). The actual revert reasons (TargetCallFailed, FailedInnerCall, UnsupportedTokenOut, transfer amount exceeds, Panic, etc.) live in `extra.details.data` but aren't searchable. After the split each revert reason gets its own issue. The TargetCallFailed slice in particular is the mobile twin of CORE-WEB-16ZF's 147-user pre-submit OOG family (same SDK gas-estimation root cause as Fusion errorCode 5006 post-submit). Investigation: https://www.notion.so/35a0e2bd518081908722e62e029170a7

## Dev testing

**Bitrise builds:** _Fill in after CI runs._

- [ ] iOS build: 8515
- [ ] Android build: 8516

**Steps:**
1. Trigger the swap form on AVAX C-Chain with a token whose pre-submit estimation reverts (any low-balance wallet attempting a Markr swap will reproduce). Confirm the app continues to surface the existing inline validation error to the user — no UX change.
2. Sentry inspection (after build is published and an event fires): the new event should appear in a per-revert-signature issue (e.g. a separate fingerprint for TargetCallFailed `0xeda86850`) instead of all going into A8W. Events that lack a `details.data` selector should fall back to Sentry's default stack-based grouping (not collapse into a single bucket).

## Related
- Jira: [CP-14260](https://ava-labs.atlassian.net/browse/CP-14260) (1 of 3 stacked PRs)

[CP-14260]: https://ava-labs.atlassian.net/browse/CP-14260?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/core-mobile/app/new/features/swap/hooks/useFeeEstimation.ts
```diff
@@ -79,7 +80,8 @@ export const useFeeEstimation = ({
       SentryService.captureMessage(
         '[useFeeEstimation] estimateNativeFee revert error',
         { ...error.details, cause: error.cause },
-        { source: SentryTag.FusionSdk }
+        { source: SentryTag.FusionSdk },
+        getFingerprintForFeeEstimationError(error)
```
onghwan: can we generalize this with something like this?

```
export const buildSentryFingerprint = (
  scope: string,
  discriminator?: string | null
): string[] =>
  typeof discriminator === 'string' && discriminator.length > 0
    ? [scope, discriminator]
    : ['{{ default }}']
```

and then we can just call

```ts
SentryService.captureMessage(
        '[useFeeEstimation] estimateNativeFee revert error',
        { ...error.details, cause: error.cause },
        { source: SentryTag.FusionSdk },
        buildSentryFingerprint(`useFeeEstimation`, error.details.data)
```

I guess we don't need to check 
```
    typeof error === 'object' &&
    'details' in error &&
    error.details &&
    typeof error.details === 'object' &&
    'data' in error.details &&
    typeof error.details.data === 'string'
```
because we have guards for !error and isEstimateNativeFeeError above.

onghwan (approved):

ruijia1in (approved):

PR #3831, opened 2026-05-11
CP-14006: fix android `PushNotificationPressed` under-reporting

onghwan:
## Description

**Ticket: [CP-14006]**

### Summary
Fixes Android `PushNotificationPressed` analytics under-reporting and brings parity to all four notification capture paths (foreground, warm-background, cold-start data-only, iOS background).

- **Cold-start (Android data-only)**: previously the press was lost because we only checked `messaging().getInitialNotification()`. Now we also check `notifee.getInitialNotification()`, which is where data-only notifications surface, and prefer it as the source of truth.
- **Warm background (Android data-only)**: notifee's background handler is now registered exactly once at app entry (`index.js`) — notifee only supports a single handler. The headless task captures analytics, stashes the notification data in `pendingBackgroundPress`, and a new `AppState` 'active' listener in `DeeplinkContext` consumes it on the next foreground transition so the deeplink runs in a mounted React context.
- **iOS background**: `PushNotificationPressed` capture in `FCMService.handleBackgroundMessageIos` was previously gated behind `shouldSkipHandlingDeeplink` (dropped balance-change taps) and behind a strict `data.channelId` guard. Capture is now performed before the skip decision and uses the shared channel-id resolver, so balance-change and walletconnect taps are no longer dropped.
- **Two new analytics dimensions** added to `PushNotificationPressed`:
  - `appState`: `'foreground' | 'background' | 'cold_start'`
  - `handler`: `'notifee' | 'fcm'`
  These let the dashboard slice volume by capture path independently and make future regressions detectable.
- **Refactors** to support the above:
  - New `services/notifications/eventChannelMap.ts` holding `EVENT_TO_CH_ID` and a pure `resolveChannelId(...)` helper used by every capture site. Breaks the prior `FCMService` ↔ `NotificationsService` import cycle.
  - Idempotence guard on `registerBackgroundNotificationHandler` to prevent silent overwrite if a future call site reintroduces the registration.
  - Conditional cold-start `pendingBackgroundPress` drain to avoid losing a legitimate warm-background press if the `getInitialNotification` effect re-runs (e.g. on `isAllNotificationsBlocked` toggle).
  - Headless background handler body wrapped in `try/catch` so a synchronous throw cannot escape as an unhandled rejection.

## Testing
iOS: 8480
Android: 8481

1. **Background path** — open the app, then minimize. Tap the notification → app foregrounds, deeplink routes correctly. Verify PostHog event:
   - **Android**: `appState: 'background'`, `handler: 'notifee'`
   - **iOS**: `appState: 'background'`, `handler: 'fcm'`
2. **Cold-start path** — kill the app from recents. Tap the notification → app cold-starts, deeplink routes correctly. Verify PostHog event:
   - **Android**: `appState: 'cold_start'`, `handler: 'notifee'`
   - **iOS**: `appState: 'cold_start'`, `handler: 'fcm'`

## Checklist

[CP-14006]: https://ava-labs.atlassian.net/browse/CP-14006?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core-mobile/app/types/analytics.ts
+++ b/packages/core-mobile/app/types/analytics.ts
@@ -215,17 +215,30 @@ export type AnalyticsEvents = {
     channelId: string
     deeplinkUrl?: string
     /**
-     * Which capture path the press came from. Added for CP-14006 verification.
+     * App state at the moment the press was received.
      *
-     * - `foreground`:      notifee.onForegroundEvent PRESS while the app is open
-     * - `ios_background`:  messaging().onNotificationOpenedApp (iOS background tap)
-     * - `notifee_initial`: notifee.getInitialNotification() — cold start tap on a
-     *                      notification displayed by notifee (Android data-only)
-     * - `fcm_initial`:     messaging().getInitialNotification() — cold start tap
-     *                      on a notification displayed by the FCM SDK directly
-     *                      (legacy notification payload)
+     * - `foreground`: app is open and in view.
+     * - `background`: app is alive but minimized (warm resume on tap).
+     * - `cold_start`: app was killed and was launched by the press.
+     *
+     * Combined with `handler` to identify the exact capture path. Added for
+     * CP-14006 verification — lets us break down underreporting by app state
+     * regardless of which RN API caught the press.
      */
-    source: 'foreground' | 'ios_background' | 'notifee_initial' | 'fcm_initial'
+    appState: 'foreground' | 'background' | 'cold_start'
+    /**
+     * Which RN notification API delivered the press to us.
+     *
+     * - `notifee`: notifee.onForegroundEvent / onBackgroundEvent /
+     *              getInitialNotification — used for all Android data-only
+     *              notifications and for foreground notifications on both
+     *              platforms.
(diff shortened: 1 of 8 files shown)


packages/core-mobile/app/types/analytics.ts
```diff
@@ -214,6 +214,31 @@ export type AnalyticsEvents = {
...
+    /**
+     * Which RN notification API delivered the press to us.
+     *
+     * - `notifee`: notifee.onForegroundEvent / onBackgroundEvent /
+     *              getInitialNotification — used for all Android data-only
+     *              notifications and for foreground notifications on both
+     *              platforms.
+     * - `fcm`:     messaging().onNotificationOpenedApp /
+     *              getInitialNotification — used when the FCM SDK displays
+     *              the notification itself (iOS APNs alert path, or legacy
```
B0Y3R-AVA, line 238: love this

B0Y3R-AVA (approved):

ruijia1in (approved):
dev-tested

PR #3861, opened 2026-05-29
Extract EVM injected provider dispatch into router module

B0Y3R-AVA:
## Description

**Ticket: [CP-14365](https://ava-labs.atlassian.net/browse/CP-14365)**

- **Summary**: Extracts the EVM injected-provider message dispatch out of the `useEvmInjectedProvider` hook into a standalone, framework-agnostic module at `app/hooks/browser/injectedProvider/`:
  - `router.ts` — `createInjectedProviderRouter(deps)` factory holding all dispatch logic (payload parsing/validation, allow-list gating, origin checks, `wallet_switchEthereumChain` / `wallet_addEthereumChain` / `wallet_revokePermissions` / `wallet_watchAsset`, signing dispatch, and read-only RPC proxying).
  - `types.ts` — shared `RouterDeps`, `ProviderRequest`, `BrowserNetwork`, `DomainMetadata`, and `MAX_MESSAGE_SIZE`.
  - `router.test.ts` — 22 unit tests covering dispatch, payload parsing, each `wallet_*` method, signing, and the RPC proxy.
  - The hook now wires React state into the router via stable getter callbacks (`getBrowserNetwork`, `getAllNetworks`, `getNativeOrigin`, `getPeerMeta`, etc.) and delegates `handleProviderMessage` to it.
- **Motivation / context**: The dispatch logic had grown to ~400 lines of `useCallback`s inside the hook with no direct unit coverage and was hard to test in isolation. Pulling it into a plain factory makes the JSON-RPC handling unit-testable without rendering the hook, and keeps the CP-14159 origin-spoofing protections (approval-method origin gating, `peerMeta` hardening) in one place.
- **Behavior**: No functional change intended — this is a structural refactor. The same methods are handled with the same responses and the same CP-14159 origin guarantees.
- **Dependencies introduced**: None.
- **Breaking changes**: None.

## Screenshots/Videos

N/A — pure refactor with no UI change.

## Testing

**Dev Testing (if applicable)**

iOS: 8694
Android: 8695
- `yarn core test app/hooks/browser/injectedProvider/router.test.ts` — 22/22 passing.
- In-app browser regression (no behavior change expected):
  - Connect to a dApp (e.g. via ConnectKit/wagmi) and confirm read-only calls (`eth_call`, `eth_blockNumber`) resolve.
  - `wallet_switchEthereumChain` to a known chain switches the tab network; to an unknown chain returns `4902` (triggers add-chain flow) without a switch loop.
[...]

PR #3862, opened 2026-05-29
Require approval for EVM dApp connections in in-app browser

B0Y3R-AVA:
## Description

**Ticket: [CP-14366](https://ava-labs.atlassian.net/browse/CP-14366)**

Adds Phase 1(b) of the injected-provider modernization plan. Previously, a dApp loaded in the in-app browser could call `eth_requestAccounts` and silently receive the user's active address with no prompt and no record. This PR makes those calls round-trip to the React Native app, open an approval UI, persist the grant, and replay it on return visits so dApps auto-reconnect.

**Stacked PR:** base is `boyer/injected-provider-updates` (Phase 1a — #3861). Review/merge #3861 first; GitHub will retarget this onto `main` once it lands. This branch was rebased onto the current Phase 1a head and squashed to a single commit.

**Summary of changes:**

- NEW `app/store/permissions/` slice — per-`(domain, address, vmType)` grants with `grantPermission` / `revokePermission` actions and `selectHasPermission` / `selectConnectedDomains` / `selectAddressesForDomain` selectors. Persisted via existing redux-persist with the `EncryptThenMac` transform. 14 slice tests.
- NEW `features/injectedProvider/screens/AuthorizeInjectedDappScreen.tsx` — connect approval modal. Duplicates the visual layout of the existing `AuthorizeDappScreen` (DappLogo + peer + SelectAccounts + ActionSheet) with injected-specific plumbing (reads `InjectedAuthParams` from `walletConnectCache`, calls `onApprove(selectedAccounts)` / `onReject()` to resolve the router's parked promise).
- NEW modal route `(signedIn)/(modals)/authorizeInjectedDapp/` — registered in `(signedIn)/_layout.tsx`.
- EXTENDED `walletConnectCache` with an `injectedAuthParams` slot and corresponding type. The HOC (`withWalletConnectCache`) is slot-agnostic, so the screen consumes the new slot without any HOC changes.
- EXTENDED router (from Phase 1a) with 4 new handlers: `handleRequestAccounts`, `handleRequestPermissions`, `handleGetPermissions`, and a rewritten `handleRevokePermissions` that dispatches into the new slice. Method routing was extracted into a `dispatchMethod` helper. Router tests grew from 22 → 33.
- EXTENDED `useEvmInjectedProvider` hook with (a) `requestConnectApproval` (a `useCallback`) that sets the cache, navigates, parks a promise, and adds a 90s timeout + settled-guard + reject-prior-in-flight (so a stuck screen can't hang the dApp or pile up concurrent requests), and (b) a permission-aware `handleDomainMetadata` that primes `_accounts` via `accountsChanged` on page load when a grant already exists for the current origin.
- MODIFIED `evmProviderShim.ts` — removed the 4 inpage implementations of `eth_requestAccounts` / `wallet_requestPermissions` / `wallet_getPermissions` / `wallet_revokePermissions`. Removed the pre-seeded `_accounts = [_address]` init — the shim now starts with an empty array and native primes it based on permissions (the `address` param is retained for caller compatibility but no longer embedded).
- EXTENDED `useConnectedDapps` — `Dapp` is now a discriminated union of `WCDapp | InjectedDapp`, and the hook merges injected grants from the permissions slice with WalletConnect sessions. `connectedSites.tsx` updated to render both shapes and route disconnects through the new `killDapp` / `killAllDapps` API.
- MIGRATION 30 — initializes `permissions: { grants: {} }` for existing persisted state (persist `VERSION` bumped to 30).

**Motivation:** closes the biggest security gap in the in-app browser. Before this PR, every site the user visits silently receives the active wallet address on `eth_requestAccounts` with zero friction. After: an approval is required on first visit, grants are tracked per-origin, and users can review / revoke from the existing Connected Sites UI.

**Dependencies:** none.

**Breaking:** no. Feature is behind a feature flag and has never shipped to production users.

## Screenshots/Videos

No screenshots — visual parity with the existing WalletConnect `AuthorizeDappScreen` (same DappLogo, account picker, ActionSheet). The only user-visible change is that the modal now also appears when connecting via the in-app browser rather than only via WalletConnect.

## Testing
iOS: 8667
Android: 8668

**Dev Testing — validated end-to-end on Android (Pixel 8 Pro):**

| # | Test | Site | Result |
|---|------|------|--------|
| 1 | Connect approval modal appears on first visit | opensea.io / app.uniswap.org / etherscan.io / revoke.cash | ✅ |
| 2 | Approve with account → dApp sees connected | Uniswap, OpenSea | ✅ |
| 3 | Auto-reconnect on return visit (different tab) | Uniswap | ✅ |
| 4 | `personal_sign` via existing signing path | Etherscan Verified Signature | ✅ |
| 5 | `eth_signTypedData_v4` | OpenSea | ✅ |
| 6 | `wallet_switchEthereumChain` (optimistic chainChanged preserved; no ConnectKit loop) | Uniswap | ✅ |
| 7 | `wallet_addEthereumChain` end-to-end | chainlist.org → Gnosis Chain | ✅ |
| 8 | `eth_sendTransaction` via connect-gated flow | revoke.cash revoke allowance | ✅ |
| 9 | Cancel modal → dApp gets clean 4001 | app.aave.com | ✅ |
| 10 | Revoke via Settings → Connected Sites → dApp shows disconnected | Uniswap | ✅ |
| 11 | Concurrent connects across tabs (prior request rejected, new one proceeds) | Two fresh tabs | ✅ |

> Note: the dev-testing matrix above was validated on the pre-rebase branch. After rebasing onto the current Phase 1a head, the change was re-verified statically (below); a fresh device pass is recommended before merge.

**Known expected behaviors (not regressions):**

- Switching active wallet account mid-session with a dApp that has a grant ONLY for the prior active address will cause the dApp to re-prompt on next interaction — our grant model is per-address. Loosening this to match MetaMask's broader "per-domain" semantics is scoped to a later polish pass (Phase 3d).
- login.xyz's Sign-In with Ethereum demo shows an empty WalletConnect widget because the site forces WC and its WC project ID is unresponsive from the in-app browser. Unrelated to injected-provider behavior.

**Static (re-run after rebase):**

- `yarn tsc` — ✅ 0 errors
- `yarn lint` (`--max-warnings=0` on all touched files) — ✅ 0 errors, 0 warnings
- Tests — ✅ 33 router + 15 permissions + hook/shim suites, 722 passing across the affected suites
- Trigger a Bitrise build and reference it here: _<build link TBD>_
- Move the ticket into the "Testing" column on Jira.

## Checklist

[CP-14159]: https://ava-labs.atlassian.net/browse/CP-14159?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[CP-14366]: https://ava-labs.atlassian.net/browse/CP-14366?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core-mobile/app/hooks/browser/injectedProvider/router.ts
+++ b/packages/core-mobile/app/hooks/browser/injectedProvider/router.ts
@@ -1,9 +1,11 @@
-import { rpcErrors } from '@metamask/rpc-errors'
+import { providerErrors, rpcErrors } from '@metamask/rpc-errors'
+import { NetworkVMType } from '@avalabs/core-chains-sdk'
 import { RpcMethod } from '@avalabs/vm-module-types'
 import Logger from 'utils/Logger'
 import { getEvmCaip2ChainId } from 'utils/caip2ChainIds'
 import { setTabChainId } from 'store/browser/slices/tabs'
 import { fetch as nitroFetch } from 'react-native-nitro-fetch'
+import { isUserRejectedRpcError } from './errors'
 import { MAX_MESSAGE_SIZE, ProviderRequest, RouterDeps } from './types'
 
 const READ_ONLY_METHODS = new Set([
@@ -41,9 +43,12 @@ const SIGNING_METHODS: Record<string, RpcMethod> = {
 const ALLOWED_METHODS = new Set([
   ...READ_ONLY_METHODS,
   ...Object.keys(SIGNING_METHODS),
+  'eth_requestAccounts',
+  'wallet_requestPermissions',
+  'wallet_getPermissions',
+  'wallet_revokePermissions',
   'wallet_switchEthereumChain',
   'wallet_addEthereumChain',
-  'wallet_revokePermissions',
   'wallet_watchAsset'
 ])
 
@@ -56,6 +61,23 @@ const APPROVAL_METHODS = new Set<string>([
   'wallet_addEthereumChain',
   'wallet_watchAsset'
 ])
+
+// EIP-2255 — only eth_accounts is supported; no other restricted methods today.
+function buildAccountsPermission(addresses: string[]): unknown[] {
+  if (addresses.length === 0) return []
+  return [
+    {
+      parentCapability: 'eth_accounts',
+      date: Date.now(),
+      caveats: [
+        {
+          type: 'restrictReturnedAccounts',
+          value: addresses
+        }
+      ]
+    }
+  ]
+}
 
 function validateProviderRequest(data: unknown): data is ProviderRequest {
   if (typeof data !== 'object' || data === null) return false
@@ -136,7 +158,12 @@ export function createInjectedProviderRouter(
     emitEvent,
     getNativeOrigin,
     trackPendingOrigin,
-    getPeerMeta
+    getPeerMeta,
+    getActiveAccount,
+    hasPermission,
+    grantPermission,
+    revokePermission,
+    requestConnectApproval
   } = deps
 
   const proxyToRpc = async (
@@ -293,7 +320,161 @@ export function createInjectedProviderRouter(
     }
   }
 
+  // EIP-2255: resolve permissions for the current origin against the active
+  // EVM account. Today only `eth_accounts` is supported — there is no notion
+  // of a restricted method beyond account access.
+  const resolveActiveAccountAddresses = (origin: string): string[] => {
+    const active = getActiveAccount()
+    if (!active?.addressC) return []
+    const granted = hasPermission({
+      domain: origin,
+      address: active.addressC,
+      vmType: NetworkVMType.EVM
+    })
+    return granted ? [active.addressC] : []
+  }
+
+  const handleRequestAccounts = async (id: number): Promise<void> => {
+    const origin = getNativeOrigin()
+    if (!origin) {
+      sendResponse(
+        id,
+        providerErrors.unauthorized('Origin unavailable — cannot connect'),
+        undefined
+      )
+      return
+    }
+    const active = getActiveAccount()
+    if (!active?.addressC) {
+      sendResponse(
+        id,
+        providerErrors.unauthorized('No active account'),
+        undefined
+      )
+      return
+    }
+
+    // Already connected: return without prompting.
+    if (
+      hasPermission({
+        domain: origin,
+        address: active.addressC,
+        vmType: NetworkVMType.EVM
+      })
+    ) {
+      sendResponse(id, null, [active.addressC])
+      return
+    }
+
+    try {
+      const peerMeta = getPeerMeta()
+      const selected = await requestConnectApproval(peerMeta)
+      const addresses: string[] = []
+      for (const account of selected) {
+        if (!account.addressC) continue
+        grantPermission({
+          domain: origin,
+          address: account.addressC,
+          vmType: NetworkVMType.EVM
+        })
+        addresses.push(account.addressC)
+      }
+      if (addresses.length === 0) {
+        sendResponse(id, providerErrors.userRejectedRequest(), undefined)
+        return
+      }
+      // Emit accountsChanged before resolving the Promise — matches
+      // MetaMask/Rabby ordering so wagmi listeners see the event alongside
+      // the resolution rather than one render later.
+      emitEvent('accountsChanged', addresses)
+      sendResponse(id, null, addresses)
+    } catch (e) {
+      sendResponse(id, e, undefined)
+    }
+  }
+
+  // EIP-2255: `wallet_requestPermissions` is user-visible permissions gate.
+  // Behaves identically to `eth_requestAccounts` in our scope (only capability
+  // is `eth_accounts`); response shape differs (array of Permission objects).
+  const handleRequestPermissions = async (id: number): Promise<void> => {
+    const origin = getNativeOrigin()
+    if (!origin) {
+      sendResponse(
+        id,
+        providerErrors.unauthorized('Origin unavailable — cannot connect'),
+        undefined
+      )
+      return
+    }
+    const active = getActiveAccount()
+    if (!active?.addressC) {
+      sendResponse(
+        id,
+        providerErrors.unauthorized('No active account'),
+        undefined
+      )
+      return
+    }
+
+    const alreadyGranted = hasPermission({
+      domain: origin,
+      address: active.addressC,
+      vmType: NetworkVMType.EVM
+    })
+
+    if (alreadyGranted) {
+      sendResponse(id, null, buildAccountsPermission([active.addressC]))
+      return
+    }
+
+    try {
+      const peerMeta = getPeerMeta()
+      const selected = await requestConnectApproval(peerMeta)
+      const addresses: string[] = []
+      for (const account of selected) {
+        if (!account.addressC) continue
+        grantPermission({
+          domain: origin,
+          address: account.addressC,
+          vmType: NetworkVMType.EVM
+        })
+        addresses.push(account.addressC)
+      }
+      if (addresses.length === 0) {
+        sendResponse(id, providerErrors.userRejectedRequest(), undefined)
+        return
+      }
+      // Emit accountsChanged before resolving the Promise (see handleRequestAccounts).
+      emitEvent('accountsChanged', addresses)
+      sendResponse(id, null, buildAccountsPermission(addresses))
+    } catch (e) {
+      sendResponse(id, e, undefined)
+    }
+  }
+
+  const handleGetPermissions = (id: number): void => {
+    const origin = getNativeOrigin()
+    if (!origin) {
+      sendResponse(id, null, [])
+      return
+    }
+    sendResponse(
+      id,
+      null,
+      buildAccountsPermission(resolveActiveAccountAddresses(origin))
+    )
+  }
+
   const handleRevokePermissions = (id: number): void => {
+    const origin = getNativeOrigin()
+    if (origin) {
+      // Whole-domain revoke (address omitted) — drops grants for every
+      // address previously approved at this origin, not just the active
+      // one. Matches MetaMask's "Disconnect this site" UX. When multi-
+      // account selection is introduced, revisit whether a narrower revoke
+      // is warranted.
+      revokePermission({ domain: origin })
+    }
     // Only emit accountsChanged([]) — NOT disconnect. Per EIP-1193,
     // 'disconnect' means the provider lost network connectivity, not
     // that the user revoked access. Emitting disconnect causes wagmi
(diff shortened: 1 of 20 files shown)


packages/core-mobile/app/hooks/browser/useEvmInjectedProvider.ts
```diff
@@ -188,12 +208,81 @@ export function useEvmInjectedProvider(
...
+          resolve(selected)
+        }
+        const safeReject = (err: unknown): void => {
+          if (settled) return
+          settled = true
+          if (prevInflightConnectReject.current === reject) {
+            prevInflightConnectReject.current = null
+          }
+          reject(err)
+        }
```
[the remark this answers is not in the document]
B0Y3R-AVA: Good catch — fixed in a8b48040a. The ref was assigned `safeReject` but the settle-time guards compared against the raw executor `reject`, so the identity check was always false and the ref never cleared on settle. Both guards now compare against `safeReject`, so a settled request clears its own ref while preserving a newer in-flight one. (Behavior was masked before — the stale handler is `settled`-guarded and idempotent — but the guard logic was dead.)

packages/core-mobile/app/hooks/browser/useEvmInjectedProvider.ts
```diff
@@ -188,12 +208,87 @@ export function useEvmInjectedProvider(
...
+        appRouter.navigate('/authorizeInjectedDapp')
+
+        // Safety net: if the screen never mounts or calls back within 90s,
+        // reject so the dApp can surface an error instead of hanging.
+        setTimeout(() => {
+          safeReject({
+            code: -32603,
+            message: 'Connect prompt timed out'
+          })
+        }, CONNECT_APPROVAL_TIMEOUT_MS)
```
onghwan, line 297: Should we clear this timeout once the promise settles? Right now the callback always fires at 90s and just no-ops via settled. Minor, but it keeps the closure alive for 90s per request.

packages/core-mobile/app/hooks/browser/useEvmInjectedProvider.ts
```diff
@@ -188,12 +208,87 @@ export function useEvmInjectedProvider(
     }
   }, [])
 
+  const activeAccountRef = useRef(activeAccount)
+  useEffect(() => {
+    activeAccountRef.current = activeAccount
+  }, [activeAccount])
+
+  // Tracks the reject fn of an in-flight connect approval so a later request
+  // (or an unmount) can unstick it. Writes and clears happen only inside
```
onghwan, line 231: are we currently unsticking the in-flight connect on unmount?

packages/core-mobile/app/store/permissions/slice.ts
```diff
@@ -0,0 +1,117 @@
...
+        if (domainGrants[address]?.length === 0) {
+          delete domainGrants[address]
+        }
+      }
+
+      if (Object.keys(domainGrants).length === 0) {
+        delete state.grants[domain]
+      }
+    },
+    revokeAllGrantsForAddress: (state, action: PayloadAction<Address>) => {
```
atn4z7: looks like revokeAllGrantsForAddress is not being called anywhere. do we still need this?

packages/core-mobile/app/hooks/browser/useEvmInjectedProvider.ts
B0Y3R-AVA: nice catch, added a clean up hook for `prevInflightConnectReject.current` when the user unmounts. Inactive tabs stay mounted so this will fire when the user closes the browser tab

packages/core-mobile/app/store/schemaMigration/schemaMigrations.ts
```diff
@@ -553,5 +553,12 @@ export const migrations = {
         }
       }
     }
+  },
+  30: (state: any) => {
```
atn4z7: don't think we need this migration. the permission slice is a new addition so we shouldn't need to bump the version to 30.

packages/core-mobile/app/store/permissions/slice.ts
B0Y3R-AVA: so my original thinking here was that we'd want to clean up any dapp connections we had for accounts that we deleted. but as this stands now its not very flexible. its only taking in one address at a time (which kindave makes sense as im just bulding for EVM up front) 

after thinking about this some more i've put together a better plan in a ticket here: https://ava-labs.atlassian.net/browse/CP-14374

i'll remove this for now and follow up with the ticket in a future PR. Please take a look at that ticket and leave some comments on it if you have any better ideas on handling this.

packages/core-mobile/app/new/features/injectedProvider/screens/AuthorizeInjectedDappScreen.tsx
```diff
@@ -0,0 +1,124 @@
...
+                width: SCREEN_WIDTH * 0.7,
+                marginBottom: 24
+              }}
+              numberOfLines={2}>
+              {peerMeta.name}
+            </Text>
+            <Text
+              variant="body1"
+              style={{ textAlign: 'center', width: SCREEN_WIDTH * 0.85 }}>
+              <Text variant="body1" style={{ fontWeight: '600' }}>
```
atn4z7, line 89: It looks like the picker and the rest of the injected pipeline don't line up.
  
**AuthorizeInjectedDappScreen** lets you select multiple accounts, and handleRequestAccounts grants + returns all of them. But downstream everything only uses the active account: the fast-path checks `hasPermission` for `active.addressC`, `handleDomainMetadata` primes only `[active.addressC]`, and for in-app requests `handleRequestViaVMModule` pins signing to `activeAccount.index`— so from is ignored.
  
So if someone connects with a non-active account B: they get re-prompted next visit (active account A has no grant), and the dApp thinks it's talking to B while A is the one that actually signs.

Worth noting this differs from WalletConnect — WC requests skip that branch, so WC can sign for a non-active account. On the injected side it's always the active one, so the picker is offering a choice the pipeline can't honor, and you can hit it just by using the screen normally.
  
Is this expected for now, and is there a plan to adjust it?

B0Y3R-AVA: this is expected for now, the proper behavior will be addressed in an upcoming pr ticket here https://ava-labs.atlassian.net/browse/CP-14382

packages/core-mobile/app/hooks/browser/useEvmInjectedProvider.ts
```diff
@@ -330,241 +208,179 @@ export function useEvmInjectedProvider(
...
+          // fallbackTimer would be no op, no need to maintain
+          clearTimeout(fallbackTimer)
+        }
+        prevInflightConnectReject.current = safeReject
+
+        walletConnectCache.injectedAuthParams.set({
+          peerMeta,
+          onApprove: safeResolve,
+          onReject: () =>
+            safeReject({
```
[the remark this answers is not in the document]
B0Y3R-AVA: Good catch — this is a real cross-tab edge. `injectedAuthParams` is a single global cache slot read once by the authorize screen, while `prevInflightConnectReject` is per-tab, so two tabs calling `eth_requestAccounts` near-simultaneously can have tab B clobber the slot while tab A's screen is already mounted, orphaning one parked promise until its 90s timeout. Impact is low (delayed timeout-reject, no wrong-grant/crash/security issue) and the same-tab supersede + unmount guards already cover the common paths, so rather than expand this PR I've filed a follow-up under the epic to coordinate connect approvals globally (module-level registry keyed by request, or unique-key navigation): https://ava-labs.atlassian.net/browse/CP-14385

packages/core-mobile/app/store/permissions/slice.ts
```diff
@@ -0,0 +1,103 @@
...
+  (domain: Domain) =>
+  (state: RootState): Address[] => {
+    const domainGrants = state.permissions.grants[domain]
+    if (!domainGrants) return []
+    return Object.keys(domainGrants)
+  }
+
+export const { grantPermission, revokePermission } = permissionsSlice.actions
+
+export const permissionsReducer = permissionsSlice.reducer
```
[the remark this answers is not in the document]
B0Y3R-AVA: The slice intentionally exports only `grantPermission` / `revokePermission`. `revokeAllGrantsForAddress` was removed (it was inflexible — single-address, EVM-only) and the broader multi-chain/all-wallet-types design is tracked in https://ava-labs.atlassian.net/browse/CP-14374. The PR description has since been updated to match, so the slice and the description are now consistent — no `revokeAllGrantsForAddress` in either.

packages/core-mobile/app/store/permissions/types.ts
```diff
@@ -0,0 +1,14 @@
+import { NetworkVMType } from '@avalabs/core-chains-sdk'
```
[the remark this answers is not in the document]
B0Y3R-AVA: Fixed in d6ecdb613 — switched to `import type { NetworkVMType }`. It's only used in type position (`NetworkVMType[]`), so this matches the repo's type-only-import pattern and avoids the runtime import under Babel TS.

packages/core-mobile/app/hooks/browser/injectedProvider/router.ts
```diff
@@ -293,7 +320,171 @@ export function createInjectedProviderRouter(
...
+      return
+    }
+
+    try {
+      const peerMeta = getPeerMeta() ?? {
+        name: origin,
+        description: '',
+        url: origin,
+        icons: []
+      }
```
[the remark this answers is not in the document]
B0Y3R-AVA: Fixed in d6ecdb613 — removed the dead `?? {…}` fallback. `getPeerMeta` always constructs and returns a `PeerMeta` (`name: hostname || 'Unknown site'`, never undefined), so `RouterDeps.getPeerMeta: () => PeerMeta` is correct and the fallback was unreachable. The other call sites already use the bare `getPeerMeta()`, so this also makes it consistent.

packages/core-mobile/app/hooks/browser/injectedProvider/router.ts
```diff
@@ -293,7 +320,171 @@ export function createInjectedProviderRouter(
...
```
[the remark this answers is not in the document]
B0Y3R-AVA: Fixed in d6ecdb613 — same dead-fallback removal as the :375 occurrence; both now call `getPeerMeta()` directly.

onghwan (approved):

atn4z7 (approved):

PR #3877, opened 2026-06-03
APPSEC-330: Update address bar management

chriselbring-avalabs:
## Description

**Ticket: [APPSEC-330]**

Please provide:

- A summary of the changes:
Update injected JS to handle window.open state sync w/ address bar 
Update spa navigation logic
- Motivation and context for this change
To prevent address bar spoofing 
- Any dependencies introduced
- (If breaking) migration steps and instructions

## Screenshots/Videos

## Testing

- Provide steps to test the happy path of your feature
- Provide steps to test edge cases and error states
- Trigger a build on bitrise and reference it here

## Checklist

[APPSEC-330]: https://ava-labs.atlassian.net/browse/APPSEC-330?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core-mobile/app/hooks/browser/useInjectedJavascript.ts
+++ b/packages/core-mobile/app/hooks/browser/useInjectedJavascript.ts
@@ -22,6 +22,7 @@ export type InjectedJsMessageWrapper = {
     | 'page_styles'
     | 'log'
     | 'walletConnect_deeplink_blocked'
+    | 'window_open'
     | 'provider_request'
     | 'domain_metadata'
   payload: string
@@ -186,20 +187,30 @@ export function useInjectedJavascript(): InjectedJavascripts {
     }, 500); //add delay to make sure we don't get overridden by something else
   })();`
 
-  // blocks dapp redirects to walletconnect deeplinks
-  // after walletConnect sent a session request
-  const injectCustomWindowOpen = `(async function(){
-    const originalWindowOpen = window.open;
+  // Routes window.open through the native side instead of letting the WebView
+  // handle it. On Android with setSupportMultipleWindows=false, calling the
+  // platform window.open updates the WebView's visited history (and therefore
+  // our address bar) even when no actual page navigation replaces the content
+  // — that mismatch lets a malicious page render attacker UI under a trusted
+  // URL. We always intercept and hand the URL to the native tab manager so the
+  // address bar and the rendered content stay in sync.
+  const injectCustomWindowOpen = `(function(){
     window.open = function(url, target, features){
-      if (url.startsWith('core://wc?requestId')){
-        const message = {
-          method: 'walletConnect_deeplink_blocked',
+      try {
+        if (typeof url !== 'string' || url.length === 0) return null;
+        if (url.indexOf('core://wc?requestId') === 0){
+          window.ReactNativeWebView.postMessage(JSON.stringify({
+            method: 'walletConnect_deeplink_blocked',
+            payload: url
+          }));
+          return null;
+        }
+        window.ReactNativeWebView.postMessage(JSON.stringify({
+          method: 'window_open',
           payload: url
-        };
-        window.ReactNativeWebView.postMessage(JSON.stringify(message));
-        return null;
-      }
-      return originalWindowOpen(url, target, features);
+        }));
+      } catch (e) {}
+      return null;
     }
   })();`
 
--- a/packages/core-mobile/app/new/features/browser/components/BrowserTab.tsx
+++ b/packages/core-mobile/app/new/features/browser/components/BrowserTab.tsx
@@ -86,7 +86,7 @@ export const BrowserTab = forwardRef<BrowserTabRef, { tabId: string }>(
       providerShimJs,
       handleProviderMessage,
       handleDomainMetadata,
-      setCurrentUrl
+      handleCommittedUrl
     } = useEvmInjectedProvider(webViewRef, tabId)
 
     const isInjectedProviderBlocked = useSelector(
@@ -404,12 +404,16 @@ export const BrowserTab = forwardRef<BrowserTabRef, { tabId: string }>(
     )
 
     // Surfaces a URL we trust as actually-rendered to the user-facing URL bar
-    // and to the EVM provider's origin tracker. Idempotent per URL.
+    // and to the EVM provider's origin tracker. The address-bar/history sync
+    // is idempotent per URL, but the provider must be notified on every
+    // commit: a reload keeps the same URL yet creates a fresh document whose
+    // shim _accounts cache starts empty and needs re-priming.
     const syncCommittedUrl = (url: string, title?: string): void => {
       if (!url || url.startsWith('about:')) return
-      if (lastSyncedUrlRef.current === url) return
 
-      setCurrentUrl(url)
+      handleCommittedUrl(url)
+
+      if (lastSyncedUrlRef.current === url) return
       lastSyncedUrlRef.current = url
 
       dispatch(
--- a/packages/core-mobile/app/new/features/browser/components/BrowserTab.tsx
+++ b/packages/core-mobile/app/new/features/browser/components/BrowserTab.tsx
@@ -36,6 +36,7 @@ import AnalyticsService from 'services/analytics/AnalyticsService'
 import WalletConnectService from 'services/walletconnectv2/WalletConnectService'
 import {
   addHistoryForActiveTab,
+  addTab,
   goBackward,
   goForward as goForwardAction,
   goToDiscoverPage,
@@ -47,6 +48,7 @@ import { selectIsInjectedProviderBlocked } from 'store/posthog/slice'
 import Logger from 'utils/Logger'
 import ErrorIcon from '../../../assets/icons/melting_face.png'
 import { useBrowserContext } from '../BrowserContext'
+import { isSameOriginSpaNavigation, isValidHttpUrl } from '../utils'
 import { WebView } from './Webview'
 
 export interface BrowserTabRef {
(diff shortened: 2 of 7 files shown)


packages/core-mobile/app/hooks/browser/useInjectedJavascript.ts
```diff
@@ -186,20 +187,30 @@ export function useInjectedJavascript(): InjectedJavascripts {
...
+          method: 'window_open',
           payload: url
-        };
-        window.ReactNativeWebView.postMessage(JSON.stringify(message));
-        return null;
-      }
-      return originalWindowOpen(url, target, features);
+        }));
+      } catch (e) {}
+      return null;
```
atn4z7, line 213: This now intercepts *all* `window.open` and always returns `null`, vs. before where non-`core://wc` URLs were delegated to the platform `window.open`. The real behavior change is navigation routing: `window.open(url)` used to load in the same WebView, now it opens a new tab — and `window.open(url, '_self')` self-navs become a new tab too. (Return-value/handle flows like `if (!w)` popup-blocked checks or `w.document.write` were likely already non-functional here since `setSupportMultipleWindows={false}`, so not a clean regression — just calling out the change.) Worth a quick QA on OAuth/onramp dApps; we could return a minimal stub window if any turn out to depend on a handle.

packages/core-mobile/app/hooks/browser/useInjectedJavascript.ts
```diff
@@ -186,20 +187,30 @@ export function useInjectedJavascript(): InjectedJavascripts {
...
-  const injectCustomWindowOpen = `(async function(){
-    const originalWindowOpen = window.open;
+  // Routes window.open through the native side instead of letting the WebView
+  // handle it. On Android with setSupportMultipleWindows=false, calling the
+  // platform window.open updates the WebView's visited history (and therefore
+  // our address bar) even when no actual page navigation replaces the content
+  // — that mismatch lets a malicious page render attacker UI under a trusted
+  // URL. We always intercept and hand the URL to the native tab manager so the
+  // address bar and the rendered content stay in sync.
+  const injectCustomWindowOpen = `(function(){
```
atn4z7, line 197: Separate point: this override is wired into `injectedJavaScript` (post-content-load), not `injectedJavaScriptBeforeContentLoaded`, so a page that calls `window.open` during its initial scripts can run before the override installs. The nav-state gating is the real defense so this is just defense-in-depth, but noting it's not airtight on its own.

packages/core-mobile/app/new/features/browser/components/BrowserTab.tsx
```diff
@@ -382,6 +403,27 @@ export const BrowserTab = forwardRef<BrowserTabRef, { tabId: string }>(
       ]
     )
 
+    // Surfaces a URL we trust as actually-rendered to the user-facing URL bar
+    // and to the EVM provider's origin tracker. Idempotent per URL.
+    const syncCommittedUrl = (url: string, title?: string): void => {
+      if (!url || url.startsWith('about:')) return
+      if (lastSyncedUrlRef.current === url) return
+
+      setCurrentUrl(url)
```
atn4z7: @B0Y3R-AVA flagging you since this touches the EVM injected provider. Moving the only `setCurrentUrl` call to commit (`onLoad`) means the provider origin (`currentUrlRef`) is now set *after* `domain_metadata` arrives instead of before, which I think regresses first-load auto-reconnect:

- priming runs via `handleDomainMetadata` → `primeAccounts`, and `domain_metadata` posts at `DOMContentLoaded` (`app/hooks/browser/evmProviderShim.ts:462`), which fires before `onLoad`
- `primeAccounts` early-returns on an empty origin (`app/hooks/browser/useEvmInjectedProvider.ts:596`), and `setCurrentUrl`'s branches both require a prior origin (`useEvmInjectedProvider.ts:172`/`:182`), so the first commit re-primes nothing
- the comment at `useEvmInjectedProvider.ts:184` actually states the precondition this breaks ("initial page load … is primed via domain_metadata"); pre-PR it held because provisional `onNavigationStateChange` set the origin before `DOMContentLoaded`
- the tests pass only because they call `setCurrentUrl` before `handleDomainMetadata` (`useEvmInjectedProvider.test.ts:1690-1696`); the "origin missing → no `accountsChanged`" case (`useEvmInjectedProvider.test.ts:1765`) is exactly the new first-load state

Net: previously-connected auto-reconnect dApps (wagmi `autoConnect`) likely load disconnected until a manual connect/account switch. MetaMask Mobile avoids this by priming from the commit handler with the URL passed in — `sendActiveAccount(siteInfo.url)` right after setting `resolvedUrlRef`: https://github.com/MetaMask/metamask-mobile/blob/b3e7eb887ee5ad923f076f3c74c2a84a5f33825d/app/components/Views/BrowserTab/BrowserTab.tsx#L724-L737 , and the "intentionally not initialized in onLoadStart" note https://github.com/MetaMask/metamask-mobile/blob/b3e7eb887ee5ad923f076f3c74c2a84a5f33825d/app/components/Views/BrowserTab/BrowserTab.tsx#L1094-L1097 . Suggested fix: drive priming from the commit path (or pass the committed URL into `primeAccounts`) + add a test where `handleDomainMetadata` runs before the origin is set.

packages/core-mobile/app/new/features/browser/components/BrowserTab.tsx
```diff
@@ -362,6 +364,24 @@ export const BrowserTab = forwardRef<BrowserTabRef, { tabId: string }>(
...
               break
+            case 'window_open': {
+              // The page called window.open(); open it in a new tab instead
+              // of letting the WebView handle it. Routing through a new tab
+              // keeps the originating page's URL and rendered content in sync
+              // (prevents address-bar spoofing on Android, where the platform
+              // can advance the visited URL without replacing the page).
+              const targetUrl = wrapper.payload
+              if (typeof targetUrl === 'string' && isValidHttpUrl(targetUrl)) {
+                dispatch(addTab())
```
atn4z7, line 375: Minor: a page looping `window.open` will spawn tabs up to the `limitMaxTabs` cap — bounded, but there's no rate-limit or telemetry on this path. Probably fine, just noting it.

atn4z7 (commented):
Thanks for the PR! The approach looks pretty solid to me. Left a few inline comments — the main one to flag is the EVM injected-provider first-load priming (tagged James). Nothing blocking the direction.

packages/core-mobile/app/new/features/browser/components/BrowserTab.tsx
B0Y3R-AVA: Confirmed — dug into the post-PR call flow and the first-load auto-reconnect regression is real. Good catch @atn4z7.

**Why it breaks (matches your analysis):**
- Pre-PR, `onNavigationStateChange` set `currentUrlRef` unconditionally on the provisional event — before `domain_metadata` (DOMContentLoaded).
- Post-PR, first-load cross-origin `setCurrentUrl` is deferred to `onLoad` (post-commit), which fires *after* `domain_metadata`.
- So `handleDomainMetadata → primeAccounts` no-ops on the empty origin (`useEvmInjectedProvider.ts:596`), and the later first `setCurrentUrl` also no-ops (no `prevOrigin`, so neither branch at `:171-189` runs). The shim's `_accounts` stays `[]` and no `accountsChanged` fires → wagmi `autoConnect` dApps load disconnected, with no automatic recovery (the already-granted `eth_requestAccounts` short-circuit at `router.ts:433-461` doesn't backfill `_accounts` either).

**Suggested fix (security-safe):** prime from the **trusted native committed URL**, not `domain_metadata` — the payload's `url`/`domain` are page-supplied, and using them for origin would reintroduce the exact spoof vector this PR closes (`getPeerMeta` keeps origin tied to the native URL for this reason). Have `syncCommittedUrl`'s provider call set `currentUrlRef` from the trusted URL **and then** call `primeAccounts` (e.g. a `handleCommittedUrl(url)` that does both). Reusing `primeAccounts` keeps CP-13772 (SPA re-prime) and CP-14382 (ungranted-active → `[]`) intact — same shape as MetaMask priming from the commit handler. Worth adding a test where `handleDomainMetadata` runs before the origin is set.

**Three more things worth handling in the same pass (same root cause — origin now set only at commit):**
1. **Early bridged RPC → `4100 Origin unavailable`.** Provider JS is injected before page scripts, but native-handled requests require `nativeOrigin` (`router.ts:739`). A dApp that fires an RPC before `onLoad` now gets rejected — worth confirming the committed-origin is set early enough (e.g. seed on navigation commit) rather than only at full `onLoad`.
2. **`window_open → addTab` new tabs** inherit the same missing-prime hole on their first load.
3. **Cross-origin cancellation moved to commit-only.** Pre-PR the provisional `setCurrentUrl` ran `cancelByOrigin` + unstuck parked connect approvals on cross-origin; a *hanging* cross-origin nav (the spoof scenario) no longer triggers that — security-adjacent, worth a look.

Happy to pair on the priming fix if useful.

chriselbring-avalabs: I think this is fixed to your spec. Also yeah, I think this could have been worse than just not priming correctly for a site that has been connected to, but could have ended up such that, on navigation from site A->B, currentUrlRef would still have A and then we would prime B's document with A's grant, exposing the address to B.

atn4z7 (dismissed):

onghwan (dismissed):

atn4z7 (approved):

onghwan (approved):

PR #3880, opened 2026-06-04
EVM state sync: non-EVM disconnect, SPA-nav re-prime, testnet-toggle resync

B0Y3R-AVA:
## Description

**Tickets: [CP-13671](https://ava-labs.atlassian.net/browse/CP-13671) · [CP-13772](https://ava-labs.atlassian.net/browse/CP-13772) · [CP-13775](https://ava-labs.atlassian.net/browse/CP-13775)**

Wraps up the EVM injected-provider work by closing the remaining state-sync gaps so the dApp stays in sync with wallet state across network changes, testnet toggles, and in-dApp (SPA) navigation. All three are in `useEvmInjectedProvider.ts`.

**Summary of changes**
- **CP-13671 — non-EVM network → `disconnect`.** The active-network sync effect emitted `chainChanged` unconditionally. It now emits an EIP-1193 `disconnect` (code 4901) when the wallet's active network is non-EVM (BTC/SVM/AVM/PVM), instead of a bogus EVM `chainChanged`.
- **CP-13772 — SPA-nav connection loss.** `domain_metadata` (and thus account priming) only fired on full page load, so a same-origin SPA route change (e.g. core.app → Stake) left the freshly-mounted provider consumer with no accounts. `setCurrentUrl` now re-primes `accountsChanged` on same-origin navigation to a new URL (shared `primeAccounts` helper; excludes the initial load, which is still primed via `domain_metadata`).
- **CP-13775 — testnet/developer-mode env mismatch.** The hook had no developer-mode awareness, so after toggling testnet it kept signing with the prior environment's chainId and the RPC env check rejected every tx with *"Invalid environment."* The hook now watches `selectIsDeveloperMode` and, on flip, drops the per-tab chain pin (`clearTabChainId`) so the sync effect re-points the tab to the active (correct-environment) network.

**Dependencies introduced**
None.

**Breaking changes**
None.

> Stacked on #3878 (active-account switch). Base will retarget to `main` as the EVM train lands. Review against the `boyer/injected-provider-account-switch` base.

## Screenshots/Videos

N/A — dApp-facing `accountsChanged`/`chainChanged`/`disconnect` events; no Core UI surface change.

## Testing

- **CP-13772:** open core.app in the in-app browser, confirm auto-connect, tap hamburger → Stake → still connected (no manual reconnect needed).
- **CP-13775:** connect on mainnet, toggle developer/testnet mode in settings, return to the dApp → transactions succeed (no "Invalid environment" toast).
- **CP-13671:** switch the wallet's active network to a non-EVM chain (Bitcoin/Solana) → dApp reflects a disconnected state (no bogus chainChanged).
- Trigger a Bitrise build and reference it here.
- Move CP-13671 / CP-13772 / CP-13775 into the "Testing" column on Jira.

## Checklist

[CP-13671]: https://ava-labs.atlassian.net/browse/CP-13671?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CP-13772]: https://ava-labs.atlassian.net/browse/CP-13772?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CP-13775]: https://ava-labs.atlassian.net/browse/CP-13775?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core-mobile/app/hooks/browser/useEvmInjectedProvider.test.ts
+++ b/packages/core-mobile/app/hooks/browser/useEvmInjectedProvider.test.ts
@@ -1877,5 +1877,76 @@ describe('useEvmInjectedProvider', () => {
         expect.stringContaining("'accountsChanged'")
       )
     })
+
+    // Build a store whose subscribe callback we can fire on demand, with a
+    // mutable `grants` so we can simulate a Connected Sites revoke (a fresh
+    // grants object, as Redux/Immer produces) with no active-account change.
+    const subscribableStore = (
+      initialGrants: Record<string, Record<string, string[]>>
+    ): {
+      store: ReturnType<typeof useStore>
+      revoke: (next: Record<string, Record<string, string[]>>) => void
+    } => {
+      let grants = initialGrants
+      let fire = (): void => undefined
+      const store = {
+        getState: () => ({ permissions: { grants } }),
+        dispatch: jest.fn(),
+        subscribe: jest.fn((cb: () => void) => {
+          fire = cb
+          return () => undefined
+        })
+      } as unknown as ReturnType<typeof useStore>
+      return {
+        store,
+        revoke: next => {
+          grants = next
+          fire()
+        }
+      }
+    }
+
+    it('emits accountsChanged([]) immediately when grants are revoked from Connected Sites (no account switch)', () => {
+      const { store, revoke } = subscribableStore({
+        'https://opensea.io': { [A]: ['EVM'] }
+      })
+      mockUseStore.mockReturnValue(store)
+      setupMocks({ account: accountA })
+      const { result } = renderProvider()
+      act(() => result.current.setCurrentUrl(ORIGIN))
+      // Prime so the dApp is seen connected to [A].
+      act(() =>
+        result.current.handleDomainMetadata(JSON.stringify({ name: 'OpenSea' }))
+      )
+      mockInjectJavaScript.mockClear()
+
+      // Revoke from Connected Sites — active account unchanged.
+      act(() => revoke({}))
+
+      expect(mockInjectJavaScript).toHaveBeenCalledWith(
+        expect.stringContaining("__coreProviderEmit('accountsChanged', [])")
+      )
+    })
+
+    it('does not emit on store changes that leave the origin grants unchanged', () => {
+      const grants = { 'https://opensea.io': { [A]: ['EVM'] } }
+      // Same grants object reference on the next store change (e.g. an unrelated
+      // slice updated) — must not produce a spurious accountsChanged.
+      const { store, revoke } = subscribableStore(grants)
+      mockUseStore.mockReturnValue(store)
+      setupMocks({ account: accountA })
+      const { result } = renderProvider()
+      act(() => result.current.setCurrentUrl(ORIGIN))
+      act(() =>
+        result.current.handleDomainMetadata(JSON.stringify({ name: 'OpenSea' }))
+      )
+      mockInjectJavaScript.mockClear()
+
+      act(() => revoke(grants))
+
+      expect(mockInjectJavaScript).not.toHaveBeenCalledWith(
+        expect.stringContaining("'accountsChanged'")
+      )
+    })
   })
 })
--- a/packages/core-mobile/app/hooks/browser/useEvmInjectedProvider.test.ts
+++ b/packages/core-mobile/app/hooks/browser/useEvmInjectedProvider.test.ts
@@ -1703,6 +1703,28 @@ describe('useEvmInjectedProvider', () => {
         )
       })
 
+      it('origin-gates the prime emit to the current page origin', () => {
+        // The prime injection must be wrapped in a window.location.origin check
+        // so a prime racing a cross-origin nav can't leak the previous origin's
+        // granted accounts into the next origin's page.
+        mockUseStore.mockReturnValue(withPermission(mockActiveAccount.addressC))
+        const { result } = renderProvider()
+        act(() => {
+          result.current.setCurrentUrl('https://opensea.io/')
+        })
+        mockInjectJavaScript.mockClear()
+
+        act(() => {
+          result.current.handleDomainMetadata(metadata)
+        })
+
+        expect(mockInjectJavaScript).toHaveBeenCalledWith(
+          expect.stringContaining(
+            'if(window.location.origin==="https://opensea.io")'
+          )
+        )
+      })
+
       it('injects accountsChanged([]) when no grant exists for the origin', () => {
         const { result } = renderProvider()
         act(() => {
@@ -1974,6 +1996,13 @@ describe('useEvmInjectedProvider', () => {
 
       expect(mockInjectJavaScript).toHaveBeenCalledWith(
         expect.stringContaining("__coreProviderEmit('accountsChanged', [])")
+      )
+      // ...and the emit is origin-gated so a revoke racing a cross-origin nav
+      // can't deliver into a different page (same guard as the switch/prime).
+      expect(mockInjectJavaScript).toHaveBeenCalledWith(
+        expect.stringContaining(
+          'if(window.location.origin==="https://opensea.io")'
+        )
       )
     })
 
--- a/packages/core-mobile/app/hooks/browser/useEvmInjectedProvider.test.ts
+++ b/packages/core-mobile/app/hooks/browser/useEvmInjectedProvider.test.ts
@@ -9,6 +9,7 @@ import {
   setActive
 } from 'store/network/slice'
 import { selectTabChainId, setTabChainId } from 'store/browser/slices/tabs'
+import { selectIsDeveloperMode } from 'store/settings/advanced'
 import { ModuleErrors, VmModuleErrors } from 'vmModule/errors'
 import {
   EIP1193_USER_REJECTED_CODE,
@@ -53,6 +54,10 @@ jest.mock('store/browser/slices/tabs', () => ({
   selectTabChainId: jest.fn(() => jest.fn(() => undefined)),
   setTabChainId: jest.fn((payload: { tabId: string; chainId: number }) => ({
     type: 'browser/tabs/setTabChainId',
+    payload
+  })),
+  clearTabChainId: jest.fn((payload: { tabId: string }) => ({
+    type: 'browser/tabs/clearTabChainId',
     payload
   }))
 }))
@@ -140,12 +145,14 @@ function setupMocks(
     account?: typeof mockActiveAccount | null
     network?: typeof mockActiveNetwork
     allNetworks?: Record<number, unknown>
+    developerMode?: boolean
   } = {}
 ): void {
   const account =
     overrides.account === undefined ? mockActiveAccount : overrides.account
   const network = overrides.network ?? mockActiveNetwork
   const allNetworks = overrides.allNetworks ?? mockAllNetworks
+  const developerMode = overrides.developerMode ?? false
 
   const mockTabChainIdSelector = jest.fn(() => undefined)
   ;(selectTabChainId as jest.Mock).mockReturnValue(mockTabChainIdSelector)
@@ -158,6 +165,7 @@ function setupMocks(
     (selector: (state: unknown) => unknown) => {
       if (selector === (selectAllNetworks as unknown)) return allNetworks
       if (selector === (selectActiveNetwork as unknown)) return network
+      if (selector === (selectIsDeveloperMode as unknown)) return developerMode
       if (selector === mockTabChainIdSelector) return undefined
       const selectorStr = selector.toString()
       if (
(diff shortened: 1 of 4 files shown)


packages/core-mobile/app/hooks/browser/useEvmInjectedProvider.ts
```diff
@@ -106,6 +108,16 @@ export function useEvmInjectedProvider(
...
+    // Non-EVM active network (BTC/SVM/AVM/PVM): the EVM provider can't serve it,
+    // so tell the dApp it's disconnected rather than emitting a bogus EVM
+    // chainChanged for a non-EVM chainId (EIP-1193 disconnect, code 4901).
+    // CP-13671.
+    if (activeNetwork.vmName !== NetworkVMType.EVM) {
+      webViewRef.current?.injectJavaScript(
+        `window.__coreProviderEmit && window.__coreProviderEmit('disconnect', { code: 4901, message: 'Disconnected from chain' }); true;`
+      )
+      return
+    }
```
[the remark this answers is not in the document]
B0Y3R-AVA: Good catch — fixed in ef8973763. Entering non-EVM mode now invalidates the cached EVM chain (`browserNetworkRef.current = { chainId: 0, rpcUrl: '' }`) before emitting disconnect. So returning to the SAME EVM chainId no longer hits the equality guard — chainChanged re-fires and a dApp that reacted to disconnect recovers.

packages/core-mobile/app/hooks/browser/useEvmInjectedProvider.test.ts
```diff
@@ -1348,6 +1356,48 @@ describe('useEvmInjectedProvider', () => {
...
+      rerender()
+
+      expect(mockInjectJavaScript).toHaveBeenCalledWith(
+        expect.stringContaining("__coreProviderEmit('disconnect'")
+      )
+      const chainChangedCalls = mockInjectJavaScript.mock.calls.filter(call =>
+        call[0].includes("__coreProviderEmit('chainChanged'")
+      )
+      expect(chainChangedCalls).toHaveLength(0)
+    })
```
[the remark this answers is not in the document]
B0Y3R-AVA: Added in ef8973763 — a recovery test: switch to a non-EVM network (disconnect) then back to the SAME EVM chain (43114) and assert chainChanged re-emits exactly once with 0xa86a.

packages/core-mobile/app/hooks/browser/useEvmInjectedProvider.ts
```diff
@@ -106,6 +108,21 @@ export function useEvmInjectedProvider(
...
+    // chainChanged for a non-EVM chainId (EIP-1193 disconnect, code 4901).
+    // CP-13671.
+    if (activeNetwork.vmName !== NetworkVMType.EVM) {
+      // Invalidate the cached EVM chain (sentinel 0 — no real EVM chainId) so
+      // returning to the SAME EVM chainId later still re-emits chainChanged.
+      // Otherwise the equality guard below would skip it and a dApp that reacted
+      // to this disconnect would never get a follow-up event to recover. CP-13671.
+      browserNetworkRef.current = { chainId: 0, rpcUrl: '' }
+      webViewRef.current?.injectJavaScript(
+        `window.__coreProviderEmit && window.__coreProviderEmit('disconnect', { code: 4901, message: 'Disconnected from chain' }); true;`
```
atn4z7, line 122: Behavioral confirm: if a user opens an EVM dApp while the wallet's active network is already non-EVM (Bitcoin/Solana), this fires on mount and emits disconnect(4901) right away — and the shim was seeded with the fallback chainId `0x1` (chainIdHex returns '0x1' for non-EVM), so the dApp briefly initializes at 0x1 and then immediately gets a disconnect. Is that the intended first-load experience for "open an EVM dApp while on a non-EVM chain"? Just want to make sure the load → instant-disconnect sequence is what we want rather than something cleaner.

packages/core-mobile/app/hooks/browser/useEvmInjectedProvider.ts
```diff
@@ -492,6 +567,33 @@ export function useEvmInjectedProvider(
...
+    const accounts = resolveActiveConnectedAccounts(granted, active.addressC)
+    lastEmittedAccountsRef.current = JSON.stringify(accounts)
+    webViewRef.current?.injectJavaScript(
+      `window.__coreProviderEmit && window.__coreProviderEmit('accountsChanged', ${JSON.stringify(
+        accounts
+      )}); true;`
+    )
+  }, [store, webViewRef])
+  // Publish to the ref so setCurrentUrl (defined above) can re-prime on SPA nav.
+  primeAccountsRef.current = primeAccounts
```
atn4z7: Nit: this assigns the ref during render, whereas `routerRef` is set in a `useEffect`. might be worth matching the `routerRef` pattern for consistency.

packages/core-mobile/app/hooks/browser/useEvmInjectedProvider.test.ts
```diff
@@ -1376,6 +1384,79 @@ describe('useEvmInjectedProvider', () => {
       )
       expect(chainChangedCalls).toHaveLength(0)
     })
+
+    it('emits disconnect (not chainChanged) when the active network is non-EVM (CP-13671)', () => {
+      setupMocks({ network: mockActiveNetwork })
```
atn4z7, line 1418: Small coverage gap: all the disconnect tests use an un-pinned tab (tabChainId === undefined). The "a pinned tab is insulated from active-network changes" invariant — i.e. a tab that's called `wallet_switchEthereumChain` should NOT disconnect when the wallet flips to a non-EVM chain — isn't covered. A quick test (set tabChainId, flip active to non-EVM, assert no disconnect/chainChanged) would lock that in, since it's the behavior that makes the whole per-tab model hold.

packages/core-mobile/app/hooks/browser/useEvmInjectedProvider.ts
```diff
@@ -117,10 +134,26 @@ export function useEvmInjectedProvider(
...
+  // follows the wallet's active network for the NEW environment (the sync effect
+  // above then re-points browserNetworkRef + emits chainChanged). Otherwise the
+  // tab keeps signing with the prior environment's chainId and the RPC env check
+  // rejects every tx with "Invalid environment". CP-13775.
+  const prevDevModeRef = useRef(isDeveloperMode)
+  useEffect(() => {
+    if (prevDevModeRef.current === isDeveloperMode) return
+    prevDevModeRef.current = isDeveloperMode
+    dispatch(clearTabChainId({ tabId }))
+  }, [isDeveloperMode, dispatch, tabId])
```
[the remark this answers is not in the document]
B0Y3R-AVA: Fixed — the dev-mode effect now only dispatches `clearTabChainId` when the tab actually has a pin (`tabChainId !== undefined`), so an un-pinned tab no longer takes a redundant store update + rerender on every toggle. Added a test covering the no-pin case (and the existing clear test now sets a pin so it exercises the guarded path).

packages/core-mobile/app/hooks/browser/useEvmInjectedProvider.ts
B0Y3R-AVA: Confirming it's intended, with the reasoning: the shim is a synchronously-injected IIFE and per EIP-1193 must initialize with *some* `chainId`, but it can't represent a non-EVM chain as an EVM chainId — so `chainIdHex` seeds the `0x1` fallback. On mount the CP-13671 effect immediately emits `disconnect(4901)`, which is the canonical EIP-1193 signal that the provider can't serve the current chain; a compliant dApp disables EVM interactions on it. The brief `0x1` is harmless — no EVM request can succeed while the active chain is non-EVM (requests route to the active network), and returning to an EVM chain re-emits `chainChanged` (the tested recovery path). A cleaner variant would be to inject a "start disconnected" flag into the shim when the active network is non-EVM at injection time, avoiding the momentary `0x1` — happy to do that as a small follow-up if we want to polish this entry path, but it's not a correctness issue.

packages/core-mobile/app/hooks/browser/useEvmInjectedProvider.ts
B0Y3R-AVA: Good catch — matched the `routerRef` pattern. `primeAccountsRef.current = primeAccounts` is now set in a `useLayoutEffect` instead of during render. It's read by `setCurrentUrl` on nav events, so (like `routerRef`) it wants to be populated synchronously on commit, and this drops the render-phase mutation.

packages/core-mobile/app/hooks/browser/useEvmInjectedProvider.test.ts
B0Y3R-AVA: Added — `does NOT disconnect a pinned tab when the active network flips to non-EVM`: it pins `tabChainId`, flips the active network to Bitcoin, and asserts neither `disconnect` nor `chainChanged` is emitted. Extended the test harness' `setupMocks` with a `tabChainId` override to make pinned-tab setup reusable. Locks in the per-tab insulation invariant.

packages/core-mobile/app/hooks/browser/useEvmInjectedProvider.ts
atn4z7: did you mean `useEffect`? `useLayoutEffect` triggers synchronously before the screen is painted.

B0Y3R-AVA: I did mean `useLayoutEffect` — `routerRef` is actually on `useLayoutEffect` now, not `useEffect`. It got moved in #3867 (merged) in response to a Copilot null-window-race comment: `setCurrentUrl` fires from the WebView's nav events and reads these refs, and a render-phase / `useEffect` assignment left a (theoretical) window where the ref could still be null when an early navigation ran. `useLayoutEffect` populates it synchronously on commit, before paint. So `primeAccountsRef` is matching the current `routerRef` pattern.

Fair to call it belt-and-suspenders here: for a WebView, `onNavigationStateChange` arrives via the bridge after mount + paint, so `useEffect` would set the ref in time too — the difference is only the theoretical early-event window. I kept it on `useLayoutEffect` purely to stay consistent with `routerRef`. If you'd rather both use `useEffect` (more idiomatic for a value-publish ref), I'm happy to switch them together — I'd just keep the two consistent with each other either way. Preference?

atn4z7: thanks for the details. I think it's cool then to use `useLayoutEffect`.

atn4z7 (dismissed):

packages/core-mobile/app/hooks/browser/useEvmInjectedProvider.ts
```diff
@@ -252,6 +296,85 @@ export function useEvmInjectedProvider(
...
+        activeAccountRef.current?.addressC
+      )
+
+      // Shared dedupe ref so this never double-fires with the switch/prime paths.
+      const serialized = JSON.stringify(accounts)
+      if (lastEmittedAccountsRef.current === serialized) return
+      lastEmittedAccountsRef.current = serialized
+      webViewRef.current?.injectJavaScript(
+        `window.__coreProviderEmit && window.__coreProviderEmit('accountsChanged', ${serialized}); true;`
+      )
```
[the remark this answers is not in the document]
B0Y3R-AVA: Fixed in 38e5cd9f3. The revoke-subscription emit now goes through a shared `injectAccountsChanged(origin, serialized)` helper that wraps the inject in the `window.location.origin===<origin>` guard (the same one the active-account-switch effect uses). Added a test asserting the revoke emit is origin-gated.

packages/core-mobile/app/hooks/browser/useEvmInjectedProvider.ts
```diff
@@ -452,6 +575,38 @@ export function useEvmInjectedProvider(
...
+      domain: origin,
+      vmType: NetworkVMType.EVM
+    })(store.getState())
+    const accounts = resolveActiveConnectedAccounts(granted, active.addressC)
+    lastEmittedAccountsRef.current = JSON.stringify(accounts)
+    webViewRef.current?.injectJavaScript(
+      `window.__coreProviderEmit && window.__coreProviderEmit('accountsChanged', ${JSON.stringify(
+        accounts
+      )}); true;`
+    )
```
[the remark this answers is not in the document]
B0Y3R-AVA: Fixed in 38e5cd9f3. `primeAccounts` now serializes once and emits through the shared `injectAccountsChanged` helper, so it's origin-gated like the switch path. Added a test asserting the prime emit is wrapped in the origin check.

packages/core-mobile/app/hooks/browser/useEvmInjectedProvider.ts
```diff
@@ -105,6 +108,21 @@ export function useEvmInjectedProvider(
...
+    // CP-13671.
+    if (activeNetwork.vmName !== NetworkVMType.EVM) {
+      // Invalidate the cached EVM chain (sentinel 0 — no real EVM chainId) so
+      // returning to the SAME EVM chainId later still re-emits chainChanged.
+      // Otherwise the equality guard below would skip it and a dApp that reacted
+      // to this disconnect would never get a follow-up event to recover. CP-13671.
+      browserNetworkRef.current = { chainId: 0, rpcUrl: '' }
+      webViewRef.current?.injectJavaScript(
+        `window.__coreProviderEmit && window.__coreProviderEmit('disconnect', { code: 4901, message: 'Disconnected from chain' }); true;`
+      )
```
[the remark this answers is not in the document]
B0Y3R-AVA: Fixed in 38e5cd9f3. The shim now tracks a `_connected` flag: `__coreProviderEmit` sets it false on `disconnect` and true on `connect`/`chainChanged` (the CP-13671 disconnect→chainChanged recovery), `isConnected()` returns the flag instead of a hard-coded true, and the legacy `provider._isConnected` property is kept in sync for dApps that read it directly. Added shim tests for the flag wiring.

packages/core-mobile/app/hooks/browser/useEvmInjectedProvider.ts
```diff
@@ -298,6 +341,40 @@ export function useEvmInjectedProvider(
...
+        activeAccountRef.current?.addressC
+      )
+
+      // Shared dedupe ref so this never double-fires with the switch/prime paths.
+      const serialized = JSON.stringify(accounts)
+      if (lastEmittedAccountsRef.current === serialized) return
+      lastEmittedAccountsRef.current = serialized
+      webViewRef.current?.injectJavaScript(
+        `window.__coreProviderEmit && window.__coreProviderEmit('accountsChanged', ${serialized}); true;`
+      )
```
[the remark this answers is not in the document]
B0Y3R-AVA: Addressed in 38e5cd9f3 — see the reply on the current revision: the revoke emit is now origin-gated via the shared `injectAccountsChanged` helper.

atn4z7 (approved):

onghwan (approved):

PR #3883, opened 2026-06-05
CP-14411: fix seedless add-account addresses and Ledger Solana skip prompt

ruijia1in:
iOS: 8775
Android: 8776

## Description

**Ticket: [CP-14411]**

Please provide:

- Bump @avalabs/crypto-sdk → 1.0.2 and @avalabs/crypto-nitro → 0.3.2. These releases normalize 65-byte uncompressed secp256k1 public keys to compressed form inside the native batch derivation (require_compressed_secp256k1 / compress_pubkey). That uncompressed-key rejection was the root cause: the batched deriveAllAddresses (introduced in #3851) rejected seedless's uncompressed Cubist pubkeys while mnemonic's compressed keys worked. With the native fix, seedless uses the same batched deriveAllAddresses path as every other wallet type — no app-side routing workaround.
- Fix missing addresses when adding a new seedless account. SeedlessWallet.getPublicKeyFor now reads keys from SeedlessPubKeysStorage on every lookup instead of a constructor-time #addressPublicKeys snapshot. addAccount refreshes that storage (refreshPublicKeys → save), so a SeedlessWallet cached in WalletDerivedDataCache resolves the freshly persisted keys on the next getAddresses. Removed the now-dead #addressPublicKeys field/constructor param and the WalletFactory.cache.clearWallet workaround.

Ledger Solana connection (AppConnectionScreen)
- Remove the eager auto-open effect that fired openApp(SOLANA) the moment the user landed on the Solana step. Sending that open-app APDU is what makes the Ledger display "Open Solana?", so the device prompted even when the user went on to tap "Skip Solana". The prompt is now deferred until the user taps Continue (handleConnectSolana -> getSolanaKeysForRange -> ensureAppReady -> openApp), so skipping never contacts the device for Solana.

## Screenshots/Videos

## Testing

- login to seedless works
- add seedless account works
- login with mnemonic works
- add mnemonic account works and no performance degrade after adding account

## Checklist

[CP-14411]: https://ava-labs.atlassian.net/browse/CP-14411?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/core-mobile/app/services/account/AccountsService.tsx
```diff
@@ -422,6 +429,18 @@ class AccountsService {
       isTestnet
     } as Network
 
+    // Seedless intentionally uses the singular per-index derivation path with
+    // its original lenient semantics (a failed module yields empty-string
+    // addresses for that chain rather than failing the whole account).
+    if (walletType === WalletType.SEEDLESS) {
```
atn4z7: What's actually failing when seedless goes through `deriveAllAddresses`? Trying to understand which module rejects and why before we route around it.

One thing I noticed: the singular path also swaps the underlying module method (`deriveAddress` vs the plural `deriveAddresses`), so this isn't just lenient-vs-fail-loud. Any chance the plural one is what's broken for seedless and we're working around it here rather than fixing it at the source? If you've got the error/stack from the failing case that'd help.

packages/core-mobile/app/services/account/AccountsService.tsx
```diff
@@ -369,6 +369,13 @@ class AccountsService {
 
         // prompt Core Seedless API to derive new keys
         await wallet.addAccount(index)
+
+        // addAccount refreshes the pubkeys in SeedlessPubKeysStorage, but any
+        // SeedlessWallet instance cached in WalletDerivedDataCache still holds
+        // the pre-add `#addressPublicKeys` snapshot. Clear the cache so the
+        // getAddresses lookup below rebuilds the wallet with the new keys —
+        // otherwise the new account's chains resolve to empty addresses.
+        WalletFactory.cache.clearWallet(walletId)
```
atn4z7: The cache clear works, but I think it's patching the symptom rather than the cause, and it does it from the wrong layer. `createNextAccount` now has to know that `WalletFactory` keeps an internal cache, that `SeedlessWallet` snapshots its pubkeys at construction, and that those two facts combine to hand back stale addresses. That's a lot of cross-layer knowledge for an orchestration method to carry — `AccountsService` reaching into `WalletFactory.cache` to fix up a seedless-internals problem is the part that feels off to me.

It also only covers the `addAccount` path. `deriveMissingKeys()` goes through the same `refreshPublicKeys → save` flow and has the identical staleness problem, but it doesn't clear anything, so I think that path is still broken after this.

What if `SeedlessWallet` just didn't keep a snapshot? `getPublicKeyFor` is the only thing that reads `#addressPublicKeys`, and it could read from storage directly:
```
public async getPublicKeyFor({ derivationPath, curve }) {
  if (derivationPath === undefined) throw new Error(...)
  const pubKeys = await SeedlessPubKeysStorage.retrieve()
  const publicKey = pubKeys.find(findPublicKey(derivationPath, curve))
  if (!publicKey) throw new Error(...)
  return publicKey.key
}
```
`retrieve()` is already memoized — it caches the array in memory and `save()` updates that cache — so this isn't a keychain hit on every lookup. With no snapshot, a cached wallet instance can't go stale, so there's nothing to invalidate: `addAccount` and `deriveMissingKeys` both just work, the `clearWallet` call goes away, and `AccountsService` no longer needs to know `WalletFactory` has a cache at all.

atn4z7 (commented):
Approach looks right to me, and this is way cleaner than the first pass. A few things before it goes in:

**The `clearWallet` call can come out now.** Since `getPublicKeyFor` reads from storage instead of the snapshot, a cached `SeedlessWallet` instance can't go stale anymore — `save()` updates the storage cache during `addAccount`, so even an old cached instance resolves the new keys on the next lookup. The other cache layers don't go stale for a new account either (`publicKeyCache` is keyed by derivation path, `xpubCache` by account index, both brand new → miss → fresh derive). So nothing needs the clear. The comment above it also still talks about the instance holding "the pre-add `#addressPublicKeys` snapshot," which isn't true anymore since that field isn't read. Dropping the line also gets `AccountsService` out of the business of knowing `WalletFactory` has a cache, which is nice.

**`#addressPublicKeys` is dead now.** It's still set in the constructor but nothing reads it after this change. Might as well drop the field and the constructor param (and the arg in `WalletFactory`) so we're not leaving a write-only field lying around.

**Tests are the thing I'd want sorted before merge.** Couple of issues:
- The new behavior (`getPublicKeyFor` reading from storage) isn't covered, and the unit tests from the first version got dropped without anything replacing them.
- The existing `SeedlessWallet` tests construct the wallet with explicit keys and mock `save()` but not `retrieve()` — and there's no global mock for `SeedlessPubKeysStorage`. The `addAccount` tests go through `getMnemonicId` → `getAddressPublicKey` → `getPublicKeyFor`, which now hits the real `retrieve()` → returns `[]` → throws "Public key not found" before the "Cannot retrieve the mnemonic id" they're asserting on. So those look like they'd fail (or pass for the wrong reason). Can you confirm the jest run is green? If it is I'd like to understand why, because on paper they're reading unmocked storage now. Either way it'd be good to mock `retrieve()` and add a test that `getPublicKeyFor` returns the latest keys from storage.

One minor thing: `getPublicKeyFor` now does an `await retrieve()` on every key lookup, so on every derive/sign. It's memoized after the first load so it's basically free, just flagging it's on the signing path now.

atn4z7 (approved):

onghwan (approved):
dev-tested

PR #3895, opened 2026-06-10
Device Farm: fire-and-forget scheduling, fail fast on empty device pool

B0Y3R-AVA:
## Description

**Ticket: [CP-]**

Fixes two problems in the nightly AWS Device Farm regression (`aws-full-regression-runs`) that cost ~2h of billed idle Bitrise machine time per platform leg per night, and could silently report green with zero tests executed:

- **Polling burn:** the Bitrise step polls `GetRun` for up to `WAIT_FOR_COMPLETION_TIMEOUT_SEC` (2h) while the device-side suite runs (full regression is ~2h per platform, 100+ cases). The machine is billed the entire wait, and when the suite outlasts the cap the build reds out as a timeout rather than reporting actual results. This polling is the largest single waste category in our Bitrise usage (~thousands of build-min/month).
- **Silent SKIPPED green:** if the device pool matches 0 devices (e.g. a temporary pool change — this happened on 2026-06-09 and the iOS leg went green having executed nothing), the run completes instantly as `SKIPPED`, which was whitelisted as success. Nothing ran, build showed green, no one was alerted.

Changes:
1. `WAIT_FOR_COMPLETION` now defaults to **false** in both wrapper scripts — the step uploads, schedules the run, and exits. Test results still reach TestRail via the reporter bundled in the device-side test package. Opt back into polling per-run with `WAIT_FOR_COMPLETION=true`.
2. **Pre-flight device-pool check fails fast** when the pool matches 0 compatible devices (would otherwise become a silent `SKIPPED` run — see 2026-06-09 iOS). Also fixes the existing broken inspection call: `ListDevicePoolDevicesCommand` doesn't exist in the AWS SDK (the `is not a constructor` warning in every run log); replaced with `GetDevicePoolCompatibleDevicesCommand`. Inspection stays warn-only on API errors (e.g. missing IAM read perms) so it can't break runs spuriously.
3. **`SKIPPED` removed from success results** for anyone who opts into waiting, with a descriptive error.

Note: with fire-and-forget, the Bitrise build status reflects "scheduled successfully," not AWS pass/fail — test outcomes live in TestRail / Device Farm console (as they already do; the Bitrise red was duplicating that signal at ~2h/leg/night of machine cost).

No dependencies introduced (`@aws-sdk/client-device-farm` already used).

## Screenshots/Videos

N/A — CI scripts only.

## Testing

- [ ] Trigger `aws-full-regression-runs` on this branch; expect: pool inspection lists compatible devices for both legs, runs scheduled, steps exit in seconds instead of polling ~2h
- [ ] Confirm results from the runs appear in TestRail after the AWS runs finish
- [ ] Optional negative test: point `DEVICEFARM_DEVICE_POOL_ARN` at a pool matching no devices and confirm the step fails fast with the 0-compatible-devices error instead of scheduling a SKIPPED run
- [ ] Bitrise build numbers: _add after triggering_

- Heads-up for QA: Bitrise no longer waits on AWS results — nightly pass/fail signal moves entirely to TestRail / Device Farm console. If a Bitrise-visible red/green gate per nightly is still wanted, we can add a cheap scheduled results-check job instead of inline polling.

## Checklist

### DIFF
--- a/packages/core-mobile/scripts/bitrise/devicefarm/android/androidDeviceFarmRegression.sh
+++ b/packages/core-mobile/scripts/bitrise/devicefarm/android/androidDeviceFarmRegression.sh
@@ -32,7 +32,10 @@ export DEVICEFARM_APP_PATH="$BITRISE_APK_PATH"
 export DEVICEFARM_TEST_PACKAGE_PATH="$TEST_PACKAGE"
 export DEVICEFARM_TEST_SPEC_PATH="$TEST_SPEC"
 export PLATFORM="android"
-export WAIT_FOR_COMPLETION="${WAIT_FOR_COMPLETION:-true}"
+# Fire-and-forget by default: results reach TestRail from the device-side reporter,
+# so polling GetRun only kept the Bitrise machine billing (~2h/night) and timed out
+# whenever the suite outlasted WAIT_FOR_COMPLETION_TIMEOUT_SEC.
+export WAIT_FOR_COMPLETION="${WAIT_FOR_COMPLETION:-false}"
 
 # 4. Extract only needed vars from env file
 if [ -f "$CORE_MOBILE_DIR/.env.production.e2e" ]; then
--- a/packages/core-mobile/scripts/bitrise/devicefarm/ios/iosDeviceFarmRegression.sh
+++ b/packages/core-mobile/scripts/bitrise/devicefarm/ios/iosDeviceFarmRegression.sh
@@ -34,7 +34,9 @@ export DEVICEFARM_APP_PATH="$BITRISE_IPA_PATH"
 export DEVICEFARM_TEST_PACKAGE_PATH="$TEST_PACKAGE"
 export DEVICEFARM_TEST_SPEC_PATH="$TEST_SPEC"
 export PLATFORM="ios"
-export WAIT_FOR_COMPLETION="${WAIT_FOR_COMPLETION:-true}"
+# Fire-and-forget by default: results reach TestRail from the device-side reporter,
+# so polling GetRun only kept the Bitrise machine billing while it waited.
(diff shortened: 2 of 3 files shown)


B0Y3R-AVA:
Reopening with the cost case spelled out, since "the build takes 2–2.5h anyway" is the natural objection:

**The 2–2.5h *is* the waste.** The nightly Bitrise build = ~35–40 min of real work (build + sign + upload) followed by **~1.5–2h of `GetRun` polling once a minute** while the tests run on AWS hardware (which costs us $0 — public fleet, free usage). This PR ends the Bitrise build right after the upload; AWS runs the exact same tests on the same schedule and reports to TestRail via the bundled reporter exactly as today. Nothing about test coverage, timing, or results changes — we just stop paying a Bitrise VM to watch a progress ticker.

Measured from 90 days of build data: ~**2,500 billed minutes/month (~13% of total account usage)** is this polling. It is the single largest zero-value minute category in the account.

What we give up: the Bitrise build tile as the nightly's pass/fail signal. Today that tile is unreliable in both directions — it reds out on the 2h timeout even when the suite passes, and it once showed green when a pool misconfiguration meant **zero tests executed** (`COMPLETED/SKIPPED`, 2026-06-09). This PR also fixes that: empty pool now fails fast in seconds, and `SKIPPED` is no longer treated as success for anyone who opts into waiting (`WAIT_FOR_COMPLETION=true` remains available per-run).

If a Bitrise-visible red/green for the nightly is still wanted, the follow-up is a tiny scheduled results-check workflow (~2 min on `linux.medium`, one `GetRun` call, 3h after the nightly) — dashboard signal for ~1% of the previous cost. Happy to add it if QA wants it.

PR #3896, opened 2026-06-10
CP-14426: fix swap token-picker gap/duplicate + iOS overscroll header-vanish in ListScreenV2

bogdandobritoiu:
## Description

**Ticket: [CP-14426]**

Two fixes affecting `ListScreenV2`-based token pickers.

### 1. Swap "Select a token" (To) — empty gap / duplicate token row

- **Summary:** removed a large empty gap that pushed rows to the bottom of the picker (and the duplicated selected-token row it turned into).
- **Context:** the pinned-token dedup matched on `getTokenKey` (`internalId ?? localId`-`chainId`), but `keyExtractor` used a different string (`token-${localId}-${chainId}`). A token selected from another source carries a different `internalId` than its API-list twin, so the dedup missed it and FlashList received **duplicate keys** — which, with FlashList v2's default `maintainVisibleContentPosition`, produced a persistent gap.
- **Fix (`SelectToSwapTokenScreen.tsx`):** `keyExtractor` now uses `getTokenKey(item)`; the selected token is pinned from the **list's own** matching object (matched by `localId` + `networkChainId`) instead of injecting the external object, so there's structurally one instance.

### 2. iOS overscroll completely hides the `ListScreenV2` sticky header

- **Summary:** on iOS, overscrolling (pull-down) at the top of any `ListScreenV2` modal left an empty void where the header was; now the header stays and rubber-bands with the content.
- **Context:** FlashList v2's `StickyHeaders` only renders the overlay while `getLayout(stickyIndex).y (=0) <= scrollOffset + stickyHeaderOffset`. On an iOS top overscroll `scrollOffset` goes negative → the binary search returns `-1` → the overlay returns `null`, and because `ListScreenV2` uses `stickyHeaderConfig={{ hideRelatedCell: true }}` the in-flow cell is hidden too → nothing renders.
- **Fix:** `patches/@shopify+flash-list+2.3.0.patch` (applied via `postinstall`): (1) clamp the sticky-index lookup to `Math.max(0, getLastScrollOffset())` so index 0 stays sticky during overscroll; (2) add an `overscrollTranslateY` so the header rubber-bands down 1:1 with the content. No-op on Android.

### Dependencies

- No new packages. Adds one `patch-package` patch for `@shopify/flash-list@2.3.0` (documented in `docs/patches.md`). ⚠️ Must be re-checked on any `@shopify/flash-list` version bump (edits target `dist/`). Likely an upstream FlashList bug — a sticky header at index 0 shouldn't vanish when overscrolling above it.

## Screenshots/Videos

**Before**

https://github.com/user-attachments/assets/a96ff4c8-7913-4807-bfbf-e46454741cec

**After**

https://github.com/user-attachments/assets/250dcc47-7240-421d-a77a-8fcc4324773d

## Testing

iOS - 8813
Android - 8814

**Dev Testing**

- Swap → tap the "to" token:
  - Happy path: the previously-selected token is pinned at the top with a single checkmark, rows are flush under the header, no empty gap.
  - Edge: switch networks / type in search — pinning + grouping (verified / "Unverified tokens" divider) still behave; no duplicate rows.
- Any `ListScreenV2` modal on iOS (e.g. token pickers, Wallets):
  - Overscroll (pull down) at the top — header stays visible and moves down with the content, then snaps back. No void, no doubled header during normal scroll.

## Checklist

[CP-14426]: https://ava-labs.atlassian.net/browse/CP-14426?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ruijia1in (approved):
👍 🎖️

atn4z7 (approved):
dev tested

PR #3898, opened 2026-06-10
CP-14428: fix CircularDial value drifting after a fast swipe-release

bogdandobritoiu:
## Description

**Ticket: [CP-14428]**

Fast-swiping the `CircularDial` and lifting the finger left the value (and the arc) continuing to change on its own.

**Root cause:** during a drag the dial emits `onChange` on every step crossing (up to once per frame). A fast swipe produces these faster than the parent can render, building a backlog of stale values that drains on the JS thread *after* the finger lifts. Because the readout number was rendered from the parent's controlled `value` prop, it kept ticking through that backlog; meanwhile stale echoes were re-synced back into the shared progress value, jerking the arc.

**Changes:**
- **Throttle the mid-drag `onChange`** to ~80ms via a leading-edge UI-thread time gate (reset on gesture start). `onEnd` always emits the final snapped value, so no crossing is lost. This keeps the parent — and the live fiat caption — from flooding and removes the backlog.
- **Settling guard** — block prop→progress sync for a short window after release (token-guarded so rapid repeated swipes each get a clean window) so an in-flight echo can't re-sync into the arc.
- **Decouple the readout number from the parent round-trip** — drive the displayed number from the shared progress value on the UI thread via `useAnimatedProps` (uncontrolled while not editing, controlled draft while editing), mirroring the sibling `LeverageDisplay`. The number now lands on the final value instantly on release instead of lagging behind it or drifting.
- Added a `formatNatural` worklet (UI-thread-safe, regex-free) mirroring the readout's `naturalDigits`, with parity unit tests, plus tests for the settling guard.

No new dependencies. Not breaking — `CircularDial`'s public API is unchanged.

**Before**

https://github.com/user-attachments/assets/5c1f8425-62e9-4271-9104-a8489694fdd4

**After** 

https://github.com/user-attachments/assets/d2159d2c-f3b8-4ed9-bba8-033a2f74c4ff

## Testing

iOS - 8815
Android - [8816](https://app.bitrise.io/app/7d7ca5af7066e290/installable-artifacts/74406b12e51ae1b6/public-install-page/a3be3653998cd0812a8690785cc388f4)

**Dev Testing**

Reproduce on the Fast Stake amount screen (or any `CircularDial` consumer):

- Swipe the dial hard left/right repeatedly (stress test), then lift.
  - Expected: the number lands on the final value immediately — no self-ticking or catch-up jump after lift.
  - Expected: the arc/knob does not jump around after release.
- Drag slowly — the fiat caption still updates live (now ~12/s).
- Manual input: tap the value, type, blur to commit — display↔edit handoff works and typed precision is preserved.
- Presets (25% / 50% / Max) animate and commit without post-animation drift.

**QA Testing**

- Acceptance: after any swipe (including a very fast one), releasing the dial leaves the value static at the released amount; no value changes occur without user input.

## Checklist

[CP-14428]: https://ava-labs.atlassian.net/browse/CP-14428?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/k2-alpine/src/components/CircularDial/CircularDial.tsx
+++ b/packages/k2-alpine/src/components/CircularDial/CircularDial.tsx
@@ -1,4 +1,4 @@
-import React, { FC, useCallback, useMemo, useRef } from 'react'
+import React, { FC, useCallback, useEffect, useMemo, useRef } from 'react'
 import type { LayoutChangeEvent } from 'react-native'
 import { Gesture, GestureDetector } from 'react-native-gesture-handler'
 import {
@@ -236,6 +236,16 @@ export const CircularDial: FC<CircularDialProps> = ({
       }, SETTLE_WINDOW_MS)
     },
     [isSettling, settleToken]
+  )
+
+  // Cancel a pending settle timer on unmount so it can't fire ~250ms later
+  // and write to shared values after the component is gone (mirrors the rAF
+  // cleanup in DialReadout).
+  useEffect(
+    () => () => {
+      if (settleTimerRef.current !== null) clearTimeout(settleTimerRef.current)
+    },
+    []
   )
 
   // manualActivation + direction-based commit: once total motion
--- a/packages/k2-alpine/src/components/CircularDial/CircularDial.tsx
+++ b/packages/k2-alpine/src/components/CircularDial/CircularDial.tsx
@@ -219,13 +219,19 @@ export const CircularDial: FC<CircularDialProps> = ({
 
   const readoutRef = useRef<DialReadoutHandle>(null)
 
-  // Clears the settling guard once the post-release window elapses, but
-  // only if no newer interaction has opened its own window since (token
-  // check) — otherwise a stale timer from an earlier swipe could reopen
-  // prop sync mid-settle and let an echo through.
+  // Pending settle-window timer, so rapid repeated swipes cancel the prior
+  // one instead of piling up dozens of timeouts that all wake to no-op.
+  const settleTimerRef = useRef<ReturnType<typeof setTimeout> | null>(null)
+
+  // Clears the settling guard once the post-release window elapses. Cancels
+  // any in-flight window first; the token check is a belt-and-suspenders
+  // against a stale timer reopening prop sync mid-settle and letting an
+  // echo through.
   const clearSettlingSoon = useCallback(
     (token: number) => {
-      setTimeout(() => {
+      if (settleTimerRef.current !== null) clearTimeout(settleTimerRef.current)
+      settleTimerRef.current = setTimeout(() => {
+        settleTimerRef.current = null
         if (settleToken.value === token) isSettling.value = false
       }, SETTLE_WINDOW_MS)
     },
@@ -341,11 +347,13 @@ export const CircularDial: FC<CircularDialProps> = ({
     })
 
   // Emit onChange on step crossings while the dial is active so consumers —
-  // and the readout text/caption, driven by the parent's controlled
-  // `value` — track the value live. Throttled on the UI thread to
-  // EMIT_THROTTLE_MS via a leading-edge time gate (no trailing timer, so
-  // nothing is queued to fire after release); onEnd always emits the final
-  // snapped value, so throttled-away crossings are never lost.
+  // and parent-derived UI such as the fiat caption — track the value live.
+  // (The readout number itself is now driven by progressSv on the UI thread
+  // in DialReadout, so it no longer depends on this round-trip.) Throttled
+  // on the UI thread to EMIT_THROTTLE_MS via a leading-edge time gate (no
+  // trailing timer, so nothing is queued to fire after release); onEnd
+  // always emits the final snapped value, so throttled-away crossings are
+  // never lost.
   useAnimatedReaction(
     () => {
       const raw = progressSv.value * vMax
--- a/packages/k2-alpine/src/components/CircularDial/CircularDial.tsx
+++ b/packages/k2-alpine/src/components/CircularDial/CircularDial.tsx
@@ -55,6 +55,18 @@ const ACTIVATION_TAN = Math.tan((ACTIVATION_HALF_ANGLE_DEG * Math.PI) / 180)
 // visible knob (KNOB_RADIUS=11) for a comfortable touch target.
 const KNOB_HIT_RADIUS = 32
 
+// Cap on how often the dial emits onChange to the parent during a drag.
+// A fast swipe crosses a step every frame; emitting each one floods the
+// JS thread faster than the parent can render and builds a backlog that
+// drains *after* the finger lifts — making the value keep changing on its
+// own. ~80ms (≈12/s) is live enough for a readout/caption yet slow enough
+// that the parent keeps up and nothing is left queued to fire post-lift.
+const EMIT_THROTTLE_MS = 80
+
+// How long after release to keep blocking prop→progressSv sync, covering
+// the last in-flight throttled onChange echo as it drains on the JS thread.
+const SETTLE_WINDOW_MS = 250
+
 export const CircularDial: FC<CircularDialProps> = ({
   value,
   onChange,
@@ -144,12 +156,20 @@ export const CircularDial: FC<CircularDialProps> = ({
     [presets]
   )
 
-  const { progressSv, isActive } = useDialValue({
+  const { progressSv, isActive, isSettling } = useDialValue({
     value,
     max: vMax,
     step: vStep
   })
   const isDragging = useSharedValue(false)
+
+  // UI-thread timestamp of the last throttled onChange emission; gates the
+  // step-crossing reaction to EMIT_THROTTLE_MS. Reset on each gesture start.
+  const lastEmitMs = useSharedValue(0)
+  // Monotonic token so only the most recent settle window clears
+  // isSettling — during rapid repeated swipes an older timer must not cut a
+  // newer window short.
+  const settleToken = useSharedValue(0)
 
   // Track translation deltas from this anchor so tapping anywhere
   // doesn't snap the knob — the knob only moves with finger motion
@@ -198,6 +218,19 @@ export const CircularDial: FC<CircularDialProps> = ({
   const TAP_SLOP = 10
 
   const readoutRef = useRef<DialReadoutHandle>(null)
+
+  // Clears the settling guard once the post-release window elapses, but
+  // only if no newer interaction has opened its own window since (token
+  // check) — otherwise a stale timer from an earlier swipe could reopen
+  // prop sync mid-settle and let an echo through.
+  const clearSettlingSoon = useCallback(
+    (token: number) => {
+      setTimeout(() => {
+        if (settleToken.value === token) isSettling.value = false
+      }, SETTLE_WINDOW_MS)
+    },
+    [isSettling, settleToken]
+  )
 
   // manualActivation + direction-based commit: once total motion
   // exceeds TAP_SLOP, activate as a pan iff horizontal motion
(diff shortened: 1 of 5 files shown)


packages/k2-alpine/src/components/CircularDial/CircularDial.tsx
```diff
@@ -199,6 +219,19 @@ export const CircularDial: FC<CircularDialProps> = ({
...
+  // check) — otherwise a stale timer from an earlier swipe could reopen
+  // prop sync mid-settle and let an echo through.
+  const clearSettlingSoon = useCallback(
+    (token: number) => {
+      setTimeout(() => {
+        if (settleToken.value === token) isSettling.value = false
+      }, SETTLE_WINDOW_MS)
+    },
+    [isSettling, settleToken]
+  )
```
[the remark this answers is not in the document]
bogdandobritoiu: Good catch — fixed in 5a0a99fef. `clearSettlingSoon` now tracks the pending timeout in a ref and `clearTimeout`s the prior window before scheduling the next, so rapid repeated swipes keep at most one timer alive instead of piling up. Kept the token check as a cheap backstop.

packages/k2-alpine/src/components/CircularDial/CircularDial.tsx
```diff
@@ -298,8 +340,12 @@ export const CircularDial: FC<CircularDialProps> = ({
...
     })
 
-  // Emit onChange on every step crossing while the dial is active so the
-  // readout text (driven by the parent's controlled `value`) updates live.
+  // Emit onChange on step crossings while the dial is active so consumers —
+  // and the readout text/caption, driven by the parent's controlled
+  // `value` — track the value live. Throttled on the UI thread to
+  // EMIT_THROTTLE_MS via a leading-edge time gate (no trailing timer, so
+  // nothing is queued to fire after release); onEnd always emits the final
+  // snapped value, so throttled-away crossings are never lost.
```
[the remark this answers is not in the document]
bogdandobritoiu: Fixed in 5a0a99fef. Updated the comment: `onChange` now feeds consumers and parent-derived UI (the fiat caption); the readout number is driven by `progressSv` on the UI thread in `DialReadout` via `useAnimatedProps`, so it no longer depends on this round-trip.

ruijia1in (dismissed):

onghwan (dismissed):

packages/k2-alpine/src/components/CircularDial/CircularDial.tsx
```diff
@@ -309,8 +363,10 @@ export const CircularDial: FC<CircularDialProps> = ({
     (stepIdx, prev) => {
       if (prev === null || stepIdx === prev) return
       if (!isActive.value) return
-      const snapped = stepIdx * vStep
-      scheduleOnRN(stableOnChange, snapped)
+      const now = Date.now()
```
atn4z7, line 376: nit: `Date.now()` runs on the UI thread here. It's whitelisted in Reanimated so this should be fine — just worth confirming it doesn't warn under the new arch since it's the line the whole throttle hangs on.

packages/k2-alpine/src/components/CircularDial/CircularDial.tsx
```diff
@@ -199,6 +219,25 @@ export const CircularDial: FC<CircularDialProps> = ({
 
   const readoutRef = useRef<DialReadoutHandle>(null)
 
+  // Pending settle-window timer, so rapid repeated swipes cancel the prior
+  // one instead of piling up dozens of timeouts that all wake to no-op.
+  const settleTimerRef = useRef<ReturnType<typeof setTimeout> | null>(null)
```
atn4z7, line 224: This timer isn't cleared on unmount — if the component goes away mid-settle it fires ~250ms later and writes to the shared values. Harmless today (no React state), but an unmount cleanup would match the rAF cleanup in DialReadout:

```ts
useEffect(() => () => {
  if (settleTimerRef.current !== null) clearTimeout(settleTimerRef.current)
}, [])
```

packages/k2-alpine/src/components/CircularDial/CircularDial.tsx
```diff
@@ -298,8 +346,14 @@ export const CircularDial: FC<CircularDialProps> = ({
       scheduleOnRN(handleTap)
     })
 
-  // Emit onChange on every step crossing while the dial is active so the
-  // readout text (driven by the parent's controlled `value`) updates live.
+  // Emit onChange on step crossings while the dial is active so consumers —
```
atn4z7, line 359: onChange no longer fires on every step crossing now that it's throttled. Fine for the fiat caption, just flagging — worth a quick check that no other CircularDial consumer counts per-step emissions (analytics etc.). onCommit is unaffected.

packages/k2-alpine/src/components/CircularDial/DialReadout.tsx
```diff
@@ -377,7 +409,13 @@ export const DialReadout = forwardRef<DialReadoutHandle, DialReadoutProps>(
           pointerEvents={isEditing ? 'box-none' : 'none'}>
           <AnimatedTextInput
             ref={inputRef}
-            value={currentText}
+            // Controlled by the draft only while editing; otherwise left
+            // uncontrolled so `displayAnimatedProps` can drive the text on
+            // the UI thread without the parent's controlled `value` fighting
+            // it on every re-render. `defaultValue` seeds the first paint.
+            value={isEditing ? currentText : undefined}
```
atn4z7, line 420: nit: flipping `value` between a string and `undefined` is the controlled↔uncontrolled switch RN usually warns about. This is the standard Reanimated live-text pattern so it's probably fine — just confirm there's no console warning at the edit→idle handoff and leave it if it's clean.

packages/k2-alpine/src/components/CircularDial/DialReadout.tsx
```diff
@@ -126,6 +132,13 @@ export const DialReadout = forwardRef<DialReadoutHandle, DialReadoutProps>(
     const inputRef = useRef<TextInput>(null)
     const isEditing = draft !== null
 
+    // UI-thread mirror of `isEditing` so the display worklet can step aside
+    // and let the controlled draft own the text while the user types.
+    const isEditingSv = useSharedValue(false)
+    useEffect(() => {
+      isEditingSv.value = isEditing
```
atn4z7: nit: syncing this in an effect means `isEditingSv` trails `isEditing` by a frame, so the controlled value and the animated text both briefly drive the input at the boundary. Setting it where the draft state changes would close the gap — only worth it if QA actually sees a flicker.

packages/k2-alpine/src/components/CircularDial/CircularDial.tsx
bogdandobritoiu: Done in 52c7e7763 — added an unmount `useEffect` that `clearTimeout`s `settleTimerRef`, mirroring DialReadout's rAF cleanup, so a pending settle can't fire ~250ms after unmount.

packages/k2-alpine/src/components/CircularDial/DialReadout.tsx
bogdandobritoiu: Done in 52c7e7763 — `isEditingSv` is now set synchronously in `startEdit`/`handleBlur` instead of via the trailing effect, so it no longer lags `isEditing` by a frame. Removed the effect.

packages/k2-alpine/src/components/CircularDial/CircularDial.tsx
bogdandobritoiu: Confirmed — `Date.now()` is whitelisted in Reanimated's worklet runtime, and `LeverageWheel` already calls it inside its gesture worklets (onStart/onUpdate/onEnd), so there's in-repo precedent it doesn't warn under the new arch. Leaving as-is.

packages/k2-alpine/src/components/CircularDial/CircularDial.tsx
bogdandobritoiu: Checked all four consumers (StakeAmountScreen, PerpetualsPlaceOrder, PerpetualsClose, predictions EventDetails) — each passes a plain state setter as `onChange`; none count per-step emissions, and `onCommit` is unaffected. Safe to throttle.

packages/k2-alpine/src/components/CircularDial/DialReadout.tsx
bogdandobritoiu: Decided to keep the current `value={isEditing ? currentText : undefined}` (standard Reanimated live-text pattern) rather than refactor to mount-the-TextInput-only-while-editing. RN's TextInput doesn't emit ReactDOM's controlled/uncontrolled warning the same way — we'll confirm no warning / cursor oddity at the edit→idle handoff in device QA and revisit only if it actually misbehaves.

packages/k2-alpine/src/components/CircularDial/CircularDial.tsx
```diff
@@ -199,6 +219,25 @@ export const CircularDial: FC<CircularDialProps> = ({
 
   const readoutRef = useRef<DialReadoutHandle>(null)
 
+  // Pending settle-window timer, so rapid repeated swipes cancel the prior
+  // one instead of piling up dozens of timeouts that all wake to no-op.
+  const settleTimerRef = useRef<ReturnType<typeof setTimeout> | null>(null)
+
```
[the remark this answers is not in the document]
bogdandobritoiu: Fixed in 52c7e7763 — added an unmount cleanup that `clearTimeout`s the pending `settleTimerRef` so it can't fire after the dial unmounts.

packages/k2-alpine/src/components/CircularDial/DialReadout.tsx
```diff
@@ -126,6 +132,13 @@ export const DialReadout = forwardRef<DialReadoutHandle, DialReadoutProps>(
...
+    }, [isEditing, isEditingSv])
```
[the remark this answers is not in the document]
bogdandobritoiu: Fixed in 52c7e7763 — `isEditingSv` is now set synchronously at the edit enter/exit points (`startEdit`/`handleBlur`) and the post-paint effect was removed, closing the one-frame window where both the controlled draft and the animated text could drive the input.

packages/k2-alpine/src/components/CircularDial/DialReadout.tsx
```diff
@@ -377,7 +409,13 @@ export const DialReadout = forwardRef<DialReadoutHandle, DialReadoutProps>(
...
           <AnimatedTextInput
             ref={inputRef}
-            value={currentText}
+            // Controlled by the draft only while editing; otherwise left
+            // uncontrolled so `displayAnimatedProps` can drive the text on
+            // the UI thread without the parent's controlled `value` fighting
+            // it on every re-render. `defaultValue` seeds the first paint.
+            value={isEditing ? currentText : undefined}
+            defaultValue={currentText}
+            animatedProps={displayAnimatedProps}
```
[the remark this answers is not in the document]
bogdandobritoiu: Keeping the current `value={isEditing ? currentText : undefined}` (standard Reanimated live-text pattern) instead of the mount-on-edit refactor. RN's TextInput doesn't emit ReactDOM's controlled/uncontrolled warning the same way; we'll confirm no warning / cursor behavior in device QA and only switch to the LeverageDisplay-style mount-while-editing approach if it actually misbehaves.

ruijia1in (approved):

atn4z7 (approved):

PR #3907, opened 2026-06-16
CP-2428: Recurring swaps (DCA)

ruijia1in:
iOS: 8936
Android: 8937

## Description

**Ticket: [CP-14394]**
**Ticket: [CP-14395]**
**Ticket: [CP-14396]**
**Ticket: [CP-14398]**
**Ticket: [CP-14399]**
**Ticket: [CP-14400]**
**Ticket: [CP-14401]**

End-to-end mobile implementation of recurring (DCA) swaps. Users can open the swap modal, toggle into recurring mode, pick a token pair / per-order amount / frequency / order count, and submit a schedule whose first fill broadcasts immediately and whose subsequent fills run server-side via the Markr orchestrator. Active schedules surface in a banner on the Activity tab and at the top of the swap modal, and can be paused / unpaused / cancelled from a dedicated management screen.

### Commits in this PR

Each commit maps to one child story under the epic:

| Commit | Story | Scope |
| --- | --- | --- |
| `1b26428da` | [CP-14394](https://ava-labs.atlassian.net/browse/CP-14394) | Foundation: `@avalabs/fusion-sdk` bump (paginated `getBridgeableAssets` migration), `SWAP_RECURRING` PostHog kill-switch + `selectIsRecurringSwapsBlocked` selector, `RequestContext.RECURRING_SWAP` slot, `RECURRING_QUOTE` / `RECURRING_SCHEDULES` query keys, `RECURRING_PENDING_ACTION` MMKV key, narrow `recurringSwap/types.ts` |
| `6a87734f3` | [CP-14395](https://ava-labs.atlassian.net/browse/CP-14395) | Hooks layer: `useRecurringEligibility`, `useRecurringQuote`, `useRecurringSchedules`, `RecurringSwapContext`, `staggeredInvalidate` helper |
| `0a3f8fbbe` | [CP-14396](https://ava-labs.atlassian.net/browse/CP-14396) | Swap modal: toggle, frequency + orders chip pickers with custom-value alert, detail rows, `submitRecurringSwap` |
| `c9666e5f8` | [CP-14398](https://ava-labs.atlassian.net/browse/CP-14398) | ApprovalController integration: `<RecurrenceDetails />`, strict zod validator over the discriminated union, per-action-type side channel keyed by SDK aggregator id, EvmSigner routing (recurring excluded from Quick Swaps bypass) |
| `29eff60db` | [CP-14399](https://ava-labs.atlassian.net/browse/CP-14399) | Manage screen: schedule cards with Pause/Unpause + Cancel, `_makeOrderActionHook` factory, `pendingActionStore` (10-min TTL, reconciled by listeners) |
| `58b383c8b` | [CP-14400](https://ava-labs.atlassian.net/browse/CP-14400) | Entry points: Activity-tab banner + swap-modal row, new `rotate_right` icon |
| `16085f8ed` | [CP-14401](https://ava-labs.atlassian.net/browse/CP-14401) | Analytics, failure watcher, pending-action reconciler, per-(account, chain) `seenFailures` dedupe with silent first-observation seeding |

### Motivation and context

Recurring DCA swaps were prototyped in [CP-13663](https://ava-labs.atlassian.net/browse/CP-13663). For shipping, the Markr REST surface was lifted into `@avalabs/fusion-sdk` so web and mobile share one source of truth. The SDK's `markr.recurring` namespace now owns calldata fetching, signing, and broadcast for fill / cancel / pause / unpause. Mobile owns React Query hooks, UI, ApprovalController integration, analytics, and the side-channel that ferries display metadata to the approval modal.

### Dependencies introduced

- **`@avalabs/fusion-sdk`** bumped to consume the new `markr.recurring` namespace

### Behind a kill-switch

Everything ships behind PostHog `SWAP_RECURRING` (default `false`). When the flag is off: toggle never renders, Activity-tab banner hidden, swap-modal banner hidden, the management screen route is unreachable.

### Breaking changes

None. The only existing-flow modifications are gated by `isRecurringAggregatorId(quote.aggregator?.id)` in `EvmSigner` (false for paraswap / avalanche-evm / lombard / wrap-unwrap quotes) and `recurring.isRecurring` in `SwapScreen` (false unless the user toggles in).

## Screenshots/Videos

videos: https://ava-labs.atlassian.net/browse/CP-14393

## Testing

**Dev Testing**

> Requires PostHog `SWAP_RECURRING = true` (and `EVERYTHING = true`) in the test environment.

- Open the swap modal on the C-Chain, pick AVAX → USDC, enter an amount, confirm the **Make this a recurring swap** toggle row appears beneath the swap card; toggle on
- Confirm the disclaimer banner and the **Frequency** + **Number of orders** rows render below; tap each row and verify chip selection + the **Custom** alert paths (range guards, Save button disabled until valid)
- Confirm **Estimated total spend** = per-order × orders, formatted correctly for ≥ 1000 tokens (regression: previously displayed 1000× too small)
- Press **Next**; verify the approval modal renders with the **Scheduling recurring swap** preview block above the standard tx details, and that the summary line matches the picker selections
- Approve → success snackbar; navigate to the Activity tab and verify the **Recurring swaps** banner now shows a count
- Tap **Manage**; verify the schedule card shows the right token pair, summary line, and cadence
- Expand the card; tap **Pause** → approve modal renders the **Pausing recurring swap** preview block → approve → the card switches into the "Pausing" spinner; wait for Markr to reflect the status → button flips to **Unpause**
- Tap **Unpause** → approve → "Unpausing" spinner → row returns to Active state
- Tap **Cancel** on a paused schedule → confirm the destructive alert → approve modal → "Cancelling" spinner → row disappears from the list when status hits Cancelled
- Edge cases:
  - Set the per-order amount below the Markr minimum → confirm the error copy appears in red and **Next** is disabled
  - Toggle into recurring with a token pair that's ineligible (e.g. BTC, a cross-chain pair) → confirm the toggle hides itself
  - Flip the `SWAP_RECURRING` flag off in PostHog → confirm the toggle, both banners, and the management route all disappear
  - With recurring on, let the quote go stale (> 30s + > expiredAt) and press Next → confirm the **Quote expired — please re-confirm and try again** snackbar fires (not the generic "Recurring swap failed" copy)
- Trigger a Bitrise build and paste the link here once green

## Checklist

[CP-14394]: https://ava-labs.atlassian.net/browse/CP-14394?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CP-14395]: https://ava-labs.atlassian.net/browse/CP-14395?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CP-14396]: https://ava-labs.atlassian.net/browse/CP-14396?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CP-14398]: https://ava-labs.atlassian.net/browse/CP-14398?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CP-14399]: https://ava-labs.atlassian.net/browse/CP-14399?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CP-14400]: https://ava-labs.atlassian.net/browse/CP-14400?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CP-14401]: https://ava-labs.atlassian.net/browse/CP-14401?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CP-14394]: https://ava-labs.atlassian.net/browse/CP-14394?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CP-14395]: https://ava-labs.atlassian.net/browse/CP-14395?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


B0Y3R-AVA (commented):
## Adversarial review (Claude + Codex)

Reviewed end-to-end by two independent engines (4 Claude finder angles + a Codex pass), then every candidate was adversarially re-verified against the code. The architecture is solid, kill-switch gating is thorough, and the `_makeOrderActionHook` factory is well-consolidated. Inline comments below carry a verdict + fix. Severity key: 🔴 confirmed correctness  •  🟡 plausible / latent  •  ⚪ quality.

**Strongest signal:** the *unlimited-orders* bug (on `submitRecurringSwap.ts`) was flagged independently by both engines — worth verifying the SDK's `quote.numberOfOrders` value first, as that single fact makes it P0 or a non-issue.

**Theme to watch:** three separate `formatTokenAmount(…)` calls omit the token-decimals precision arg — the same class of “amount displayed too small” regression the PR description calls out as previously fixed.

_Adversarially **refuted** (checked, not bugs — noted so you don't re-investigate):_ (1) same-type side-channel slot collision is unreachable — a screen-wide `isPending` flag disables all same-type buttons during an in-flight approval (note: the defense is UI-only); (2) the `cancel(l|led)` regex missing American “canceled” — the real rejection path emits `"User rejected the request."` / code 4001, which matches (still worth the cleanup below); (3) `seenFailures` load→mutate→save is fully synchronous, so no re-entrancy race.

packages/core-mobile/app/new/features/recurringSwap/utils/submitRecurringSwap.ts
```diff
@@ -0,0 +1,250 @@
...
+  // change.
+  const amountPerOrderFormatted = formatTokenAmount(
+    bigintToBig(amountPerOrder, fromTokenDecimals)
+  )
+  setActiveRecurringActionContext({
+    type: 'fill',
+    fromTokenSymbol,
+    toTokenSymbol,
+    amountPerOrderFormatted,
+    numberOfOrders: quote.numberOfOrders,
```
B0Y3R-AVA, line 203: 🔴→✅ **RETRACTED — verified false against `@avalabs/fusion-sdk@0.22.0`.**

My original concern was that `quote.numberOfOrders` would be the `-1` unlimited sentinel and fail the `min(2)` schema. It isn't. The SDK has two distinct fields:
- `RecurringQuoteResponse.numberOfOrders` (what this line reads) — *"Server-side clamped value (may be 365 if the request was unlimited)"* → a finite int in `[2, 365]`.
- `RecurringOrder.numberOfOrders` (the `listOrders` shape, used on the manage screen) — `-1` when unlimited.

The `-1` only exists on the wire request and the order object; the **quote response** that feeds this context is clamped to 365 (corroborated by `totalAmountIn = amount × numberOfOrders`, which needs a finite count). So `min(2).max(365)` passes. **No bug — unlimited schedules validate fine.** Apologies for the noise.

packages/core-mobile/app/vmModule/ApprovalController/validators/shared.ts
```diff
@@ -58,6 +62,87 @@ const readAutoApproveContext = (
...
+const recurringFillContextSchema = z
+  .object({
+    type: z.literal('fill'),
+    fromTokenSymbol: z.string().min(1),
+    toTokenSymbol: z.string().min(1),
+    amountPerOrderFormatted: z.string().min(1),
+    // Markr's documented floor for finite schedules is 2 (a "1-order
+    // schedule" is just a one-shot swap). Matches the recurring details
+    // picker's `MIN_ORDERS = 2`.
+    numberOfOrders: z.number().int().min(2).max(RECURRING_FREQUENCY_VALUE_MAX),
```
B0Y3R-AVA: 🔴→✅ **RETRACTED — verified false against `@avalabs/fusion-sdk@0.22.0`.**

This schema is correct. `submitRecurringSwap` feeds it `RecurringQuoteResponse.numberOfOrders`, which the SDK documents as *"server-side clamped value (may be 365 if the request was unlimited)"* — a finite int in `[2, 365]` (`RECURRING_FREQUENCY_VALUE_MAX = 365`). The `-1` unlimited sentinel lives on `RecurringOrder` (listOrders) and the wire request, not on the quote response. So no `isUnlimited` branch is needed here. Apologies for the noise.

packages/core-mobile/app/new/features/recurringSwap/utils/submitRecurringSwap.ts
```diff
@@ -0,0 +1,250 @@
...
+  // so the schema stays decimals-agnostic.
+  //
+  // The slot is keyed by action type ('fill'), so a concurrent cancel /
+  // pause / unpause writes a different slot and won't be observed by this
+  // call's signs. The outer try/finally guarantees the slot clears even if
+  // a post-broadcast effect (analytics, snackbar, invalidate) throws —
+  // otherwise a stale 'fill' could bleed into a later unrelated swap that
+  // shares the prefix detector but predates a future SDK aggregator-id
+  // change.
+  const amountPerOrderFormatted = formatTokenAmount(
```
B0Y3R-AVA, line 195: 🔴 **Approval preview shows `0.00` for small per-order amounts.** `formatTokenAmount` defaults to 2-decimal rounding when no precision arg is passed, so `0.001 AVAX → "0.00"`. This formatted string is consumed verbatim by `<RecurrenceDetails />`. The sibling `totalSpend` site (`RecurringDetailsRows.tsx:373`) already passes `fromTokenDecimals` — do the same here. (Same “amount too small” class the PR description says was previously fixed.)

packages/core-mobile/app/new/features/recurringSwap/screens/RecurringSchedulesScreen.tsx
```diff
@@ -0,0 +1,705 @@
...
+  return { symbol: shortenAddress(address), decimals: 18 }
+}
+
+function formatSummary(
+  s: RecurringOrder,
+  fromToken: ResolvedToken,
+  toToken: ResolvedToken
+): string {
+  const cadence = formatFrequencyShort(s.frequency)
+  const amount = formatTokenAmount(bigintToBig(s.amount, fromToken.decimals))
```
B0Y3R-AVA: 🔴 **Manage-screen summary rounds per-order amount to 2 decimals.** Same omission as the approval preview — `formatTokenAmount(bigintToBig(s.amount, fromToken.decimals))` with no precision arg renders any sub-0.005 order as `0.00`. Pass `fromToken.decimals` as the 2nd arg.

packages/core-mobile/app/new/features/recurringSwap/screens/RecurringSchedulesScreen.tsx
```diff
@@ -0,0 +1,705 @@
...
+      'decimals' in match && typeof match.decimals === 'number'
+        ? match.decimals
+        : 18
+    return {
+      symbol: match.symbol ?? shortenAddress(address),
+      decimals,
+      logoUri: match.logoUri
+    }
+  }
+  return { symbol: shortenAddress(address), decimals: 18 }
```
B0Y3R-AVA: 🟡 **18-decimal fallback corrupts amounts for non-18-decimal tokens missing from the local catalog.** When a schedule references a token not in the active network's `useNetworkContractTokens`, this defaults to `decimals: 18`; a 6-decimal USDC order (`1_000_000n`) then formats to ~`0`. Reachable for un-indexed/removed tokens. Consider resolving decimals from the schedule/SDK rather than defaulting.

---
**↻ Re-verified against SDK source.** Refinement: the *matched-token* path is fine — recurring `tokenIn` is an ERC-20, and `ERC20Token` in `@avalabs/vm-module-types` always carries `decimals`, so listed tokens resolve correctly. Only the **not-in-catalog** branch (this line) hits the 18 default. The SDK's `RecurringOrder` (`fusion-sdk@0.22`) carries **no decimals field**, so the correct fix is an on-chain `decimals()` lookup (or hiding the amount) for unlisted tokens — not reading it off the order. Compounds #4: a 6-dp token rendered at 18 dp collapses to `0.00`.

packages/core-mobile/app/new/features/recurringSwap/store/listeners.ts
```diff
@@ -0,0 +1,231 @@
...
+
+  const byId = new Map(schedules.map(s => [s.orderId, s]))
+  for (const orderId of orderIds) {
+    if (isExpired(orderId)) {
+      clearPending(orderId)
+      continue
+    }
+    const match = byId.get(orderId as `0x${string}`)
+    const entry = pending[orderId]
+    if (!match || !entry) {
```
B0Y3R-AVA, line 163: 🔴 **Account/network switch wrongly clears another account's in-flight pending action.** `pendingActionStore` is keyed by `orderId` only, but `reconcilePendingActions` runs against a *single* account/chain's schedule list and clears any entry not present in it. Switch from account A (pending cancel on `0xABC`) to B → B's query-success event reconciles the global map against B's list → `clearPending('0xABC')`. Back on A, the spinner is gone and buttons re-enable mid-flight, so the user can fire a second redundant on-chain action. **Fix:** scope reconciliation to the entry's (account, chain), or key the pending map by them. (Amplified by the `staggeredInvalidate` timer leak below.)

---
**↻ Re-verified — worse than first stated.** `ReactQueryProvider.tsx:27` sets `gcTime: Infinity`, so the previous account's `['recurringSchedules', addr, chain]` query is **never garbage-collected**. Any later success/refetch event from *either* account reconciles the global orderId-keyed map against that account's list, so the cross-account clobber is **bidirectional and permanent**, not only at the instant of switch.

packages/core-mobile/app/new/features/recurringSwap/store/listeners.ts
```diff
@@ -0,0 +1,231 @@
...
+ */
+function reconcilePendingActions(schedules: readonly RecurringOrder[]): void {
+  const { pending, clearPending, isExpired } = pendingActionStore.getState()
+  const orderIds = Object.keys(pending)
+  if (orderIds.length === 0) return
+
+  const byId = new Map(schedules.map(s => [s.orderId, s]))
+  for (const orderId of orderIds) {
+    if (isExpired(orderId)) {
+      clearPending(orderId)
```
B0Y3R-AVA, line 158: 🟡 **A reverted/dropped action TX leaves a stuck spinner for the full 10-min TTL.** `isPendingActionResolved` only clears on reaching the destination status (or this TTL). If a cancel/pause/unpause broadcasts but then reverts on-chain, the status never advances, so the card sits non-interactive on “Pausing…” with buttons disabled until the TTL fires — with no failure feedback. Bounded, but consider detecting the failed-action case explicitly.

---
**↻ Re-verified against SDK source → upgrading 🟡→🔴.** `RecurringOrderStatus` in `fusion-sdk@0.22` is a closed 4-value enum (`active/completed/cancelled/paused`) with **no failed/reverted state**, and `executeCancellation/Pause/Unpause` resolve at **broadcast**, not confirmation (`_namespace.ts:314-320`). A reverted/dropped action TX never advances the status `isPendingActionResolved` keys on, so the 10-min TTL is genuinely the *only* escape — confirmed, not just plausible.

packages/core-mobile/app/new/features/recurringSwap/hooks/_makeOrderActionHook.ts
```diff
@@ -0,0 +1,234 @@
...
+          address: args.address as Address,
+          sourceChain: args.sourceChain
+        })
+
+        // Mark pending after broadcast so the card's button stays in its
+        // "-ing" spinner state until Markr's next listOrders refetch
+        // reports the destination status. The reconciler in listeners.ts
+        // clears this entry once the server reflects the new state (or
+        // the TTL elapses).
+        pendingActionStore.getState().markPending(args.orderId, config.type)
```
B0Y3R-AVA, line 161: 🔴 **Pending state is lost on broadcast-then-reject.** `markPending` runs only *after* `await execute()` resolves and is never called in `catch`/`finally`. If the SDK broadcasts the TX but the awaited promise then rejects (post-broadcast timeout/read error), the action hit the chain but the UI shows a failure snackbar with no pending spinner — the user may re-tap and double-submit. Also, the shared `pendingActionStore` entry is unset for the entire sign+broadcast window (the local `isPending` covers the in-place button, but not other consumers). Consider marking pending optimistically around the broadcast.

packages/core-mobile/app/new/features/recurringSwap/utils/staggeredInvalidate.ts
```diff
@@ -0,0 +1,68 @@
...
+  if (existing) {
+    for (const id of existing) clearTimeout(id)
+    existing.clear()
+  }
+
+  const set = new Set<ReturnType<typeof setTimeout>>()
+  pending.set(k, set)
+
+  for (const delayMs of STAGGERED_DELAYS_MS) {
+    const id = setTimeout(() => {
```
B0Y3R-AVA, line 42: 🟡 **Module-level timers leak across unmount/account-switch.** These 5/15/30s `setTimeout`s are stored in a module-scoped map and only cleared by a same-key re-schedule — never on unmount or account switch. After navigating away/switching accounts they still fire and, via React Query prefix matching on `[RECURRING_SCHEDULES]`, invalidate all accounts. On its own that's a benign extra refetch, but it re-triggers the cross-account reconcile clear (`listeners.ts:158`). Clear timers on unmount.

---
**↻ Re-verified → upgrading 🟡→🔴.** The timers invalidate the bare `['recurringSchedules']` key, which (React Query v5 `partialMatchKey`) prefix-matches **every** account/chain query, and `invalidateQueries` defaults to `refetchType:'active'`. So a timer firing up to 30s after an account switch refetches the *new* account's query → success event → `reconcilePendingActions` → clears the *previous* account's in-flight entry. This is a time-delayed trigger for #2, not benign extra refetches.

packages/core-mobile/app/new/features/swap/screens/SwapScreen.tsx
```diff
@@ -983,21 +1209,125 @@ export const SwapScreen = (): JSX.Element => {
     isSeedlessWallet
   })
 
+  const isRecurringReady =
```
B0Y3R-AVA, line 1185: 🔴 **Recurring submit is not hard-gated on the minimum amount.** `isRecurringReady` checks `!!fromTokenValue` (truthy only), not `>= recurringMinimumTransferAmount` and not `!validationError`. The below-minimum error is computed and displayed but never reaches `canSubmit`, so **Next stays enabled** and a sub-minimum `amountPerOrder` is submitted — the one-shot path gates this in `canSwap`. Fold the minimum / `validationError` check into `isRecurringReady`.

packages/core-mobile/app/new/features/swap/screens/SwapScreen.tsx
```diff
@@ -770,12 +964,34 @@ export const SwapScreen = (): JSX.Element => {
...
 
+  // Reset recurring toggle only when the pair becomes ineligible (e.g. user
+  // picks a cross-chain or unsupported destination) or the feature flag flips
+  // off. Clearing the amount intentionally does NOT reset the flag — the user's
+  // on/off choice persists until they manually toggle off or the token is no
+  // longer supported.
+  useEffect(() => {
+    const recurringDisallowed = isRecurringBlocked || !eligibility.eligible
+    if (recurringDisallowed && recurring.isRecurring) {
+      recurring.setIsRecurring(false)
```
B0Y3R-AVA, line 962: 🟡 **No synchronous kill-switch guard in the submit path.** This reset is a deferred `useEffect`; `handleNext` / `submitRecurring` / `submitRecurringSwap` never re-check `isRecurringBlocked` or `eligibility.eligible`. If the flag flips (or the pair becomes ineligible) and the user taps Next in the same commit frame, a recurring schedule is still dispatched. Low reachability (single-frame race, partly shadowed by the quote/amount presence guards), but the kill-switch is exactly what should be hard to bypass — add a guard in `submitRecurring`.

---
**↻ Re-verified → downgrading 🟡→⚪ (theoretical).** No synchronous guard exists, but it isn't user-reachable: the reset `useEffect` commits before any tap can land, and even in a race `submitRecurring` requires `recurringQuote.data`, which tears down when the flag flips (the quote hook is gated on `isRecurring`/eligibility). Worst case is submitting a swap that was valid microseconds earlier. A one-line `if (isRecurringBlocked || !eligibility.eligible) return` in `submitRecurring` closes it cheaply, but there's no real bug.

packages/core-mobile/app/new/features/recurringSwap/services/activeActionContext.ts
```diff
@@ -0,0 +1,112 @@
...
+    aggregatorId !== undefined &&
+    aggregatorId.startsWith(RECURRING_AGGREGATOR_PREFIX)
+  )
+}
+
+export function aggregatorIdToActionType(
+  aggregatorId: string | undefined
+): RecurringActionType | undefined {
+  if (!aggregatorId) return undefined
+  if (aggregatorId === 'markr-recurring') return 'fill'
```
B0Y3R-AVA, line 107: 🟡 **`isRecurringAggregatorId` (startsWith) and this exact-equality map can de-sync.** A future suffixed id (e.g. `markr-recurring-v2`) would pass the loose recurring check in `EvmSigner` (suppressing auto-approve) but map to no action type here → no `RECURRING_SWAP` context → a recurring action rendered as a bare one-shot swap with no `<RecurrenceDetails />`. Unreachable with today's 4 exact ids; latent. **Fix:** derive `isRecurring` from `aggregatorIdToActionType(…) !== undefined` so both gates share one source of truth.

---
**↻ Re-verified against SDK source — today-safe, latent only.** `fusion-sdk@0.22` emits exactly these 4 ids (`_namespace.ts:350` → `markr-recurring`; `:383` → `markr-recurring-${'cancel'|'pause'|'unpause'}`), so all four flows render correctly today. The `startsWith` vs exact-equality divergence only bites if a future SDK adds a 5th `markr-recurring-*` id. Future-proofing, not a present bug.

packages/core-mobile/app/new/features/recurringSwap/hooks/_makeOrderActionHook.ts
```diff
@@ -0,0 +1,234 @@
...
+
+// Matches the user-rejection detector in SwapScreen.submitRecurring. The
+// underlying signer throws an Error whose message starts with
+// "User rejected" / "User cancelled" / "User canceled" when the user taps
+// Reject (or closes the modal) on the in-app approval sheet. These aren't
+// failures we should surface — they're the user's deliberate action.
+function isUserRejectionError(err: unknown): boolean {
+  const message =
+    err instanceof Error ? err.message : typeof err === 'string' ? err : ''
+  return /User (rejected|cancel(l|led))/i.test(message)
```
B0Y3R-AVA, line 217: ⚪ **Duplicated user-rejection detector + dead branch.** This regex is duplicated in `SwapScreen.tsx:1273`, and the comment claims it matches `"User canceled"` — but `cancel(l|led)` does not match the single-l American spelling (no live path emits it today, so not a bug). A canonical detector already exists: `fusionErrors.isUserRejectionError` and the code-4001 check in `walletConnect/utils.isUserRejectedError`. Reuse the code-based detector instead of two divergent string regexes.

---
**↻ Re-verified — live-bug ruled out.** Traced every rejection path: in-app / Ledger / Keystone all normalize to messages containing "rejected" (`@metamask/rpc-errors` 4001 → "User rejected the request."), so nothing slips through the regex. Purely duplication + a wrong comment; reusing `fusionErrors.isUserRejectionError` (or a `code === 4001` check) is the fix.

packages/core-mobile/app/new/features/approval/components/RecurrenceDetails.tsx
```diff
@@ -0,0 +1,98 @@
+import React from 'react'
+import { Text, View } from '@avalabs/k2-alpine'
+import type { RecurringSwapApprovalContext } from 'vmModule/ApprovalController/validators/shared'
+
+type Props = {
+  context: RecurringSwapApprovalContext
+}
+
+const unitToLabel = (unit: string, value: number): string =>
```
B0Y3R-AVA: ⚪ **Duplicates `formatFrequencyShort`.** This hand-rolled pluralizer is a third place frequency formatting lives (alongside `formatFrequency` used in `RecurringDetailsRows`). Reuse `formatFrequencyShort` from `recurringSwap/utils/formatFrequency` (harmonize the `value === 1` singular form when consolidating).

---
**↻ Re-verified — equivalence overstated.** `formatFrequencyShort` is **not** a clean drop-in: for `value === 1` it renders "every **day**" whereas this code renders "every **1 day**". Still worth consolidating, but only with a conscious decision on that singular-form copy change.

packages/core-mobile/app/new/features/recurringSwap/utils/seenFailures.ts
```diff
@@ -0,0 +1,107 @@
+import Logger from 'utils/Logger'
+import { commonStorage } from 'utils/mmkv/storages'
+
+// Cap the persisted set so a power user with years of recurring schedules
+// can't grow it unboundedly. Each entry is `${orderId}:${executionIndex}` —
+// realistically ~50 bytes — so 1000 entries ≈ 50KB per (account, chain). On
+// overflow we drop the oldest insertions (Set preserves insertion order),
+// which is fine because the failure snackbar is a one-shot notification and
+// stale "seen" state for a long-dead schedule has no UI consequence.
+const MAX_SEEN_FAILURES = 1000
```
B0Y3R-AVA, line 11: ⚪→✅ **RETRACTED — verified false against the actual helpers.** `BoundedMap` is a `Map` with per-insert eviction and **no persistence**; the mmkv array helpers persist a bare `T[]` with no slot for the `initialised` flag or the legacy-shape migration this file performs. The real substance of `seenFailures` — per-(account, chain) MMKV keying, the `initialised` envelope, parse-failure recovery — isn't covered by either, and the cap itself is ~3 lines. Reuse would not simplify it. Apologies for the noise.


<details><summary>Original comment (withdrawn)</summary>

⚪ **Re-implements an existing primitive.** This capped, insertion-ordered, MMKV-persisted Set (FIFO eviction + parse/validate) duplicates `common/utils/boundedMap.ts` + the mmkv array helpers (`saveArrayToStorage`/`loadArrayFromStorage`). Composing those removes ~100 lines of bespoke, separately-tested storage code.

</details>

packages/core-mobile/app/new/features/swap/services/signers/EvmSigner.ts
```diff
@@ -232,6 +236,14 @@ export function createEvmSigner(
     // Markr (proves networkFees was loaded). The bypass skips the
     // modal's fee picker, so missing-fees would broadcast as 0.
     const hasUpstreamFees = typeof tx.maxFeePerGas === 'bigint'
+    // §A13: recurring synthetic Quotes have `serviceType === ServiceType.MARKR`
+    // (same as one-shot Markr swaps), so the existing serviceType gate alone
+    // would let recurring through. Exclude them explicitly — recurring TXs
+    // must always render the approval modal so the user sees the
+    // "Scheduling / Cancelling / Pausing / Unpausing recurring swap" preview.
+    const isRecurring = isRecurringAggregatorId(
```
B0Y3R-AVA: ⚪→✅ **RETRACTED — verified false against SDK source.** The side channel is necessary, not wrong-altitude. For the first fill the **SDK** (`markrRecurring.executeFirstFill`) owns request creation and signing — the app never holds the `request`, so there is no `request.context` to thread through (unlike a normal in-app `request()` call). `SWAP_AUTO_APPROVE` isn't threaded by the app either; it's synthesized *inside* `EvmSigner` from the quote it's handed. The recurring display strings aren't on the synthetic quote, so a slot into `signOne` (the only app-controlled seam) is the right design — and the `RECURRING_SWAP` type comment already documents exactly this. Apologies for the noise.


<details><summary>Original comment (withdrawn)</summary>

⚪ (altitude / optional) The recurring path is special-cased into the shared signer via `markr-recurring*` string-sniffing + a module-global side channel (`activeActionContext`), while `SWAP_AUTO_APPROVE` threads its context directly through `request.context`. Recurring could do the same — keeping recurring-specific policy out of the shared EVM signing path and removing the global-slot lifetime concern.

</details>

packages/core-mobile/app/new/features/recurringSwap/services/activeActionContext.ts
ruijia1in: skip. assuming we always just use markr-recurring

packages/core-mobile/app/new/features/recurringSwap/hooks/_makeOrderActionHook.ts
ruijia1in: different from fusionErrors.isUserRejectionError  and walletConnect/utils.isUserRejectedError

atn4z7 (commented):
Review of the recurring-swaps work. Solid PR overall — the security model holds, the kill-switch gating is correct, and the hard parts (dedupe seeding, staggered invalidation, the 1000x total-spend fix) are right and tested. No blockers. Inline comments below: a few RISK items worth confirming, plus quality/nits.

**Follow-ups (not for this PR):**
- The `onAppUnlocked` schedules listener could be a static `addRecurringSwapListeners(startListening)` alongside the failure watcher, matching `addFusionListeners` — drops the `@ts-expect-error`.
- If a failed-order push notification is planned alongside the filled-order one, the whole failure-watcher + `seenFailures` setup is interim (foreground-only, drops failures that happen while the app is closed) and can be removed once push lands.
- If the SDK ever echoes back a `signerContext`, the `activeActionContext` Map can be deleted entirely.

packages/core-mobile/app/new/features/recurringSwap/services/activeActionContext.ts
```diff
@@ -0,0 +1,112 @@
...
+export type ActiveRecurringActionContext =
+  | RecurringFillPreview
+  | RecurringOrderActionPreview
+
+export type RecurringActionType = ActiveRecurringActionContext['type']
+
+// Per-type slots. The signer derives the type from the quote's
+// `aggregator.id` and reads from the matching slot — concurrent flows of
+// different types stay disjoint.
+const slots: Map<RecurringActionType, ActiveRecurringActionContext> = new Map()
```
atn4z7, line 58: **RISK** — the slot is keyed by action type, not by request id. Two same-type actions in flight (two fills, two cancels) would overwrite each other's slot, and the second's metadata could end up on the first's approval modal. Safe today only because the UI gates (`recurringSubmitting` / `pendingActionStore`) prevent that — correctness lives in other files. Blast radius is display-only (wrong preview text, not a wrong tx), and both stashes pass schema validation so the malformed-context gate won't catch a mix-up. Real fix is an SDK passthrough so the metadata rides with the request; until then this is fine but worth a comment noting the dependency on the UI gates.

packages/core-mobile/app/new/common/components/ScrollScreen.tsx
```diff
@@ -433,9 +433,15 @@ export const ScrollScreen = ({
...
+              // Footer overlay's full occlusion = measured footer height +
+              // its `insets.bottom + 16` wrapper padding. Without including
+              // `insets.bottom` here the bottom of the scroll content sits
+              // ~18px under the footer on iOS notched devices and can't be
+              // scrolled into view. Matches the non-keyboard branch below
+              // (CP-14413).
               paddingBottom: disableStickyFooter
                 ? insets.bottom + 32
-                : (footerLayout?.height ?? 0) + 32,
+                : (footerLayout?.height ?? 0) + insets.bottom + 48,
```
atn4z7: **RISK** — this went from `+ 32` to `+ insets.bottom + 48`, which adds ~`insets.bottom`+16 of bottom padding to *every* sticky-footer ScrollScreen, not just the recurring screens. Right fix for the footer occlusion, but it's app-wide — please confirm that's intended.

packages/core-mobile/app/new/features/recurringSwap/components/RecurringDetailsRows.tsx
```diff
@@ -0,0 +1,520 @@
...
+      amountPerOrder === undefined ||
+      numberOfOrders === undefined ||
+      numberOfOrders === UNLIMITED_ORDERS ||
+      fromTokenDecimals === undefined ||
+      !fromTokenSymbol
+    ) {
+      return null
+    }
+    const totalRaw = amountPerOrder * BigInt(numberOfOrders)
+    return `${formatTokenAmount(
```
atn4z7, line 404: **RISK** — passing token `decimals` (18 for most tokens) as `formatTokenAmount`'s `maximumFractionDigits` lets total spend render up to 18 fractional digits. The rest of the app calls `formatTokenAmount(x)` with the default 2. The math is correct now (the 1000x fix is good) — just drop the second arg for display.

packages/core-mobile/app/new/features/recurringSwap/utils/seenFailures.ts
```diff
@@ -0,0 +1,107 @@
...
+  // Trim from the front (oldest insertions) when over the cap. Array.from on a
+  // Set preserves insertion order, so this keeps the most recently observed
+  // failures — the ones whose snackbars the user most recently saw.
+  const arr = Array.from(set)
+  const trimmed =
+    arr.length > MAX_SEEN_FAILURES
+      ? arr.slice(arr.length - MAX_SEEN_FAILURES)
+      : arr
+  const payload: PersistedShape = { initialised: true, entries: trimmed }
+  commonStorage.set(
```
atn4z7: **RISK** — no try/catch around `commonStorage.set`. If the write fails, `initialised` never flips to `true`, so the next refetch re-seeds and silently suppresses those failures; a failed post-seed write can also re-fire a snackbar on restart. Low severity (one-shot toast) but the failure mode is invisible.

packages/core-mobile/app/new/features/recurringSwap/hooks/useRecurringQuote.ts
```diff
@@ -0,0 +1,116 @@
...
+    !!params.amountPerOrder &&
+    !!params.numberOfOrders &&
+    !!params.frequency
+
+  return useQuery<RecurringQuoteResponse>({
+    enabled,
+    // eslint-disable-next-line @tanstack/query/exhaustive-deps -- intentional fine-grained queryKey
+    queryKey: [
+      ReactQueryKeys.RECURRING_QUOTE,
+      params.fromToken?.networkChainId,
```
atn4z7: **QUALITY** — the key includes `fromToken.networkChainId` but not `toToken.networkChainId`. Asymmetric — for a same-chain recurring swap it's latent, but key both chains or neither.

packages/core-mobile/app/new/features/recurringSwap/components/RecurringDetailsRows.tsx
```diff
@@ -0,0 +1,520 @@
...
+  { id: '5', label: '5' },
+  { id: '10', label: '10' },
+  { id: '15', label: '15' },
+  { id: '20', label: '20' },
+  { id: 'unlimited', label: 'Unlimited' },
+  { id: 'custom', label: 'Custom' }
+]
+
+const MIN_ORDERS = 2
+const MAX_ORDERS = RECURRING_FREQUENCY_VALUE_MAX
```
atn4z7: **QUALITY** — `MAX_ORDERS` borrows `RECURRING_FREQUENCY_VALUE_MAX` (the frequency-value cap, 365). It's the right number today only by coincidence — an upstream bump of the frequency cap would silently move the order ceiling. Give it its own constant. Same reuse in the validator (`shared.ts`).

packages/core-mobile/app/new/features/recurringSwap/store/listeners.ts
```diff
@@ -0,0 +1,231 @@
...
+  let failuresDirty = false
+
+  for (const failure of schedule.failures) {
+    const key = makeFailureKey(schedule.orderId, failure.executionIndex)
+    if (seenFailures.has(key)) continue
+
+    seenFailures.add(key)
+    failuresDirty = true
+
+    showSnackbar('Recurring swap execution failed')
```
atn4z7: **QUALITY** — N new failures in one refetch fire N identical toasts. The seeding path prevents this storm on first observation, but the steady-state path has no per-refetch coalescing. Rare, but worth a 'one snackbar per refetch' guard.

packages/core-mobile/app/new/features/recurringSwap/utils/submitRecurringSwap.ts
```diff
@@ -0,0 +1,250 @@
...
+  // the approval modal then immediately fail. Catch it synchronously so the
+  // caller can surface "Quote expired" without any modal flicker, and invalidate
+  // the cached quote so a retry refetches instead of replaying the same stale
+  // payload from React Query's 30s staleTime window. Throw the same
+  // `InvalidParamsError(QUOTE_EXPIRED)` class as the SDK so SwapScreen's
+  // `isInvalidParamsError` catch lights up the "Quote expired" snackbar
+  // (the previous `new Error('QUOTE_EXPIRED:...')` slipped through to the
+  // generic "try again" copy).
+  if (Date.now() >= quote.expiredAt * 1000) {
+    queryClient.invalidateQueries({
```
atn4z7, line 167: **QUALITY** — prefix-only invalidation clears every cached recurring quote, not just the expired one. Harmless with a single active quote, but broader than needed.

packages/core-mobile/app/new/features/recurringSwap/hooks/useRecurringEligibility.ts
```diff
@@ -0,0 +1,106 @@
...
+      Logger.info(
+        '[useRecurringEligibility] checkEligibility threw — treating as ineligible',
+        String(err)
+      )
+      return {
+        eligible: false,
+        reason: RecurringEligibilityReason.UnsupportedSourceChain
+      }
+    }
+  }, [isFusionServiceReady, fromToken, toToken, ownerAddress, amount])
```
atn4z7: **QUALITY** — this memo depends on the whole `fromToken`/`toToken` objects, so it re-runs on unrelated ref changes from the swap store. That's the exact churn `useRecurringQuote` was written to avoid (it keys on stable primitives). `checkEligibility` is cheap so not a correctness bug, just inconsistent with the sibling hook.

packages/core-mobile/app/new/features/recurringSwap/utils/submitRecurringSwap.test.ts
```diff
@@ -0,0 +1,227 @@
...
+      isUnlimited: false,
+      frequency: { unit: 'week', value: 1 }
+    })
+    // No finally-clear assertion: the slot is intentionally left in place
+    // so the dispatched request's captured snapshot still matches if the
+    // SDK fires a second signOne (e.g. swap leg after approve leg). The
+    // next recurring action's setActive overwrites unconditionally.
+  })
+
+  it('marks isUnlimited=true in analytics when numberOfOrders is UNLIMITED_ORDERS', async () => {
```
atn4z7: **QUALITY** — this 'unlimited' test passes `UNLIMITED_ORDERS` as a param but reuses the finite `QUOTE`, so no test ever feeds a real unlimited quote through the validator. Worth an end-to-end unlimited case (the value the snapshot writes is `quote.numberOfOrders`, server-clamped to 365 — confirm that path).

packages/core-mobile/app/types/analytics.ts
```diff
@@ -486,6 +486,38 @@ export type AnalyticsEvents = {
...
+      amountPerOrder: string
+      numberOfOrders: number
+      isUnlimited: boolean
+      intervalSeconds: number
+    }
+  }
+  RecurringSwapCancelledByUser: {
+    encrypted: {
+      orderId: string
+      chainId: number
```
atn4z7: **QUESTION** — all four recurring events nest `chainId` inside `encrypted`. The data team usually wants `chainId` in plaintext so they can segment without decrypting, and there's precedent here (`PortfolioTokenSelected`, `SendTransactionFailed` keep it top-level). Did they sign off on encrypting it? If not, easier to pull `chainId` out into the plaintext props now (keep `orderId`/`scheduleUuid`/amounts encrypted) than to migrate the schema later.

packages/core-mobile/app/new/features/swap/services/signers/EvmSigner.ts
atn4z7: **NIT** — `isRecurringAggregatorId(stepDetails.quote.aggregator?.id)` is recomputed here and again at ~291 and ~413. Hoist to a single const.

packages/core-mobile/app/new/features/approval/components/RecurrenceDetails.tsx
```diff
@@ -0,0 +1,98 @@
...
+        </Text>
+      </View>
+    )
+  }
+
+  // type === 'fill' — full schedule preview matching the Figma frame
+  // (`21654-62903`). The hook pre-formats `amountPerOrderFormatted` so the
+  // ApprovalScreen side stays decimals-agnostic. Explicit narrow keeps TS
+  // happy across the discriminated-union branch above.
+  if (context.type !== 'fill') return <></>
```
atn4z7, line 64: **NIT** — this `if (context.type !== 'fill')` is unreachable; the earlier branch already returns for cancel/pause/unpause and the union has no other members. Fine as a TS narrow, just flagging the redundant runtime check.

packages/core-mobile/app/new/features/recurringSwap/components/RecurringDetailsRows.tsx
```diff
@@ -0,0 +1,520 @@
...
+  const selectedOrdersChip = useMemo(
+    () => ordersToChipId(numberOfOrders),
+    [numberOfOrders]
+  )
+
+  const promptCustomOrders = useCallback(() => {
+    const seed =
+      typeof numberOfOrders === 'number' &&
+      numberOfOrders !== UNLIMITED_ORDERS &&
+      ![5, 10, 15, 20].includes(numberOfOrders)
```
atn4z7: **NIT** — `[5, 10, 15, 20]` is hardcoded here while the same set lives in `ORDERS_PRESET` and `ordersToChipId`. Three copies — drift risk if presets change.

packages/core-mobile/app/new/features/recurringSwap/hooks/useRecurringSchedulesListener.ts
```diff
@@ -0,0 +1,44 @@
...
+): void => {
+  const dispatch = useDispatch()
+
+  // `dispatch(addListener(...))` returns the unsubscribe fn when the
+  // listener middleware is registered, but the default redux typings for
+  // `dispatch` don't reflect that overload — the return is widened to
+  // `unknown` and the `useEffect` cleanup contract rejects it. Use
+  // `@ts-expect-error` (not `@ts-ignore`) so this suppression breaks
+  // loudly the day redux's typings catch up.
+  // @ts-expect-error see comment above
```
atn4z7: **NIT** — the `@ts-expect-error` sits above the whole `useEffect`, so it silences more than the `dispatch(addListener(...))` return it's meant for. Narrow it to that statement.

packages/core-mobile/app/new/features/recurringSwap/hooks/_makeOrderActionHook.ts
```diff
@@ -0,0 +1,234 @@
...
+    )
+
+    // Empty deps below are intentional: `config` is captured by the outer
+    // `makeOrderActionHook(...)` call at module-import time (not on each
+    // render), so it's referentially stable for the lifetime of the app.
+    // Don't "fix" this by adding `[config]` — that would change `run`'s
+    // identity unnecessarily and break the SchedulesScreen's stable
+    // `mutate` reference that downstream `useCallback`s depend on.
+
+    const run = useCallback(async (args: OrderActionArgs): Promise<void> => {
```
atn4z7, line 126: **NIT** — `run` is fire-and-forget with no in-flight guard, so two quick `mutate` calls start two runs and the first to finish flips `isPending` off while the second is still going. The disabled button is the practical guard, but there's no re-entrancy protection in the hook itself.

packages/core-mobile/app/new/features/recurringSwap/store/listeners.ts
```diff
@@ -0,0 +1,231 @@
...
+  const orderIds = Object.keys(pending)
+  if (orderIds.length === 0) return
+
+  const byId = new Map(schedules.map(s => [s.orderId, s]))
+  for (const orderId of orderIds) {
+    if (isExpired(orderId)) {
+      clearPending(orderId)
+      continue
+    }
+    const match = byId.get(orderId as `0x${string}`)
```
atn4z7: **NIT** — `orderId as `0x${string}`` is a no-op at runtime (Map compares by value, `orderId` is a plain string from `Object.keys`). Slightly misleading; drop the cast.

packages/core-mobile/app/new/features/recurringSwap/store/listeners.test.ts
```diff
@@ -0,0 +1,461 @@
...
+    })
+  }
+}))
+
+// ─── Subject under test ───────────────────────────────────────────────────────
+
+import AnalyticsService from 'services/analytics/AnalyticsService'
+import { showSnackbar } from 'common/utils/toast'
+import { RecurringOrderStatus, type RecurringOrder } from '../types'
+import { startRecurringFailureWatcher } from './listeners'
```
atn4z7, line 104: **NIT** — there's no `failureWatcher.ts`; this suite tests `listeners.ts` (imports `startRecurringFailureWatcher` from `./listeners`). The colocated convention here is test-name = file-name (cf. `pendingActionStore.test.ts`). Suggest renaming to `listeners.test.ts` — also more accurate since it exercises both the watcher and the reconciler.

packages/core-mobile/app/new/features/swap/services/signers/EvmSigner.ts
```diff
@@ -232,6 +236,14 @@ export function createEvmSigner(
     // Markr (proves networkFees was loaded). The bypass skips the
     // modal's fee picker, so missing-fees would broadcast as 0.
     const hasUpstreamFees = typeof tx.maxFeePerGas === 'bigint'
+    // §A13: recurring synthetic Quotes have `serviceType === ServiceType.MARKR`
```
atn4z7: **NIT** — `§A13` references an internal spec a future code reader won't have. The explanation itself is good — just drop the label prefix. Same applies to the other `§A13`/`§A13.5`/`AC4` references (this file, `listeners.ts`, `shared.ts`, `submitRecurringSwap.ts`, `reactQueryKeys.ts`, and a couple of test files).

packages/core-mobile/app/new/common/components/ScrollScreen.tsx
ruijia1in: that is intended

packages/core-mobile/app/new/features/swap/services/signers/EvmSigner.ts
ruijia1in: as a follow up pr

packages/core-mobile/app/new/features/recurringSwap/utils/submitRecurringSwap.ts
ruijia1in: fixed.

packages/core-mobile/app/new/features/recurringSwap/screens/RecurringSchedulesScreen.tsx
ruijia1in: fixed.

packages/core-mobile/app/new/features/approval/components/RecurrenceDetails.tsx
ruijia1in: fixed.

packages/core-mobile/app/new/features/recurringSwap/components/RecurringDetailsRows.tsx
ruijia1in: fixed.

packages/core-mobile/app/new/features/recurringSwap/utils/seenFailures.ts
ruijia1in: fixed.

packages/core-mobile/app/new/features/recurringSwap/hooks/useRecurringQuote.ts
ruijia1in: fixed.

packages/core-mobile/app/new/features/recurringSwap/hooks/useRecurringSchedulesListener.ts
ruijia1in: fixed.

packages/core-mobile/app/new/features/recurringSwap/hooks/_makeOrderActionHook.ts
ruijia1in: fixed.

packages/core-mobile/app/new/features/recurringSwap/store/listeners.ts
ruijia1in: fixed.

packages/core-mobile/app/new/features/recurringSwap/store/listeners.test.ts
ruijia1in: fixed.

packages/core-mobile/app/new/features/swap/services/signers/EvmSigner.ts
ruijia1in: fixed.

atn4z7: Thanks — looks like the `AC4` / `§A13.5` / `post-§A13` labels got cleaned up everywhere else (listeners.ts, shared.ts, reactQueryKeys.ts, the tests), but the three `§A13` ones are still here in EvmSigner.ts at lines 239, 283, and 405. Mind dropping those too and then I'll resolve this?

ruijia1in: removed

packages/core-mobile/app/new/features/recurringSwap/services/activeActionContext.ts
ruijia1in: will handle the follow-ups in a separate PR, leave this for now since it will be deleted entirely anyway.

packages/core-mobile/app/new/features/swap/services/signers/EvmSigner.ts
```diff
@@ -232,6 +236,14 @@ export function createEvmSigner(
     // Markr (proves networkFees was loaded). The bypass skips the
     // modal's fee picker, so missing-fees would broadcast as 0.
     const hasUpstreamFees = typeof tx.maxFeePerGas === 'bigint'
+    // Recurring synthetic Quotes have `serviceType === ServiceType.MARKR`
+    // (same as one-shot Markr swaps), so the existing serviceType gate alone
+    // would let recurring through. Exclude them explicitly — recurring TXs
+    // must always render the approval modal so the user sees the
+    // "Scheduling / Cancelling / Pausing / Unpausing recurring swap" preview.
+    const isRecurring = isRecurringAggregatorId(
```
meeh0w: What if instead of guessing by the aggregator id, we would add more "signature reasons" to the Fusion SDK?

It could then pass along things like:

```ts
stepDetails.currentSignatureReason = TransferSignatureReason.ScheduleRecurringSwap;
stepDetails.currentSignatureReason = TransferSignatureReason.PauseRecurringSwap;
stepDetails.currentSignatureReason = TransferSignatureReason.ResumeRecurringSwap;
stepDetails.currentSignatureReason = TransferSignatureReason.CancelRecurringSwap;
```

ruijia1in: good call, I am going to update the sdk with this change.  👍

ruijia1in: here is the new [PR](https://github.com/ava-labs/consumer-sdks/pull/259) for this change.

packages/core-mobile/app/vmModule/ApprovalController/validators/shared.ts
```diff
@@ -58,6 +62,123 @@ const readAutoApproveContext = (
...
+}
+
+export const readRecurringSwapApprovalContext = (
+  request: RpcRequest
+): RecurringSwapApprovalContext | undefined => {
+  const value = readCtx(request)?.[RequestContext.RECURRING_SWAP]
+  if (value === null || value === undefined || typeof value !== 'object') {
+    return undefined
+  }
+  const parsed = recurringSwapApprovalContextSchema.safeParse(value)
```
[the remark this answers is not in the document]
ruijia1in: fixed

packages/core-mobile/app/new/features/approval/components/RecurrenceDetails.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+  const ordersClause = context.isUnlimited
```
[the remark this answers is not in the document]
ruijia1in: wil be handled in next pr

packages/core-mobile/app/new/features/recurringSwap/utils/submitRecurringSwap.ts
```diff
@@ -0,0 +1,255 @@
...
+    sourceChain,
+    fromTokenSymbol,
+    fromTokenDecimals,
+    toTokenSymbol,
+    frequency,
+    numberOfOrders,
+    amountPerOrder
+  } = params
+  const isUnlimited = numberOfOrders === UNLIMITED_ORDERS
+
```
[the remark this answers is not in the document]
ruijia1in: fixed

packages/core-mobile/app/new/features/approval/screens/ApprovalScreen/ApprovalScreen.tsx
```diff
@@ -42,10 +44,40 @@ import {
...
+const ApprovalScreen = (props: {
+  params: ApprovalParams
+}): JSX.Element | null => {
+  const rejectAndClose = useCallback(
+    (message?: string) => {
+      props.params.onReject(message)
+      if (router.canGoBack()) router.back()
+    },
+    [props.params.onReject]
+  )
```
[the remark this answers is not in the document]
ruijia1in: fixed.

B0Y3R-AVA (dismissed):

atn4z7 (approved):

B0Y3R-AVA (approved):

PR #3916, opened 2026-06-24
CP-14507: fix C→P CCT swap on Ledger (empty xpAddressDictionary)

atn4z7:
## Description

**Ticket: [CP-14507](https://ava-labs.atlassian.net/browse/CP-14507)**

A C → P cross-chain transfer (CCT) swap failed on Ledger wallets. The handler logged `[cctCallbacks] xpAddressDictionary empty for active account`, so no X/P signing indices could be derived and the swap couldn't proceed.

**Root cause**

The CCT swap resolves the active account's X/P addresses through `transformXPAddresses`. Its dictionary branch only checked `if (queryData?.xpAddressDictionary)`. A successful address lookup can still return an empty `{}` (e.g. a Ledger account with no X/P activity yet, which is the normal state for the P-chain destination of a C → P swap). An empty object is truthy, so the code kept it and never fell back to the account's primary P-chain address. The `xpAddresses` branch right above it already falls back on an empty list, so the two branches were inconsistent. The empty dictionary made `getInternalExternalAddrs` return no signing indices, which surfaced the error above.

**Changes**

- `transformXPAddresses`: require a non-empty dictionary before using it; otherwise fall back to the account's primary P-chain address (`addressPVM`). This mirrors the existing `xpAddresses` branch and core-web's `createXpAddressDictionary`.
- `getCachedXPAddresses`: on a failed lookup, fall back via `transformXPAddresses(undefined, account)` instead of returning empty. Covers the Ledger Live case where the Avalanche xpub can't be derived (the lookup throws rather than returns empty).
- `getSwapErrorMessage`: map the `[cctCallbacks]` guard errors to a clear user-facing message instead of the raw internal string, as a safety net for the rare case an account genuinely has no P-chain address. The shared tag is extracted into a `CCT_CALLBACKS_ERROR_TAG` constant referenced by both the producer (`createCctCallbacks`) and consumer (`fusionErrors`).

## Screenshots/Videos
C->P CCT swap
https://github.com/user-attachments/assets/59a9e3b9-71bb-41ef-82e2-39e8191a176a

Error

## Testing
**iOS: 0.0.0.8963
Android: 0.0.0.8964**

1. Connect a Ledger wallet (via ledger live).
2. Start a swap from C-Chain to P-Chain (CCT).
3. Approve the export and import legs on the device.
4. Confirm the swap completes and the funds land on the P-chain (previously it failed with no progress).

## Checklist

[CP-14507]: https://ava-labs.atlassian.net/browse/CP-14507?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core-mobile/app/hooks/useXPAddresses/transformXPAddresses.test.ts
+++ b/packages/core-mobile/app/hooks/useXPAddresses/transformXPAddresses.test.ts
@@ -79,6 +79,33 @@ describe('transformXPAddresses', () => {
     })
   })
 
+  // CP-14507: a successful lookup can return an empty dictionary (e.g. a Ledger
+  // account with no X/P activity yet). The empty `{}` must NOT be used as-is —
+  // it has to fall back to the account's primary P-chain address, otherwise XP
+  // signing yields no indices and C->P CCT swaps fail.
+  describe('when queryData.xpAddressDictionary is an empty object', () => {
+    it('falls back to account.addressPVM for the dictionary', () => {
+      const queryData = {
+        xpAddresses: [],
+        xpAddressDictionary: {}
+      }
+      const account = createMockAccount({
+        addressPVM: 'P-avax1fallback'
+      })
+
+      const result = transformXPAddresses(queryData, account)
+
+      expect(result.xpAddresses).toEqual(['avax1fallback'])
+      expect(result.xpAddressDictionary).toEqual({
+        avax1fallback: {
+          space: 'e',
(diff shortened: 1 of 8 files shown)


B0Y3R-AVA (approved):

ruijia1in (approved):
dev-tested.

PR #3918, opened 2026-06-24
CP-14522: DCA polish — on-chain pause/cancel/resume via fusion-sdk 0.23.1

ruijia1in:
iOS: 8969
Android: 8970

## Description

**Ticket: [CP-14522](https://ava-labs.atlassian.net/browse/CP-14522)**

Builds on `dca-part1` to pick up the fusion-sdk release that signs and broadcasts recurring-swap cancel / pause / resume on-chain (replacing the prior HTTP-only Markr endpoints) and to ship the AC.1 / AC.2 polish under the DCA Swaps: UX polish ticket:

- **`@avalabs/fusion-sdk` 0.22.0 → 0.23.1**, the foundation for everything else in this PR.
- **Approval-modal context now rides on `step.signerContext`.** The Zustand-backed `activeActionContext` is gone — producers (`submitRecurringSwap`, `_makeOrderActionHook`) attach the payload through `TransferManager`'s signing config and `EvmSigner.signOne` reads it back via the new `readRecurringSignerContext` helper, gated on `TransferSignatureReason` being one of the four recurring values.
- **Order-action hooks unified.** `useCancel` / `usePause` / `useResume` collapse onto a single `_makeOrderActionHook` factory parameterised by the SDK enum + analytics event. `useUnpauseRecurringSchedule` is renamed to `useResumeRecurringSchedule` end-to-end (hook, test, analytics event, comments, store keys) so the code matches the existing CP-14399 UI copy.
- **Schedules screen lifted to a top-level modal** at `/recurringSwapSchedules`, deeplinkable as `core://recurringSwapSchedules`. Activity-tab banner + swap-modal row + future notifications all converge here.
- **Approval schema tightened.** `recurringSwapApprovalContextSchema` drops the `type` discriminator (fill vs order-action is now structural — fills carry `frequency`, order actions carry `action: TransferSignatureReason.*`) and the `isUnlimited` cross-field refinement (the `-1` wire sentinel is the only source of truth). `<RecurrenceDetails />` keys its copy table off `TransferSignatureReason` enum values directly.
- **NotificationCenterService**: log + degrade-to-generic on metadata parse failure instead of silently dropping the row.

**Dependencies:** `@avalabs/fusion-sdk@0.23.1` (already published).

**Not breaking** for callers outside `recurringSwap/` and `approval/` — the changes are internal to the recurring-swap surface.

## Screenshots/Videos

RecurringSchedulesScreen auto scroll to correct card and show recurring schedule not found error

https://github.com/user-attachments/assets/1f478db3-8c34-415f-b96a-33891da01529

https://github.com/user-attachments/assets/2a4927d2-c9f4-4745-86ef-fc35f081bff0

## Testing

**Dev Testing**

Schedule a recurring swap and verify the approval-modal preview reads correctly (token pair, per-order amount, frequency, total orders / Unlimited):

1. Pick AVAX → USDC, per-order = 1, frequency = 5 min, orders = 2. Approve. Schedule lands in `/recurringSwapSchedules`.
2. Repeat with orders = Unlimited; approval preview should say "Unlimited" and the schedule row should not display a total.

Pause / Resume / Cancel each fire an on-chain TX with the right modal copy:

3. From the schedules screen, Pause an Active row → approval modal title = "Pausing recurring swap", body mentions allowance preserved + "may still execute one more fill". Tap Approve → row flips to Paused after the TX indexes.
4. From the same screen, Resume the Paused row → title = "Resuming recurring swap" → row flips back to Active.
5. Cancel from Active and Cancel-from-Paused both land on "Cancelling recurring swap" → row flips to Cancelled. The pending spinner does not flicker back to Pause/Resume in the cancel-from-Paused case.

Top-level modal entry points all open the same screen:

6. From Activity tab, tap the "Recurring schedules" banner → opens `/recurringSwapSchedules`.
7. From the swap modal, tap the inline schedules row → same destination.
8. Open `core://recurringSwapSchedules` via a deeplink (terminal: `xcrun simctl openurl booted core://recurringSwapSchedules`) → opens the modal directly.

**QA Testing**

Acceptance criteria from CP-14522:
- **AC.1** Contextual approval messages: cancel / pause / resume each show their own title + body in the approval modal preview, with token symbols injected.
- **AC.2** Pause + Resume work end-to-end on-chain (no HTTP fallback). A paused schedule stops executing fills; resuming it picks back up without re-approval.

## Checklist

[CP-14522]: https://ava-labs.atlassian.net/browse/CP-14522?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/core-mobile/app/new/features/swap/services/signers/EvmSigner.ts
```diff
@@ -244,7 +241,9 @@ export function createEvmSigner(
     // would let recurring through. Exclude them explicitly — recurring TXs
     // must always render the approval modal so the user sees the
     // "Scheduling / Cancelling / Pausing / Unpausing recurring swap" preview.
```
atn4z7: Two stale comments here now that detection moved to `currentSignatureReason`:

- line 243 still lists the preview as "Scheduling / Cancelling / Pausing / **Unpausing** recurring swap" (should be "Resuming" to match the rename, and `store/rpc/types.ts` which already says Resuming).
- the comment just below (~285) still says detection keys on "`aggregator.id` starts with `markr-recurring`", but the code reads `readRecurringSignerContext(stepDetails)` off the signature reason now, not the aggregator id.

packages/core-mobile/app/new/features/recurringSwap/store/listeners.ts
```diff
@@ -117,39 +120,57 @@ function processAllSchedules(
...
  * the next listOrders refetch lands (the order was already Paused before
  * the cancel TX was even broadcast), causing a UI flicker back to
- * Pause/Unpause until the Cancelled status finally indexes.
+ * Pause/Resume until the Cancelled status finally indexes.
  */
 function isPendingActionResolved(
   entry: PendingActionEntry,
   order: RecurringOrder
 ): boolean {
   switch (entry.type) {
```
atn4z7, line 135: This `switch` is exhaustive over the 3 action types today, but there is no `default` / `assertNever(entry.type)`. If a 4th `RecurringOrderActionType` is ever added, this silently returns `undefined` (falsy, so the spinner just waits for the TTL) instead of failing the build. An `assertNever` default would turn that into a compile error. Low impact given the TTL backstop.

packages/core-mobile/app/new/features/notifications/services/NotificationCenterService.ts
```diff
@@ -34,6 +34,12 @@ function transformNotification(
   switch (response.type) {
```
atn4z7, line 34: Same note as the reconciler switch: `transformNotification`'s `switch (response.type)` has no `default`. It is exhaustive over the validated zod enum so it is unreachable today, but a future enum member would fall through to an implicit `undefined` return. A `default` with `assertNever` would harden it.

atn4z7 (commented):
Nice PR.

packages/core-mobile/app/new/features/notifications/services/NotificationCenterService.ts
```diff
@@ -34,6 +34,12 @@ function transformNotification(
...
     case 'BALANCE_CHANGES': {
       const parsed = BalanceChangesMetadataSchema.safeParse(response.metadata)
+      if (!parsed.success) {
+        Logger.warn(
+          '[NotificationCenterService] BALANCE_CHANGES metadata parse failed; falling back to generic row',
+          parsed.error
+        )
+      }
```
atn4z7, line 42: shouldn't this be Logger.error so we can see it on sentry?

packages/core-mobile/app/new/features/recurringSwap/hooks/_makeOrderActionHook.ts
```diff
@@ -88,13 +113,69 @@ type OrderActionConfig = {
...
+        `[${hookName}] action TX ${txHash} reverted on chain ${chainId}; cleared pending entry for order ${orderId}`
+      )
+    }
+  } catch (err) {
+    // Timeout / RPC error — don't clear; the reconciler + TTL remain the
+    // backstop in case a confirmation lands later.
+    Logger.warn(
+      `[${hookName}] action receipt watch did not confirm for order ${orderId}; leaving pending for reconciler/TTL`,
+      err
+    )
```
atn4z7: same here for logger.error

packages/core-mobile/app/new/features/recurringSwap/hooks/_makeOrderActionHook.ts
```diff
@@ -109,6 +190,13 @@ export function makeOrderActionHook(
     // the first to settle would flip `isPending` off mid-flight on the
     // second. Drop calls while one is already in flight.
     const inFlightRef = useRef(false)
+    // The receipt watcher needs the source network for `args.chainId`, which
+    // is only known per-call. Hold the live networks map in a ref so the
+    // empty-deps `run` reads the latest without churning its identity (which
+    // the SchedulesScreen's stable `mutate` reference depends on).
+    const { networks } = useNetworks()
+    const networksRef = useRef(networks)
+    networksRef.current = networks
```
atn4z7: **NIT** — this writes to the ref during render. React discourages mutating refs in the render body (breaks render purity, and under concurrent rendering a discarded render would still mutate it). The documented form is to write it in an effect:

```js
useEffect(() => {
  networksRef.current = networks
}, [networks])
```

There's already a `useEffect` right below, so it's a small move. No behavioral impact here since `run` only reads `networksRef.current` post-commit on a user tap, so purely a future-safety nit.

packages/core-mobile/app/new/features/recurringSwap/types.ts
```diff
@@ -62,7 +62,8 @@ export function resolveRecurringTokenAddress(
   token: LocalTokenWithBalance
 ): string | null {
   if (token.type === TokenType.NATIVE) return ERC_ZERO_ADDRESS
-  if ('address' in token && token.address) return token.address
+  if (token.type === TokenType.ERC20 && 'address' in token && token.address)
+    return token.address
```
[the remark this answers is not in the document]
ruijia1in: Good catch. Aligned `isUsable` to delegate to `resolveRecurringTokenAddress` so the gate and the address resolution can't drift — an SPL token (non-empty Solana address) is now rejected up front instead of resolving to `''`. Also dropped the `?? ''` coercion in the queryFn in favor of the same invariant throw used by the sibling guards, so an empty address can never reach the SDK. Added a regression test (SPL token with a non-empty address does not fire the query). Fixed in c70e221bd.

packages/core-mobile/app/new/features/swap/services/signers/EvmSigner.ts
ruijia1in: Updated both — "Unpausing" → "Resuming" on line 243, and the detection note now describes `currentSignatureReason` (the `isRecurring` check) instead of the old "aggregator.id starts with markr-recurring" wording. c70e221bd.

packages/core-mobile/app/new/features/recurringSwap/store/listeners.ts
ruijia1in: Added an `assertNever` default — a 4th `RecurringOrderActionType` is now a compile error instead of a silent `undefined` return that would hang the spinner until the TTL. The throw is the runtime backstop. c70e221bd.

packages/core-mobile/app/new/features/notifications/services/NotificationCenterService.ts
ruijia1in: Added the same `assertNever` default. Unreachable today (the `response.type` is zod-validated against the enum upstream in `fetchNotifications`), and the throw is caught there (returns `[]`), which is safer than the previous implicit `undefined` leaking into the list. c70e221bd.

packages/core-mobile/app/new/features/recurringSwap/hooks/_makeOrderActionHook.ts
ruijia1in: Moved the `networksRef.current = networks` write into `useEffect(..., [networks])`. `useRef(networks)` seeds the first render and `run` only reads the ref post-commit (on a user tap), so there's no behavior change — purely the render-purity / concurrent-safety fix. c70e221bd.

packages/core-mobile/app/new/features/swap/screens/SwapScreen.tsx
```diff
@@ -965,7 +975,8 @@ export const SwapScreen = (): JSX.Element => {
     isRecurringBlocked,
     eligibility.eligible,
     recurring.isRecurring,
-    recurring.setIsRecurring
+    recurring.setIsRecurring,
+    recurring
```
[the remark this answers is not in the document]
ruijia1in: Dropped `recurring` from the effect deps and key on the stable `recurring.isRecurring` / `recurring.setIsRecurring` fields (with the `react-hooks/exhaustive-deps` suppression pattern used elsewhere in this file), so it no longer re-runs on the unstable context object identity. c70e221bd.

packages/core-mobile/app/new/features/notifications/services/NotificationCenterService.ts
ruijia1in: Done — this (and the PRICE_ALERTS / NEWS parse-failure logs) are now `Logger.error` so backend metadata/schema drift surfaces in Sentry.

packages/core-mobile/app/new/features/recurringSwap/hooks/_makeOrderActionHook.ts
ruijia1in: Done — both the on-chain-revert log and the receipt-watch-failure log are now `Logger.error` so they surface in Sentry.

packages/core-mobile/app/new/features/recurringSwap/hooks/usePauseRecurringSchedule.test.ts
```diff
@@ -189,7 +202,11 @@ describe('usePauseRecurringSchedule', () => {
     expect(mockSnackbar).not.toHaveBeenCalled()
   })
 
-  it('shows "Try again" on a generic signer/network failure (not a user rejection)', async () => {
+  // A generic (non-HTTP, non-rejection) failure thrown out of `executePause`
+  // is ambiguous — the TX may already have broadcast. The shared hook keeps the
+  // order pending (blocking a double-submit) and suppresses the toast rather
+  // than showing "Try again" and re-enabling the button.
+  it('keeps the order pending without a toast on a generic (ambiguous) failure', async () => {
```
B0Y3R-AVA, line 255: The revert path (status 0 clears the entry and fires the snackbar) only gets exercised through the cancel hook test. Since the watcher lives in the shared builder I get why, but it means the pause and resume `reverted` strings never get asserted by anything. Swap one of those copy lines by accident and no test notices. A small case here that mocks a status-0 receipt and checks the pause snackbar fires would cover it.

packages/core-mobile/app/new/features/swap/screens/SwapScreen.tsx
```diff
@@ -1182,6 +1193,18 @@ export const SwapScreen = (): JSX.Element => {
...
+  // Mirror the one-shot `canSwap` gate: a blocking (non-warning) validation
+  // error must disable Next. Most importantly this covers below-minimum, which
+  // `computeValidationError` evaluates against the recurring per-token minimum
+  // (`effectiveMinimumTransferAmount`). Without it the below-minimum error was
+  // displayed but never reached `canSubmit`, and because the recurring quote
+  // succeeds below the minimum (`useRecurringQuote` only gates on a non-zero
+  // amount, and `/recurring/quote` doesn't enforce the per-order minimum), a
+  // sub-minimum `amountPerOrder` could be submitted. Warnings (e.g.
+  // gas-estimation) are tolerated, matching `canSwap`'s `isWarning` allowance.
+  const hasBlockingValidationError =
```
B0Y3R-AVA: This is the real fix in the PR, the thing keeping a sub-minimum amountPerOrder from going through, and it's the part with no test. The recurring quote happily succeeds below the minimum, so this gate is the only thing in the way. I'd want a test pinning it down: blocking validation error present means canSubmit is false, warning-only still submits. Easy to regress otherwise.

packages/core-mobile/app/new/features/recurringSwap/screens/RecurringSchedulesScreen.tsx
```diff
@@ -566,6 +596,57 @@ export function RecurringSchedulesScreen(): JSX.Element {
...
+    [initialExpandedOrderId, headerHeight]
+  )
+
+  // Deep-linked orderId may not show up in the manageable list — the order
+  // could be `cancelled` / `completed` (filtered out), belong to a different
+  // account, or not exist on the active chain. Surface a one-shot snackbar
+  // once the fetch settles so the user isn't left wondering why nothing
+  // expanded. Only fires when the fetch succeeded (an error already renders
+  // the Retry CTA inline).
+  const notFoundShownRef = useRef(false)
```
B0Y3R-AVA, line 638: None of the deep-link behavior is covered yet: the auto-expand from initialExpandedOrderId, the scroll-into-view in handleCardMeasureY, and this not-found snackbar. The scroll one leans on rAF plus a 250ms timeout plus a forwarded ref, which is the kind of thing that quietly stops working and nobody catches until a notification tap lands in the wrong spot. At a minimum I'd test that the not-found snackbar fires once when the orderId isn't in the list.

packages/core-mobile/app/new/features/recurringSwap/store/pendingActionStore.ts
```diff
@@ -33,22 +46,86 @@ export interface PendingActionState {
...
+
+function isPersistedEntry(value: unknown): value is PendingActionEntry {
+  if (!value || typeof value !== 'object') return false
+  const e = value as Partial<PendingActionEntry>
+  // Validate every field the listener-side reconciler reads. An entry
+  // missing any of these would either fail the (account, chain) scope
+  // check (and stick around until the TTL fired) or crash the
+  // reconciler on the next list refetch.
+  return (
+    typeof e.type === 'string' &&
```
B0Y3R-AVA: Tiny one. `type` is only checked for typeof string here, not against the three real action values. A corrupted entry with some junk string type survives the migration, then never resolves in isPendingActionResolved (no default branch) and just sits there until the TTL evicts it. The TTL saves you so it's not a real bug, but the enum is already imported, so an `includes` check would close it for free.

packages/core-mobile/app/new/features/swap/screens/SwapScreen.tsx
```diff
@@ -1275,12 +1299,13 @@ export const SwapScreen = (): JSX.Element => {
     recurring.numberOfOrders,
     fromTokenValue,
     getNetwork,
+    recurringSlippageBps,
     dismissAll
   ])
 
-  const handleNext = useCallback(() => {
+  const handleNext = useCallback(async () => {
```
B0Y3R-AVA, line 1306: handleNext went async for the recurring await, but handleSwap (the normal path) is still called without one. If it ever throws synchronously that's now an unhandled rejection inside the async wrapper instead of bubbling up the way it used to. Probably moot since onPress doesn't catch either way, just noting the behavior shifted.

B0Y3R-AVA (commented):
Did a careful pass, mostly on the signerContext refactor and the pending-action plumbing. The SDK side checks out. I pulled the 0.23.1 tarball and the recurring TransferSignatureReason members and signerContext are really in there, so the gating is sound. Moving the display metadata onto the request itself instead of the old activeActionContext map is a clear win, that map's whole reason for existing was the concurrent same-type stomp and this just deletes the problem.

Nothing here blocks merge. What stood out was test coverage on a few of the new bits, plus a couple of tiny things. Left them inline. The nits are take-it-or-leave-it.

packages/core-mobile/app/new/features/recurringSwap/hooks/usePauseRecurringSchedule.test.ts
ruijia1in: Added. Both `usePauseRecurringSchedule.test.ts` and `useResumeRecurringSchedule.test.ts` now have a status-0 revert case that mocks a resolvable network + a reverted receipt and asserts `clearPending` fires and the action-specific snackbar shows (`Pause failed on-chain. Please try again` / `Resume failed on-chain. Please try again`). Mirrors the existing cancel-hook revert test, so a copy swap on any of the three would now be caught. be87029a4.

packages/core-mobile/app/new/features/swap/screens/SwapScreen.tsx
ruijia1in: Agreed this was the riskiest gap. Extracted the gate into `swap/utils/recurringSubmitGate.ts` (`hasBlockingValidationError` / `isRecurringReady` / `computeCanSubmit`) — a behavior-preserving extraction so it's unit testable without standing up the whole screen — and added `recurringSubmitGate.test.ts`. It pins down exactly what you described: a blocking (non-warning) validationError ⇒ `canSubmit` false, warning-only ⇒ still submits, plus each presence guard and the one-time path deferring to `canSwap`. be87029a4.

packages/core-mobile/app/new/features/recurringSwap/screens/RecurringSchedulesScreen.tsx
ruijia1in: Covered the decision logic: extracted `shouldShowDeepLinkNotFound` into `recurringSwap/utils/shouldShowDeepLinkNotFound.ts` and unit-tested it — absent id after a successful fetch fires (once, via the `alreadyShown` guard), and loading / error / already-shown / id-present all suppress. I left the rAF + 250ms scroll-into-view itself out of scope here: it leans on real timers + a forwarded ref and a full render harness for this screen doesn't exist yet, so a test there would be brittle. The pure predicate carries the not-found coverage you asked for. be87029a4.

packages/core-mobile/app/new/features/recurringSwap/store/pendingActionStore.ts
ruijia1in: Closed it. Added `isRecurringOrderActionType` (backed by a runtime array of the three enum values) and `isPersistedEntry` now validates `type` through it instead of `typeof string`, so a junk-string `type` is dropped during migration rather than sitting until the TTL. Added a migration test for junk/empty `type`. be87029a4.

packages/core-mobile/app/new/features/swap/screens/SwapScreen.tsx
ruijia1in: Left as-is — agree it's moot. `submitRecurring` fully encapsulates its own error handling (try/catch/finally, swallows user-rejection, toasts the rest), so nothing rejects out of the recurring branch. And `handleSwap` is the pre-existing one-time path that already manages its own errors. `onPress` doesn't await or catch either way, so the sync-throw → async-reject shift you noted is behavior-neutral in practice. Happy to wrap it defensively if you'd prefer, but it didn't seem worth touching the swap path.

atn4z7 (approved):

B0Y3R-AVA (approved):

PR #3922, opened 2026-06-25
CP-14511, CP-14512, CP-14516: CCT swap fixes — P/X balance, Seedless two-tx banner, P/X Swap button

atn4z7:
## Description

**Tickets: [CP-14511](https://ava-labs.atlassian.net/browse/CP-14511), [CP-14512](https://ava-labs.atlassian.net/browse/CP-14512), [CP-14516](https://ava-labs.atlassian.net/browse/CP-14516)**

Three CCT-related swap fixes, now that CCT supports swapping native AVAX across C/P/X:

- **CP-14511** — P/X AVAX balance showed `0` on the swap destination ("to") token screen. The merge keyed native AVAX on `NATIVE-AVAX`, but P/X balances are stored under `AVAX-P`/`AVAX-X`, so the lookup missed.
- **CP-14512** — the "two transactions" banner was hidden for Seedless (copied from core-web). On mobile every CCT leg goes through the in-app approval pipeline, so Seedless users do sign twice and need the warning.
- **CP-14516** — the Swap button was missing on the P/X AVAX token detail screen; it was hard-disabled for X/P before CCT existed.

## Screenshots/Videos
P/X AVAX balance
https://github.com/user-attachments/assets/3fd458b5-66e1-4a1b-8e9d-a126a301a844

P/X Swap button
https://github.com/user-attachments/assets/2c40b6de-27db-4258-b1b6-beb7e5a341ee

## Testing
**iOS: 0.0.0.8986
Android: 0.0.0.8987**

_CP-14511 — P/X balance on the swap destination screen_
- Open Swap, pick any "from" token, open the destination ("to") token picker, switch the network tab to **P-Chain** → AVAX shows your real P-Chain balance (not 0). Repeat for **X-Chain**.
- Confirm C-Chain AVAX still shows its balance (regression check).

_CP-14512 — Seedless two-tx banner_
- On a **Seedless** wallet, set up a CCT swap (e.g. P-Chain AVAX → C-Chain AVAX) → the "This swap will require signing two transactions" banner is shown.
- On a non-Seedless wallet, the banner still shows for the same CCT swap.

_CP-14516 — Swap button on P/X AVAX detail_
- Open the **P-Chain AVAX** token detail → a **Swap** button appears; tapping it lands on Swap with from = P-Chain AVAX and destination preselected to **C-Chain AVAX**. Repeat for **X-Chain AVAX**.
- A non-AVAX X-Chain asset (ANT) shows **no** Swap button.
- With the CCT feature flag off, or on a wallet without X/P addresses, the Swap button does **not** appear on P/X AVAX.
- On **testnet (Fuji)**, confirm the destination preselects Fuji C-Chain AVAX (not empty).

## Checklist

[CP-14511]: https://ava-labs.atlassian.net/browse/CP-14511?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CP-14512]: https://ava-labs.atlassian.net/browse/CP-14512?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CP-14516]: https://ava-labs.atlassian.net/browse/CP-14516?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/core-mobile/app/vmModule/handlers/avalancheSendTransaction.ts
```diff
@@ -118,6 +119,15 @@ export const avalancheSendTransaction = async ({
       signedData: signedTransactionHex as Hex
     })
   } catch (error) {
+    Logger.error('[avalancheSendTransaction] sign failed', {
```
B0Y3R-AVA, line 124: One thing on this new log: `Logger.error` ends up at `SentryService.captureException` in prod, and right now we're handing it the whole raw `error` plus `error.stack`. Depending on what the signer throws, that can carry the unsigned tx payload off-device, and we're also sending the derivation indices alongside it. The indices aren't secrets on their own, but I'd rather not ship either to Sentry by default.

Can we trim this to `error.message` and drop the raw `error`/`stack`? Or gate the full dump behind dev mode if we want it locally. The intent (surfacing the swallowed cause) is good, just want to narrow what leaves the device.

packages/core-mobile/app/new/features/swap/hooks/useFusionTokenLookup.ts
```diff
@@ -191,11 +203,8 @@ export function useFusionTokenLookup({
         : undefined
     )
 
-    // In testnet (developer mode), skip preselecting a "to" token.
-    // Initial to-token IDs (e.g. USDC) are mainnet-specific and no services
-    // support them in testnet, which would lead to a broken no-quotes state.
     setToToken(
-      initialTokenIdTo && !isDeveloperMode
+      shouldPreselectToToken(initialTokenIdTo, isDeveloperMode)
```
B0Y3R-AVA, line 207: This is the path I'd want eyes on before merge. We now let `NATIVE-avax` through the testnet preselect, but `resolveInitialToken` still returns `undefined` when the aggregator has no Fuji `NATIVE-avax` entry, and then we call `setToToken(undefined)`. If that lookup comes back empty the destination just stays blank, which is the opposite of what the P/X to C deep-link is trying to do.

Your own testnet step says Fuji should preselect C-Chain AVAX, so it'd be good to confirm that actually happens on a real Fuji run rather than silently falling back to empty.

packages/core-mobile/app/new/features/swap/hooks/useSwapTokens.ts
```diff
@@ -58,10 +60,19 @@ const toFusionSourceAsset = (
...
+// `AVAX-P` / `AVAX-X` ids (see BalanceService), not `NATIVE-AVAX`, so the
+// merge would otherwise miss and the balance would render as 0.
+const getBalanceLookupId = (
+  asset: BridgeableUiAsset,
+  targetChainId: number
+): string => {
+  if (asset.type !== FusionTokenType.NATIVE) return asset.address.toLowerCase()
+  if (isPChain(targetChainId)) return AVAX_P_ID
+  if (isXChain(targetChainId)) return AVAX_X_ID
+  return `NATIVE-${asset.symbol}`
```
B0Y3R-AVA, line 79: Minor, not a blocker. `getBalanceLookupId` returns `AVAX_P_ID`/`AVAX_X_ID` for any native asset on a P/X target, keyed on the chain rather than the symbol. Fine today since P/X only expose native AVAX, but if the Fusion "to" list ever surfaces a non-AVAX native asset on those chains, its balance would get the AVAX balance merged onto it. A quick `asset.symbol === 'AVAX'` guard would keep it honest if that ever changes.

packages/core-mobile/app/vmModule/handlers/avalancheSendTransaction.ts
atn4z7: done

packages/core-mobile/app/new/features/swap/hooks/useFusionTokenLookup.ts
atn4z7: I believe I already captured this case in one of the 2 videos in the pr. Lmk if the video doesn't cover the case you had in mind.

packages/core-mobile/app/new/features/swap/hooks/useSwapTokens.ts
atn4z7: added the guard

B0Y3R-AVA (approved):

onghwan (approved):
dev-tested

PR #3928, opened 2026-06-29
Recurring swap: execution push notifications, Markr per-order floor, first-fill double-approval fix

ruijia1in:
iOS: 9057
Android: 9058

## Description

**Ticket: [CP-14581]** **Ticket: [CP-14582]** **Ticket: [CP-14584]**

Adds push notifications for recurring swap (DCA) schedule execution, re-introduces a Markr-specific per-order minimum, fixes a first-fill double-approval bug, adds a pre-submit balance guard for under-funded schedules, hardens the per-order-minimum submit gate against a loading-race, polishes the manage-screen loading state, and upgrades the fusion-sdk to its `0.25.0` release.

**Summary of changes**

- **Execution push notifications (core-notification-sender integration)**
  - `recurringSwapNotifications`: subscribes a device to per-order push via `/v1/push/recurring-swaps/subscribe` (AppCheck-authed). Re-subscribe is an idempotent reactivation, so retries are safe; teardown is webhook/event-driven server-side.
  - `subscribedOrders`: per-`(owner, chain)` MMKV set of already-subscribed order IDs so the `listOrders` watcher skips redundant subscribe calls (bounded to 2000 entries, evicted oldest-first).
  - `recurringSwap/store/listeners`: auto-subscribe watcher over `listOrders`.
  - Notifications feature: new `RecurringSwapItem` row + `NotificationIcon` / `NotificationsScreen` / `NotificationCenterService` / `schemas` / `types` / `utils` to parse and route the recurring-swap notification type (title/body formatted server-side; `data.status` drives the status badge).
  - FCM (`FCMService`, `types`): handle the recurring-swap remote-message type.

- **Notification row polish + subscribe efficiency**
  - `RecurringSwapItem` reads the structured `data` (status + order counts), not status alone: a mid-schedule fill renders a green **"Executed"** badge while the final leg renders **"Completed"**; failures render red **"Failed"**. Infinite/DCA schedules (`numberOfOrders === -1`) always read as a mid-schedule fill.
  - `isTerminalRecurringSwapNotification` + `hasActionableUrl`: terminal / last-leg recurring rows point at a schedule the manage screen no longer lists (Active / Paused only), so they're non-actionable (no chevron, press no-ops).
  - `RecurringSwapMetadataSchema.reasonCode` → `z.coerce.string()`: the FCM push stringifies every `data` value while the notification-center history API returns raw JSON. A failed notification always carries a `reasonCode`, so the non-coercing schema dropped the whole `data` block on the history path (breaking the failure badge + terminal detection).
  - `ensureOrderSubscriptions` resolves the device ARN (`/v1/push/register`) once per batch and memoizes it across batches (skipping register in the steady state); cache invalidated on subscribe failure, `onFcmTokenChange`, and `onLogOut`.

- **Per-order minimum + submit-gate race**
  - `useMinimumTransferAmount` gains an optional `serviceType` to pin a single service's floor instead of the blended "lowest across services" value.
  - `SwapScreen`: recurring validates the per-order amount against **Markr's** floor specifically (recurring is Markr-only) — blending in a lower non-Markr floor could let a sub-fillable per-order amount through.
  - **Race fix:** `computeValidationError` skips the below-minimum check while the minimum is unknown, and `isRecurringReady` only required a resolved recurring quote — so a quick tap in the window where the quote resolved but the Markr minimum query was still loading could submit a sub-minimum amount. Added `isRecurringMinimumReady` to the submit gate (`markrMinimumTransferAmount !== undefined`); Next stays disabled until the minimum query settles. No deadlock — if Fusion isn't ready the recurring quote can't resolve either, so `hasRecurringQuote` already gates Next.

- **Under-funded schedule guard (pre-submit balance validation)**
  - A native-source recurring swap surfaced a generic "Recurring swap failed — please try again" toast on submit. Root cause: the first fill pre-wraps the **entire** `totalAmountIn` (per-order × order count) to WAVAX in one tx (the schedule can't pull native per order), so when that total exceeds the native balance, `eth_estimateGas` reverts with `INSUFFICIENT_FUNDS` and the EVM module throws `rpcErrors.internal`. The app validated only the per-order amount against balance, never the total.
  - New tested `recurringBalanceGate` (`computeRecurringBalanceError`), wired into `computeValidationError` so it renders inline **and** disables Next via `isRecurringReady`:
    - **Native source:** require `nativeBalance ≥ totalAmountIn + additive native fees` (the SDK flags the one-time schedule fee with `fee.extra` as "balance-check separately") — one combined message quoting the full requirement.
    - **ERC-20 source:** principal checked against the token balance, the native schedule fee against the native balance — independent shortfalls surfaced together when both fall short ("needs X USDC and Y AVAX for the schedule fee").
    - Uses the authoritative `recurringQuote.totalAmountIn` + summed `extra` fees. Estimated **gas** is left to the SDK's own `estimateGas` (the quote reports it in gas units, not a reservable wei cost). Skipped for **Unlimited** (no finite total).
  - `fusionErrors`: new `recurringTotalExceedsBalance` / `recurringScheduleFeeExceedsNativeBalance` / `recurringInsufficientForTotalAndFee`.

- **Manage screen — loading state**
  - `RecurringSchedulesScreen`: while the first `listOrders` fetch is in flight, show only a centered spinner over the modal instead of flashing the large "0 recurring swaps scheduled" header + "Loading…" caption (the count is unknown until data resolves). Spinner color uses the `$textPrimary` theme token (light/dark aware) rather than a hardcoded hex.

- **First-fill double-approval fix**
  - `submitRecurringSwap`: pass `fallbackToDefaultOnBatchFailure: false`. Mobile's `signBatch` is non-atomic for recurring (it signs/broadcasts wrap → approve → swap sequentially), so the SDK's batch-failure fallback silently re-ran the whole sequence on any throw — producing a **second spend-limit approval** and a spurious **"User rejected the request."** from a superseded approval sheet. `false` surfaces a real failure once, with no double-execution. Ledger/WC peers are unaffected (recurring never issues a true atomic batch).

**Dependencies**

- Bumps `@avalabs/fusion-sdk` to `0.25.0` (recurring native-input release).

## Screenshots/Videos

https://github.com/user-attachments/assets/7dda133d-9c00-401e-9b46-f3573ed1b0b6

https://github.com/user-attachments/assets/d05942b8-b5a6-465c-914b-420c7a742241

## Testing

**Dev Testing**

Happy path
- Create a recurring swap (e.g. AVAX → USDC), set frequency + number of orders, tap Next and confirm.
- Confirm you are prompted for the spend-limit approval **exactly once** (then the swap approval), and the schedule is created — no duplicate spend-limit prompt.
- Confirm no "Recurring swap failed — please try again" toast appears on a successful schedule.
- When an order executes (or fails) on-chain, confirm a push notification arrives; tapping it opens the notification center showing the recurring-swap row with the correct status badge (Executed for a mid-schedule fill, Completed for the final leg, Failed for a failure). Confirm a terminal/last-leg row has no chevron and doesn't navigate.

Per-order minimum + gate race
- Enter a per-order amount below Markr's minimum for the pair → "Next" is disabled and the below-minimum error shows.
- Enter an amount at/above the minimum → "Next" enables.
- Immediately after configuring frequency + orders, confirm "Next" does not enable until pricing/the minimum has loaded — i.e. you can't tap through before the per-order floor is known.

Under-funded schedule (balance guard)
- Native source: with AVAX selected and a **total** (per-order × orders) greater than your AVAX balance, confirm "Next" is disabled and an inline "Insufficient balance — N orders require X AVAX" message shows — instead of a "Recurring swap failed" toast appearing only after you submit.
- Lower the order count or per-order amount so the total (plus the schedule fee) fits your balance → the message clears and "Next" enables.
- ERC-20 source where the token covers the principal but you lack the native gas-token for the schedule fee → confirm the "needs … for the schedule fee" message; if the token is also short, confirm the combined "needs X and Y for the schedule fee" message.
- "Unlimited" orders (where offered) → confirm the total-balance message does not appear (no finite total to check).

Manage screen
- Open the recurring-swap management modal while schedules are still loading → confirm only a centered spinner shows (no "0 recurring swaps scheduled" header flash, no "Loading…" text). Once loaded, the header + schedule cards (or the empty / error state) render normally.

**QA Testing**

- Verify the recurring swap schedule + first fill complete with a single spend-limit approval across software wallets (mnemonic, seedless).
- Verify execution/failure push notifications render correctly (Executed / Completed / Failed) and route into the notification center; terminal rows are non-actionable.
- Verify a recurring swap whose total exceeds balance is blocked at input time with a clear message (native and ERC-20 sources), and that a sufficiently-funded schedule proceeds normally.
- Verify the manage modal shows a centered spinner while loading and never flashes a "0 recurring swaps scheduled" header.
- Acceptance: scheduling a recurring swap never shows a duplicate spend-limit approval, never reports a rejection the user did not make, never fails on-chain for an under-funded total that could have been caught up front, and never lets a sub-minimum per-order amount through.

## Checklist

[CP-14581]: https://ava-labs.atlassian.net/browse/CP-14581?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CP-14582]: https://ava-labs.atlassian.net/browse/CP-14582?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CP-14584]: https://ava-labs.atlassian.net/browse/CP-14584?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core-mobile/app/new/features/notifications/components/RecurringSwapItem.tsx
+++ b/packages/core-mobile/app/new/features/notifications/components/RecurringSwapItem.tsx
@@ -1,6 +1,10 @@
 import React, { FC, useMemo } from 'react'
 import { Text, useTheme, View } from '@avalabs/k2-alpine'
-import { AppNotification, isRecurringSwapNotification } from '../types'
+import {
+  AppNotification,
+  isRecurringSwapNotification,
+  RecurringSwapMetadata
+} from '../types'
 import NotificationListItem from './NotificationListItem'
 import NotificationIcon from './NotificationIcon'
 
@@ -14,27 +18,39 @@ type RecurringSwapItemProps = {
 // Per Sarp's PR #174 the notification-sender service formats the push
 // envelope's `title` + `body` server-side (e.g. "Recurring swap executed",
 // "Insufficient balance — your recurring swap schedule will be cancelled"),
-// so the row just renders those verbatim. The `data.status` field is used
-// only to drive the colored status badge under the subtitle — that's the
-// one piece of UI we can synthesize client-side without lossy formatting.
+// so the row just renders those verbatim. The machine-readable `data` block
+// drives the colored status badge under the subtitle — the one piece of UI we
+// synthesize client-side without lossy formatting.
 //
-// Status → badge mapping (matches the Figma design):
-//   - 'failed'                → red "Failed"
-//   - 'active' / 'completed'  → green "Completed"   (a fill landed; either a
-//                                                    mid-schedule leg or the
-//                                                    final one — both are
-//                                                    surfaced the same way)
-//   - anything else           → no badge
-type BadgeKind = 'success' | 'failure' | null
+// We read the structured `data` fields (NOT the human-facing copy) to tell a
+// mid-schedule fill apart from the final leg:
+//   - 'failed'                       → red "Failed"
+//   - final leg                      → green "Completed"
+//       (`status === 'completed'`, or — for finite schedules — no fills left)
+//   - other success fill             → green "Executed"
+//       (a mid-schedule leg landed; infinite/DCA schedules never "complete",
+//        so they always read here)
+//   - anything else                  → no badge
+type Badge = { kind: 'success' | 'failure'; label: string } | null
 
-function resolveBadgeKind(status: string | undefined): BadgeKind {
-  if (status === undefined) return null
-  const lower = status.toLowerCase()
-  if (lower === 'failed') return 'failure'
-  if (lower === 'active' || lower === 'completed' || lower === 'executed') {
-    return 'success'
-  }
-  return null
+function resolveBadge(data: RecurringSwapMetadata | undefined): Badge {
+  if (data === undefined) return null
+  const status = data.status.toLowerCase()
+  if (status === 'failed') return { kind: 'failure', label: 'Failed' }
+
+  const isSuccessFill =
+    status === 'active' || status === 'completed' || status === 'executed'
+  if (!isSuccessFill) return null
+
+  // The schedule is finished when the backend marks it completed, or — for
+  // finite schedules — when no fills remain. Infinite/DCA schedules
+  // (numberOfOrders === -1) never reach a completed terminal, so they always
+  // read as a mid-schedule fill.
+  const isFinalLeg =
+    status === 'completed' ||
+    (data.numberOfOrders !== -1 && data.remainingOrders === 0)
+
+  return { kind: 'success', label: isFinalLeg ? 'Completed' : 'Executed' }
 }
 
 const RecurringSwapItem: FC<RecurringSwapItemProps> = ({
@@ -47,10 +63,10 @@ const RecurringSwapItem: FC<RecurringSwapItemProps> = ({
     theme: { colors }
   } = useTheme()
 
-  const badgeKind = useMemo(
+  const badge = useMemo(
     () =>
       isRecurringSwapNotification(notification)
-        ? resolveBadgeKind(notification.data?.status)
+        ? resolveBadge(notification.data)
         : null,
     [notification]
   )
@@ -59,10 +75,9 @@ const RecurringSwapItem: FC<RecurringSwapItemProps> = ({
   // in a View (rather than passing a string) lets NotificationListItem render
   // both lines without us re-implementing its typography defaults.
   const subtitle = useMemo<React.ReactNode>(() => {
-    if (badgeKind === null) return notification.body
+    if (badge === null) return notification.body
     const badgeColor =
-      badgeKind === 'failure' ? colors.$textDanger : colors.$textSuccess
-    const badgeLabel = badgeKind === 'failure' ? 'Failed' : 'Completed'
+      badge.kind === 'failure' ? colors.$textDanger : colors.$textSuccess
     return (
       <View sx={{ gap: 2 }}>
         <Text
@@ -73,11 +88,11 @@ const RecurringSwapItem: FC<RecurringSwapItemProps> = ({
           {notification.body}
         </Text>
         <Text variant="body2" sx={{ color: badgeColor }}>
-          {badgeLabel}
+          {badge.label}
         </Text>
       </View>
     )
-  }, [badgeKind, notification.body, colors.$textDanger, colors.$textSuccess])
+  }, [badge, notification.body, colors.$textDanger, colors.$textSuccess])
 
   return (
     <NotificationListItem
--- a/packages/core-mobile/app/new/features/notifications/components/RecurringSwapItem.tsx
+++ b/packages/core-mobile/app/new/features/notifications/components/RecurringSwapItem.tsx
@@ -5,6 +5,7 @@ import {
   isRecurringSwapNotification,
   RecurringSwapMetadata
 } from '../types'
+import { classifyRecurringSwapStatus } from '../utils'
 import NotificationListItem from './NotificationListItem'
 import NotificationIcon from './NotificationIcon'
 
@@ -35,19 +36,21 @@ type Badge = { kind: 'success' | 'failure'; label: string } | null
 
 function resolveBadge(data: RecurringSwapMetadata | undefined): Badge {
   if (data === undefined) return null
-  const status = data.status.toLowerCase()
-  if (status === 'failed') return { kind: 'failure', label: 'Failed' }
+  // Shared classifier (see notifications/utils.ts) so the badge stays in lockstep
+  // with `isTerminalRecurringSwapNotification`'s row-actionability logic.
+  const kind = classifyRecurringSwapStatus(data.status)
+  if (kind === 'failed') return { kind: 'failure', label: 'Failed' }
 
-  const isSuccessFill =
-    status === 'active' || status === 'completed' || status === 'executed'
-  if (!isSuccessFill) return null
+  // Only success fills get a badge; 'cancelled' and unrecognized statuses show
+  // none (a cancelled schedule was deliberately stopped, not a fill event).
+  if (kind !== 'completed' && kind !== 'fill') return null
 
   // The schedule is finished when the backend marks it completed, or — for
   // finite schedules — when no fills remain. Infinite/DCA schedules
   // (numberOfOrders === -1) never reach a completed terminal, so they always
   // read as a mid-schedule fill.
   const isFinalLeg =
-    status === 'completed' ||
+    kind === 'completed' ||
     (data.numberOfOrders !== -1 && data.remainingOrders === 0)
 
   return { kind: 'success', label: isFinalLeg ? 'Completed' : 'Executed' }
--- a/packages/core-mobile/app/new/features/notifications/components/RecurringSwapItem.tsx
+++ b/packages/core-mobile/app/new/features/notifications/components/RecurringSwapItem.tsx
@@ -1,0 +1,93 @@
+import React, { FC, useMemo } from 'react'
+import { Text, useTheme, View } from '@avalabs/k2-alpine'
+import { AppNotification, isRecurringSwapNotification } from '../types'
+import NotificationListItem from './NotificationListItem'
+import NotificationIcon from './NotificationIcon'
+
+type RecurringSwapItemProps = {
+  notification: AppNotification
+  showSeparator: boolean
+  accessoryType: 'chevron' | 'none'
+  testID?: string
+}
+
+// Per Sarp's PR #174 the notification-sender service formats the push
+// envelope's `title` + `body` server-side (e.g. "Recurring swap executed",
+// "Insufficient balance — your recurring swap schedule will be cancelled"),
+// so the row just renders those verbatim. The `data.status` field is used
+// only to drive the colored status badge under the subtitle — that's the
+// one piece of UI we can synthesize client-side without lossy formatting.
+//
+// Status → badge mapping (matches the Figma design):
+//   - 'failed'                → red "Failed"
+//   - 'active' / 'completed'  → green "Completed"   (a fill landed; either a
+//                                                    mid-schedule leg or the
+//                                                    final one — both are
+//                                                    surfaced the same way)
+//   - anything else           → no badge
+type BadgeKind = 'success' | 'failure' | null
+
+function resolveBadgeKind(status: string | undefined): BadgeKind {
+  if (status === undefined) return null
+  const lower = status.toLowerCase()
+  if (lower === 'failed') return 'failure'
+  if (lower === 'active' || lower === 'completed') return 'success'
+  return null
+}
+
+const RecurringSwapItem: FC<RecurringSwapItemProps> = ({
+  notification,
+  showSeparator,
+  accessoryType,
+  testID
+}) => {
+  const {
+    theme: { colors }
+  } = useTheme()
+
+  const badgeKind = useMemo(
+    () =>
+      isRecurringSwapNotification(notification)
+        ? resolveBadgeKind(notification.data?.status)
+        : null,
+    [notification]
+  )
+
+  // Subtitle = backend-formatted body + optional colored status line. Wrapping
+  // in a View (rather than passing a string) lets NotificationListItem render
+  // both lines without us re-implementing its typography defaults.
+  const subtitle = useMemo<React.ReactNode>(() => {
+    if (badgeKind === null) return notification.body
+    const badgeColor =
+      badgeKind === 'failure' ? colors.$textDanger : colors.$textSuccess
+    const badgeLabel = badgeKind === 'failure' ? 'Failed' : 'Completed'
+    return (
+      <View sx={{ gap: 2 }}>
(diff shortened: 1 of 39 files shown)


packages/core-mobile/app/new/features/notifications/services/schemas.ts
```diff
@@ -93,6 +94,39 @@ export const NewsMetadataSchema = z.object({
...
+ *
+ * `reasonCode` is set only on `status === 'failed'`; the push sends it as a
+ * string (`String(reasonCode)`) but history sends the raw number — so it MUST
+ * be coerced too. Without this, every failed notification (which always
+ * carries a reasonCode) fails to parse on the history path, dropping the whole
+ * `data` block and with it the failure badge + terminal-state detection. Known
+ * codes are documented at the call site; unknown codes pass through so newer
+ * backend codes don't fail parsing.
+ */
+export const RecurringSwapMetadataSchema = z.object({
```
atn4z7, line 129: The coercion here is the right call for the two wire shapes, no notes on that. But making all the progress fields required worries me a bit. The row only ever reads `status`, `numberOfOrders`, `remainingOrders`, and `chainId` (`owner`, `executedOrders`, `tokenIn/Out`, `amountIn/Out` aren't touched by any consumer). Since the parse is all-or-nothing, if the backend ever drops one of those unused fields the whole `data` block goes `undefined`, and `isTerminalRecurringSwapNotification` then falls back to "not terminal" and the row turns tappable again, sending people to a schedule the manage screen won't show. That's the exact thing this PR is trying to prevent, reintroduced by a field we don't even render. Could we make everything but `status` optional so a stray missing field doesn't take down the failure badge and terminal detection? Is the full payload actually contractual on both the push and history channels?

packages/core-mobile/app/new/features/swap/screens/SwapScreen.tsx
```diff
@@ -386,6 +417,14 @@ export const SwapScreen = (): JSX.Element => {
   const [validationError, setValidationError] =
     useState<FusionQuoteError | null>(null)
   const minimumTransferAmount = useMinimumTransferAmount({ fromToken, toToken })
+  // Recurring is Markr-only, so its per-order floor is Markr's minimum
+  // specifically — not the blended "lowest across services" value, which could
+  // pin a lower non-Markr floor and let a sub-fillable per-order amount through.
+  const markrMinimumTransferAmount = useMinimumTransferAmount({
```
atn4z7: This fires on every swap, not just recurring ones, since it isn't gated on `recurring.isRecurring` and the `MARKR` serviceType gives it its own query key so it won't dedupe with the blended one. So a regular swap now does an extra `getMinimumTransferAmount` round trip it never uses. Could we pass `serviceType: recurring.isRecurring ? ServiceType.MARKR : undefined` so it collapses back onto the existing query when recurring is off? Just want to make sure it doesn't trip up the minimum-ready gate.

packages/core-mobile/app/new/features/notifications/components/RecurringSwapItem.tsx
```diff
@@ -0,0 +1,110 @@
...
+//   - 'failed'                       → red "Failed"
+//   - final leg                      → green "Completed"
+//       (`status === 'completed'`, or — for finite schedules — no fills left)
+//   - other success fill             → green "Executed"
+//       (a mid-schedule leg landed; infinite/DCA schedules never "complete",
+//        so they always read here)
+//   - anything else                  → no badge
+type Badge = { kind: 'success' | 'failure'; label: string } | null
+
+function resolveBadge(data: RecurringSwapMetadata | undefined): Badge {
```
atn4z7, line 37: Small thing: `resolveBadge` and `isTerminalRecurringSwapNotification` (over in `utils.ts`) each hardcode their own status sets in different files, and they don't quite line up (`cancelled` is terminal but gets no badge, for instance). Probably the intended look, but they'll drift the next time someone touches one and forgets the other. Worth a shared status classifier, or at least a comment in each pointing at the other.

packages/core-mobile/app/new/features/swap/utils/recurringBalanceGate.ts
```diff
@@ -0,0 +1,108 @@
...
+// inline and disables submission via `isRecurringReady`.
+
+import { formatTokenAmount } from 'utils/Utils'
+import { bigintToBig } from '@avalabs/core-utils-sdk'
+import { TokenType } from '@avalabs/vm-module-types'
+import type { LocalTokenWithBalance } from 'store/balance'
+import type { NumberOfOrders } from 'features/recurringSwap/types'
+import { fusionErrors, type FusionQuoteError } from './fusionErrors'
+
+const formatTokenWithSymbol = (
```
atn4z7, line 14: The `else` branch returns a bare wei value with no symbol, which would land in a user-facing "Insufficient balance" string if it ever hit. A `LocalTokenWithBalance` always has `decimals` so this looks dead, and the tests only cover the happy path. Either drop the fallback and lean on the type, or assert it can't happen, so nobody trusts it later.

packages/core-mobile/app/new/features/recurringSwap/services/recurringSwapNotifications.ts
```diff
@@ -0,0 +1,40 @@
...
+
+type RecurringSwapNotificationSubscription = {
+  orderId: string
+  // Resolved once per batch by the caller (`ensureOrderSubscriptions`) and
+  // reused across orders, mirroring the balance-change subscribe flow — so a
+  // multi-order snapshot triggers one `/v1/push/register`, not one per order.
+  deviceArn: string
+}
+
+export async function subscribeRecurringSwapNotifications(
```
atn4z7: The other notification types keep their backend subscribe call in `services/notifications/<type>/subscribeForX.ts` (balance change, news, price alert all follow this) and their orchestration in `store/notifications/listeners/`. This one puts both halves inside the recurringSwap feature folder, so anyone auditing what the app subscribes to by reading those two shared locations won't find recurring swap. This file in particular is a plain parameterized POST with no recurring-specific logic, structurally identical to `subscribeForBalanceChange.ts`, so it'd move cleanly to `services/notifications/recurringSwap/subscribeForRecurringSwap.ts` with no downside. The orchestration in `store/listeners.ts` I get is more coupled to the order cache so it's more of a judgment call, but is the split intentional or just where it landed?

packages/core-mobile/app/new/features/recurringSwap/store/listeners.ts
```diff
@@ -14,11 +17,43 @@ import {
...
+// session — this cache collapses that across batches.
+//
+// Invalidated on:
+//   - subscribe failure inside the loop (ARN may have been retired backend-side;
+//     drop it so the next batch re-registers and the live ARN gets reused)
+//   - `onFcmTokenChange` (the SNS endpoint mapping is stale until /v1/push/register
+//     posts the new FCM token, so we MUST force a fresh register or push
+//     deliveries will dead-letter against the old token)
+//   - `onLogOut` (a fresh sign-in may use a different device identity)
+let cachedDeviceArn: string | null = null
```
atn4z7: Tied to the structure point on `recurringSwapNotifications.ts`: every other subscriber just calls `registerAndGetDeviceArn()` and leans on its built-in concurrent dedup, none of them keep their own ARN cache. Here we've got a module-scoped `cachedDeviceArn` plus an `onFcmTokenChange`/`onLogOut` listener to invalidate it. I think you can drop both: the ARN is already persisted in `commonStorage` under `NOTIFICATIONS_OPTIMIZATION`, the global notification flow already re-registers and overwrites it on `onFcmTokenChange`, and logout's `clearAll()` already wipes it. So reading from storage gets you the same freshness with less machinery, and it survives cold start too. The one thing to keep is forcing a `registerAndGetDeviceArn()` on a subscribe failure, since a dead endpoint won't be caught by re-reading the same key.

atn4z7 (commented):
Nice PR.

packages/core-mobile/app/new/features/notifications/services/schemas.ts
ruijia1in: Good catch — fixed in 24a3d2086. Everything but `status` is now `.optional()` on `RecurringSwapMetadataSchema`, so a dropped field that no consumer renders no longer takes the whole `data` block down with it and flips a terminal row back to tappable. The only other fields any consumer reads are `numberOfOrders`/`remainingOrders` (final-leg "Completed" vs "Executed" detection); when absent they degrade gracefully — the status-based terminal states (`completed`/`cancelled`/`failed`) and the failure badge still resolve, only the last-leg nuance is lost. On your contract question: it is not safe to assume the full payload is contractual on both the push and history channels, which is exactly why I tightened to status-only-required. Added two schema tests: a status-only payload parses, and a payload missing `status` fails.

packages/core-mobile/app/new/features/swap/screens/SwapScreen.tsx
ruijia1in: Fixed in 24a3d2086. `serviceType` is now `recurring.isRecurring ? ServiceType.MARKR : undefined`, so when recurring is off it collapses back onto the blended `minimumTransferAmount` query key and React Query dedupes them — no extra `getMinimumTransferAmount` round trip on one-shot swaps. The minimum-ready gate tolerates the brief loading window on toggle-on (Next stays disabled until the Markr query settles), so this does not reopen the sub-minimum bypass.

packages/core-mobile/app/new/features/notifications/components/RecurringSwapItem.tsx
ruijia1in: Fixed in 24a3d2086. Added `classifyRecurringSwapStatus` in `notifications/utils.ts` as the single source of truth for the raw-status vocabulary, returning a semantic kind (`failed`/`cancelled`/`completed`/`fill`/`unknown`). Both `resolveBadge` (RecurringSwapItem) and `isTerminalRecurringSwapNotification` (utils) now route through it instead of re-listing literals, so the badge and row-actionability can no longer drift. The `cancelled`-is-terminal-but-no-badge asymmetry you flagged is intentional (a cancelled schedule was deliberately stopped, not a fill event) and is now explicit in one place. Added unit tests for the classifier.

packages/core-mobile/app/new/features/swap/utils/recurringBalanceGate.ts
ruijia1in: Fixed in 24a3d2086. Dropped the bare-wei fallback — `formatTokenWithSymbol` now throws if `decimals` is absent, with a message naming the offending token. The gate only ever receives a swap source or its native gas token (never an NFT), so the type-level union guard was effectively dead; asserting the invariant means nobody downstream trusts a symbol-less wei value leaking into the "Insufficient balance" string.

packages/core-mobile/app/new/features/recurringSwap/services/recurringSwapNotifications.ts
ruijia1in: Moved in 24a3d2086 — it now lives at `services/notifications/recurringSwap/subscribeForRecurringSwap.ts` (renamed the export to `subscribeForRecurringSwap`), matching `subscribeForBalanceChange.ts` et al., so an audit of what the app subscribes to from the shared location now surfaces recurring swap too. The split was just where it landed, not intentional. I left the orchestration in `recurringSwap/store/listeners.ts` as you noted — it is coupled to the order cache / listOrders watcher, so it is the judgment-call half; happy to move it under `store/notifications/listeners/` in a follow-up if you would prefer the strict parity.

packages/core-mobile/app/new/features/recurringSwap/store/listeners.ts
ruijia1in: Done in 24a3d2086. Dropped the module-scoped `cachedDeviceArn` and the `onFcmTokenChange`/`onLogOut` invalidation listener entirely. `ensureOrderSubscriptions` now reads the persisted ARN from `commonStorage` under `NOTIFICATIONS_OPTIMIZATION` — kept fresh by the global notification flow on FCM token rotation, wiped by logout `clearAll()` — so it gets the same freshness as the memo with none of the machinery, and it survives cold start. The one thing I kept, per your note: on a subscribe failure I force a `registerAndGetDeviceArn()` (re-reading the same key would just hand back the same dead ARN) and reuse the live ARN for the rest of the loop.

atn4z7 (dismissed):

atn4z7 (dismissed):

atn4z7 (approved):

B0Y3R-AVA (approved):

PR #3934, opened 2026-06-30
CP-14615: build injected provider shim once per tab to fix swap spinner after network switch

B0Y3R-AVA:
## Summary

The in-app browser's injected EVM provider froze with an endless spinner when a dApp (e.g. Uniswap) switched networks and then tried to swap. Seedless + Ethereum worked and the first tx worked; the freeze surfaced after switching networks — matching QA's "first tx works, switching confuses eth/avax". This is **not** a signer/MFA/C-Chain-specific bug.

### Root cause
`providerShimJs` was memoized on `chainIdHex`. A `wallet_switchEthereumChain` dispatches `setTabChainId`, which changes `chainIdHex` → rebuilds `providerShimJs` → changes the WebView's `injectedJavaScriptBeforeContentLoaded` prop. react-native-webview responds by calling `resetupScripts` (removeAllUserScripts + re-add) on the live WKWebView, and the `injectJavaScript` issued in that same tick — the switch's own `__coreProviderRespond` — is silently dropped. The dApp's `switchChain` promise never resolves, so wagmi / the swap saga hangs forever. (Confirmed on device: the switch's `sendResponse` ran natively but its injected JS never executed.)

### Fix
- **Build `providerShimJs` exactly once per tab** (`useRef(chainIdHex)` + `useMemo(..., [])`). A chain switch no longer changes the WebView prop, so `resetupScripts` no longer drops the switch response. The shim keeps its live chain current via the optimistic emit + the re-assert below, so the baked initial chain only matters on first load. Does not touch the optimistic-emit path → no React #185 regression.
- **Re-assert the live chain on every committed URL** so a fresh document (full reload / cross-origin nav / pull-to-refresh) isn't stuck on the stale baked chainId. Guarded on the resolved chain's own `vmName` (EVM only), via the new `__coreProviderReassertChain`.
- **`__coreProviderReassertChain`** dedupes numerically against the shim's live `_chainId` (no churn on same-origin SPA navigation; handles non-canonical hex), kept separate from `__coreProviderEmit` so the disconnect→same-chain recovery path is unaffected.
- **Normalize the optimistic `wallet_switchEthereumChain` chainId** to canonical lowercase hex so `eth_chainId` is always well-formed.

Unit tests added for each part.

## Dev testing

Device-validated on a physical iOS build (seedless wallet):
1. In-app browser → Uniswap → connect the seedless account.
2. Switch the tab between Ethereum and Avalanche repeatedly.
3. Swap on Ethereum, then swap AVAX → USDC on Avalanche C-Chain.

Expected: every switch resolves and both swaps complete — **no endless spinner**. Confirmed: both swaps landed (FCM "received USDC on Avalanche (C-Chain)"), and the wallet held the correct chain across a pull-to-refresh.

Bitrise builds: iOS #9045, Android #9046.

### DIFF
--- a/packages/core-mobile/app/hooks/browser/evmProviderShim.test.ts
+++ b/packages/core-mobile/app/hooks/browser/evmProviderShim.test.ts
@@ -93,6 +93,36 @@ describe('buildEvmProviderShim', () => {
     })
   })
 
+  describe('__coreProviderReassertChain (CP-14615)', () => {
+    it('defines an idempotent chain re-assert helper', () => {
+      const shim = buildEvmProviderShim(defaultParams)
+      expect(shim).toContain(
+        'window.__coreProviderReassertChain = function(data)'
+      )
+    })
+
+    it('dedupes numerically so a non-canonical hex (0x01 / 0xA86A) is not re-emitted', () => {
+      // Native always sends canonical '0x'+n.toString(16); a dApp-supplied chain
+      // (wallet_addEthereumChain) may be zero-padded/uppercase. A string compare
+      // would miss those; the numeric compare is the correct dedupe.
+      const shim = buildEvmProviderShim(defaultParams)
+      expect(shim).toContain('parseInt(data, 16) === parseInt(_chainId, 16)')
+    })
+
+    it('emits chainChanged via __coreProviderEmit when the chain actually changed', () => {
+      const shim = buildEvmProviderShim(defaultParams)
+      expect(shim).toContain("window.__coreProviderEmit('chainChanged', data)")
+    })
+
+    it('normalizes the optimistic wallet_switchEthereumChain chainId to canonical hex', () => {
+      // A dApp may send a non-canonical chainId ('1' / '0x01' / '0xA86A'); the shim
+      // must store canonical '0x'+n.toString(16) so eth_chainId is well-formed and
+      // the numeric re-assert dedupe stays consistent. CP-14615.
+      const shim = buildEvmProviderShim(defaultParams)
+      expect(shim).toContain("swTargetChainId = '0x' + swParsed.toString(16)")
+    })
+  })
+
   describe('EIP-1193 provider object', () => {
     let shim: string
 
--- a/packages/core-mobile/app/hooks/browser/useEvmInjectedProvider.ts
+++ b/packages/core-mobile/app/hooks/browser/useEvmInjectedProvider.ts
@@ -223,7 +223,13 @@ export function useEvmInjectedProvider(
       // tab that retains a live EVM chain after the wallet moved to a non-EVM
       // network (CP-13672) is still re-asserted.
       const liveChainId = browserNetworkRef.current.chainId
-      const liveNetwork = allNetworksRef.current[liveChainId]
+      // Fall back to the store like requestReadOnly does: allNetworksRef syncs via
+      // a useEffect, so a chain just added through wallet_addEthereumChain can be
+      // in Redux but not yet in the ref. Without the fallback a reload in that
+      // window would skip the re-assert for that fresh EVM chain.
+      const liveNetwork =
+        allNetworksRef.current[liveChainId] ??
+        selectAllNetworks(store.getState())[liveChainId]
       if (liveNetwork?.vmName === NetworkVMType.EVM) {
         webViewRef.current?.injectJavaScript(
           `window.__coreProviderReassertChain && window.__coreProviderReassertChain('0x${liveChainId.toString(
@@ -241,9 +247,13 @@ export function useEvmInjectedProvider(
       // prime point that re-establishes connected state. CP-13772.
       primeAccountsRef.current?.()
     },
-    [tabId, webViewRef]
+    [tabId, webViewRef, store]
   )
 
+  // Only consumed by initialChainIdHexRef below (read once at mount). Kept as a
+  // useMemo rather than inlined into the hook body so this branch stays in a
+  // nested function — inlining it tips useEvmInjectedProvider over the
+  // sonarjs/cognitive-complexity limit.
   const chainIdHex = useMemo(() => {
     if (activeNetwork.vmName !== NetworkVMType.EVM) return '0x1'
     return '0x' + initialChainId.toString(16)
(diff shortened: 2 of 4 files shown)


packages/core-mobile/app/hooks/browser/useEvmInjectedProvider.ts
```diff
@@ -212,20 +241,36 @@ export function useEvmInjectedProvider(
...
+  // tick is silently dropped. Previously this memo depended on `chainIdHex`, so a
+  // `wallet_switchEthereumChain` (which dispatches `setTabChainId` -> new
+  // chainIdHex) rebuilt the shim and dropped the switch's OWN response injection,
+  // leaving the dApp's switchChain promise — and the swap awaiting it — hung
+  // forever (CP-14615). The baked chainId only seeds the first document load; the
+  // live chain is kept current at runtime via the shim's optimistic `_chainId`
+  // update on switch, the unpinned sync effect's chainChanged emit, and the
+  // per-document-load re-assert in handleCommittedUrl. So bake the initial value
+  // once and never change the prop.
+  const initialChainIdHexRef = useRef(chainIdHex)
```
atn4z7, line 275: `chainIdHex` is now near-dead computation. Its only consumer is this `useRef(chainIdHex)` initializer, which reads it once on mount — the `useMemo` above still recomputes on every `activeNetwork.vmName` / `initialChainId` change but the result is discarded after first render. Harmless, but the memo could be simplified (or inlined into the ref initializer) now that nothing reactive depends on it. Non-blocking — leaving it is also defensible for readability.

packages/core-mobile/app/hooks/browser/useEvmInjectedProvider.ts
```diff
@@ -203,6 +203,35 @@ export function useEvmInjectedProvider(
...
+      // listeners or fight a still-pending switch (React #185).
+      //
+      // Guard on the *resolved chain's* own vmName — NOT tabChainId/activeNetwork
+      // — so a non-EVM chainId is never emitted to the EVM provider: a tab pinned
+      // to a non-EVM chain (wallet_switchEthereumChain validates only membership
+      // in the cross-VM network map, not EVM-ness) is skipped, while an unpinned
+      // tab that retains a live EVM chain after the wallet moved to a non-EVM
+      // network (CP-13672) is still re-asserted.
+      const liveChainId = browserNetworkRef.current.chainId
+      const liveNetwork = allNetworksRef.current[liveChainId]
```
atn4z7: Minor asymmetry: this re-assert reads `allNetworksRef.current[liveChainId]` with no fallback to `selectAllNetworks(store.getState())`, unlike the read-only path (`requestReadOnly`) which explicitly falls back to the store to close the stale-ref race for a chain just added via `wallet_addEthereumChain`. So a freshly-added EVM chain that's in Redux but not yet in the ref would skip the re-assert on a reload. The window is tiny (ref synced via `useEffect`), so this is a very minor edge case, not a defect — just flagging in case it's worth mirroring the fallback here.

packages/core-mobile/app/hooks/browser/evmProviderShim.test.ts
```diff
@@ -93,6 +93,36 @@ describe('buildEvmProviderShim', () => {
...
+        'window.__coreProviderReassertChain = function(data)'
+      )
+    })
+
+    it('dedupes numerically so a non-canonical hex (0x01 / 0xA86A) is not re-emitted', () => {
+      // Native always sends canonical '0x'+n.toString(16); a dApp-supplied chain
+      // (wallet_addEthereumChain) may be zero-padded/uppercase. A string compare
+      // would miss those; the numeric compare is the correct dedupe.
+      const shim = buildEvmProviderShim(defaultParams)
+      expect(shim).toContain('parseInt(data, 16) === parseInt(_chainId, 16)')
```
atn4z7, line 109: These assertions check the shim *source text* (`toContain('parseInt(data, 16) === parseInt(_chainId, 16)')` etc.) rather than behavior, so they'd pass even if the dedupe/normalize logic were wrong and break on a harmless rename. Matches the existing pattern in this file, and the `useEvmInjectedProvider.test.ts` additions are properly behavioral, so this is the weaker half only. Optional: one behavioral test that `eval`s the shim and calls `window.__coreProviderReassertChain` with a non-canonical hex would lock in the actual dedupe behavior.

atn4z7 (commented):
Nice PR.

atn4z7 (dismissed):
dev tested

packages/core-mobile/app/hooks/browser/useEvmInjectedProvider.ts
B0Y3R-AVA: Good catch. I tried inlining it into the ref initializer, but moving that ternary out of the useMemo callback into the hook body tips `useEvmInjectedProvider` over the `sonarjs/cognitive-complexity` limit (the memo callback counts as a separate function). So I kept it as a `useMemo` per your "defensible for readability" note and added a comment explaining why it stays. No functional change.

packages/core-mobile/app/hooks/browser/useEvmInjectedProvider.ts
B0Y3R-AVA: Done in 603b8e4f7 — mirrored the `requestReadOnly` pattern: `allNetworksRef.current[liveChainId] ?? selectAllNetworks(store.getState())[liveChainId]`, and added `store` to the useCallback deps (stable identity). Closes the add-chain-then-reload window you described.

packages/core-mobile/app/hooks/browser/evmProviderShim.test.ts
B0Y3R-AVA: Agreed the source-text assertions are the weaker half. I looked at adding a behavioral eval test but the RN jest env is node-based (no DOM) and jsdom isn't a dependency, so there's no way to `eval` the shim IIFE here without introducing a new jsdom env/dep — which is exactly why the existing shim tests in this file are string-based. The RN-side re-assert behavior (EVM-guard, non-EVM skip) is covered behaviorally in useEvmInjectedProvider.test.ts. Leaving the string assertions to match the file's established pattern; happy to revisit if we adopt a jsdom env for shim tests more broadly.

atn4z7 (approved):

bogdandobritoiu (approved):

PR #3937, opened 2026-07-01
CP-14631, CP-14632: fix My Wallets missing balance/empty list & portfolio header animation

bogdandobritoiu:
## Description

**Tickets: [CP-14631](https://ava-labs.atlassian.net/browse/CP-14631), [CP-14632](https://ava-labs.atlassian.net/browse/CP-14632)**

Fixes a cluster of "My wallets" / portfolio regressions surfaced on the 1.0.34 RC after the RN 0.85 / Expo 56 (New Architecture / Fabric) upgrade, plus the underlying performance problem behind them.

### 1. My Wallets — flatten the accordion into a virtualized list (perf + card overlap)

Previously each wallet's `WalletCard` rendered **all of its accounts inside a single FlashList cell**. That one decision caused both the slow performance (nothing virtualized; every account mounted and computed balances at once; a wallet switch re-rendered the whole giant cell) and the on-device/Release **card overlap** (a giant, dynamically-sized cell is exactly what FlashList mis-measures on Fabric, so the following wallet cards rendered on top of the expanded rows).

Refactored to a **flat, mixed-item virtualized list**:
- `WalletsScreen` builds a flat `ListRow[]` (`walletHeader` | `account` | `addAccount`) via the new pure `buildWalletListRows` (unit-tested), with `getItemType` per kind and a stable `keyExtractor`.
- New presentational components: `WalletHeaderRow`, `AccountRow` (computes its own balance, so only visible rows compute), `AddAccountRow`, and `CardRow` (recreates the grouped-card visuals — rounded top/bottom, borders, spacing — from a per-row `cardPos`).
- `WalletCard` is deleted.

Result: account rows are individually virtualized (perf), and every cell is small/uniform so there's no giant dynamically-sized cell (overlap gone by construction). Expand/collapse is now **instant** (the previous Fabric-fragile height animation is removed).

### 2. Portfolio balance / price-change header animation (CP-14632)

The shared `AnimateFadeScale` primitive "failed invisible" on Fabric — it started at `opacity: 0` and relied on a mount effect to reveal, which could be dropped before the view attached, leaving balances/price-change characters stuck blank. Rewrote it to a declarative `entering` keyframe whose **resting style is visible** (fails visible), and keyed `AnimatedText` characters by index (not glyph) so digit changes update in place instead of remounting and re-fading. Central fix; covers all balance/price-change sites.

### 3. Supporting Fabric fixes

- **`ContentReveal`** (balance-error banner): measured via `onLayout` instead of the legacy `ref.measure()` (which returns 0 on Fabric/Android) and added `overflow: 'hidden'`, so the absolutely-positioned banner no longer bleeds **under the list** on Android.
- **`EmptyState`**: replaced `ref.measure()` with `onLayout` (same Fabric-reliability class; also removes a per-render re-measure).
- **Stack navigation animation** (`screenOptions`): `slide_from_right` → `ios_from_right` — after the upgrade, Android stack screens opened from the bottom like a modal; `ios_from_right` restores the right-slide (iOS unchanged).
- **My Wallets route background** (`(modals)/wallets/_layout`): set the native screen `contentStyle.backgroundColor` to match the list, so the route background doesn't flash during the push transition / overscroll.
- **Pull-to-refresh position** on `WalletsScreen`.

## Testing

**Dev Testing**

- **My Wallets**: with a many-account wallet (e.g. 48 accounts), expand/collapse repeatedly, scroll hard, and switch the active wallet. Confirm: account lists always populate, balances stay visible, no card overlap (verify on a **Release** build / real device — the overlap was device/Release-only), and scrolling is smooth.
- **Portfolio header**: let balances refresh and force updates; confirm balance and price-change digits animate in cleanly with no blank/stuck characters, and update in place.
- **Android**: confirm stack screens (account settings, etc.) slide in **from the right**, not the bottom; confirm the balance-error banner sits above the list (not bleeding underneath).

**QA Testing**

- Verify My Wallets balances/account lists and the portfolio balance/price-change header render correctly across repeated interactions on both iOS and Android.
- Expected: no overlap, no missing/blank balances, smooth scroll, correct Android push animation.

## Checklist

[CP-14631]: https://ava-labs.atlassian.net/browse/CP-14631?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CP-14632]: https://ava-labs.atlassian.net/browse/CP-14632?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/core-mobile/app/new/features/wallets/components/AccountListItem.tsx
```diff
@@ -39,18 +39,20 @@ const AccountListItem = ({
...
           flexDirection: 'row',
           alignItems: 'center',
-          paddingLeft: 10,
-          borderRadius: 12,
-          height: 52,
-          backgroundColor: isActive
-            ? alpha(theme.colors.$textPrimary, 0.1)
-            : 'transparent'
+          paddingLeft: 14
         }}>
```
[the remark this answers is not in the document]
bogdandobritoiu: Fixed

packages/core-mobile/app/new/features/wallets/utils/buildWalletListRows.ts
```diff
@@ -0,0 +1,97 @@
...
+      })
+      continue
+    }
+
+    rows.push({
+      kind: 'walletHeader',
+      wallet,
+      isActive: isActiveWalletId(wallet.id),
+      isExpanded: true,
+      cardPos: 'top'
```
atn4z7: An expanded wallet with zero accounts loses some behavior the old `WalletCard` had. The deleted card rendered a "No accounts in this wallet." placeholder for `wallet.accounts.length === 0`; here the account loop just produces nothing, so an empty mnemonic wallet shows only the header + Add-account button with no message, and an empty `PRIVATE_KEY` wallet emits a single header row stuck at `cardPos: 'top'` (square bottom, no bottom border) because no account/addAccount row ever closes the card.

Low severity since wallets normally have at least one account, but the old code guarded this case explicitly. Worth either restoring an empty-state row or falling back to `cardPos: 'single'` when a wallet ends up with no closing row.

atn4z7 (commented):
Nice PR.

packages/core-mobile/app/new/features/wallets/utils/buildWalletListRows.ts
```diff
@@ -0,0 +1,97 @@
...
+    })
+
+    const hasAddAccount = wallet.type !== WalletType.PRIVATE_KEY
```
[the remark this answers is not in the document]
bogdandobritoiu: Guarded this. The builder now computes whether any rows follow the header and uses `cardPos: 'single'` when an expanded wallet has no accounts and no Add-account row (the empty `PRIVATE_KEY` case), so the card closes properly instead of a square `top` with the seam-overlap margin. Added unit tests. Note this path is currently unreachable — `WalletsScreen` returns `null` for zero-account wallets — but the builder is now correct on its own.

packages/core-mobile/app/new/features/wallets/utils/buildWalletListRows.ts
bogdandobritoiu: Fixed via a `cardPos: 'single'` fallback: an expanded wallet with no closing row now emits a single header that closes the card, instead of a square, unclosed `top`. I did not restore the "No accounts in this wallet." placeholder — zero-account wallets are filtered out upstream in `WalletsScreen` (`return null`), so it would be dead code; the fallback keeps the pure builder correct on its own. Added tests for the empty `PRIVATE_KEY` (single header) and empty mnemonic (header + Add-account closes the card) cases.

packages/core-mobile/app/new/features/wallets/components/WalletHeaderRow.tsx
```diff
@@ -0,0 +1,254 @@
...
+  next: React.ComponentProps<typeof WalletHeaderRow>
+): boolean {
+  return (
+    prev.isActive === next.isActive &&
+    prev.isExpanded === next.isExpanded &&
+    prev.isRefreshing === next.isRefreshing &&
+    prev.cardPos === next.cardPos &&
+    prev.showMoreButton === next.showMoreButton &&
+    prev.onToggleExpansion === next.onToggleExpansion &&
+    prev.wallet === next.wallet
```
B0Y3R-AVA: **Important (perf regression).** This is pure reference equality, but `walletsDisplayData` spreads a fresh `{...wallet}` object per wallet whenever `allAccountsArray` or `activeAccountId` changes (see `WalletsScreen.tsx` `primaryWalletsDisplayData`/`importedWalletsDisplayData`). The deleted `WalletCard` deep-compared exactly to avoid this, and the sibling `AccountRow` already does field-level comparison — so every header re-renders (re-running `WalletBalance` + `getEnabledNetworksForAccount` for all its accounts) on any account/active-account change, on the screen this PR is making smoother. Suggest a field-level compare on the account-id set (what drives the balance query + count text):

```suggestion
    (prev.wallet === next.wallet ||
      (prev.wallet.id === next.wallet.id &&
        prev.wallet.name === next.wallet.name &&
        prev.wallet.type === next.wallet.type &&
        prev.wallet.accounts.length === next.wallet.accounts.length &&
        prev.wallet.accounts.every(
          (acc, i) => acc.account.id === next.wallet.accounts[i]?.account.id
        )))
```

packages/core-mobile/app/new/features/wallets/screens/WalletsScreen.tsx
```diff
@@ -333,7 +342,8 @@ export const WalletsScreen = (): JSX.Element => {
       flatListRef={listRef}
       title="My wallets"
       subtitle={`An overview of your wallets\nand associated accounts`}
-      data={walletsDisplayData}
+      data={rows}
+      contentInset={{ top: headerHeight }}
```
B0Y3R-AVA, line 346: **Verify on iOS device.** `contentInset` flows through to the underlying `KeyboardAwareScrollView` (iOS-only; Android ignores it), but `ListScreenV2`'s header sentinel already reserves `paddingTop: headerHeight + 16` in scroll content — so `headerHeight` is accounted for twice on iOS. This is also the only `ListScreenV2` caller passing `contentInset`. It may be the intended pull-to-refresh-position fix and already device-tested; if so ignore. If not, the resting top gap will be oversized / initial offset off by ~`headerHeight`. Quick A/B: remove this line on an iOS device and check the top gap + post-refresh scroll position.

packages/core-mobile/app/new/features/wallets/components/AccountRow.tsx
```diff
@@ -0,0 +1,111 @@
...
+    tokenVisibility,
+    isBalancesError
+  ])
+
+  return (
+    <CardRow cardPos={cardPos}>
+      <View
+        sx={{
+          paddingHorizontal: 10,
+          paddingBottom: cardPos === 'bottom' ? 10 : 0
```
B0Y3R-AVA: **Visual (both platforms).** `CardRow`'s `marginBottom: -1` seam overlap makes the next opaque cell cover the bottom 1px of this one — which is exactly where `AccountListItem`'s 1px `Separator` sits (fixed `height:52` box → `flex:1` touchable + 1px separator, and middle rows get `paddingBottom:0`). So visible dividers between adjacent non-active accounts get painted over and vanish (iOS too — the "no-op on iOS" comment refers to the gap, not this overlap). Give middle rows a 1px bottom pad so the overlap eats padding, not the divider:

```suggestion
          paddingBottom: cardPos === 'bottom' ? 10 : 1
```

B0Y3R-AVA (commented):
Adversarial review (Claude + Codex, independent models). Core refactor is sound — no correctness/data bugs found; flattening the accordion into a virtualized mixed-item list is the right fix, and the AnimateFadeScale "fails-visible" rewrite + onLayout swaps are genuine Fabric improvements. A few polish/perf items inline below. Ranked: (1) header memo perf regression, (2) contentInset double-offset (verify on iOS), (3) seam overlap eats account separators, (4) defensive guard in the row builder, (5) nit. Nothing blocks correctness.

packages/core-mobile/app/new/features/wallets/utils/buildWalletListRows.ts
B0Y3R-AVA: **Minor (defensive / latent).** Not reachable today (`WalletsScreen` filters empty primary wallets and pools PRIVATE_KEY accounts into one virtual wallet), but an expanded wallet with 0 accounts AND no add-account row (PRIVATE_KEY) would emit only this `cardPos: 'top'` header with no closing row — an open-bottom card that still overlaps the next cell. Cheap insurance is to self-close the header when it has no body rows:

```ts
const hasAddAccount = wallet.type !== WalletType.PRIVATE_KEY
const hasChildRows = wallet.accounts.length > 0 || hasAddAccount

rows.push({
  kind: 'walletHeader',
  wallet,
  isActive: isActiveWalletId(wallet.id),
  isExpanded: true,
  cardPos: hasChildRows ? 'top' : 'single'
})
if (!hasChildRows) continue
```

Worth an accompanying unit test for the expanded-empty case (current tests only cover collapsed-empty).

packages/core-mobile/app/new/features/wallets/components/AccountListItem.tsx
```diff
@@ -39,18 +39,21 @@ const AccountListItem = ({
...
 
   return (
-    <View>
+    <View
+      sx={{
+        backgroundColor: isActive
+          ? alpha(theme.colors.$textPrimary, 0.1)
+          : 'transparent',
+        borderRadius: 12,
+        height: 52
```
B0Y3R-AVA, line 48: **Nit (optional).** A fixed `height: 52` on this wrapper containing both the `flex:1` touchable and the sibling 1px `Separator` shrinks the touchable to 51px when the divider renders (vs. the old auto-height ~53px). `minHeight: 52` avoids the squeeze:

```suggestion
        minHeight: 52
```

packages/core-mobile/app/new/features/wallets/components/WalletHeaderRow.tsx
bogdandobritoiu: Fixed. Replaced the reference-equality check with a field-level compare (id, name, type, `accounts.length`, and the account-id set) — restores the old `WalletCard` behavior so headers no longer re-render (and re-run `WalletBalance` + `getEnabledNetworksForAccount`) on every active-account change that re-spreads `walletsDisplayData`.

packages/core-mobile/app/new/features/wallets/screens/WalletsScreen.tsx
bogdandobritoiu: Intentional — the `contentInset` is the iOS pull-to-refresh position fix and was verified on an iOS device, so keeping it as-is.

packages/core-mobile/app/new/features/wallets/components/AccountRow.tsx
bogdandobritoiu: Fixed. Middle account rows now get `paddingBottom: 1`, so `CardRow`'s `-1` seam overlap covers that padding instead of the `AccountListItem` divider — dividers stay visible.

packages/core-mobile/app/new/features/wallets/utils/buildWalletListRows.ts
bogdandobritoiu: Already handled in a later commit — the builder self-closes the header (`cardPos: 'single'`) when an expanded wallet has no following rows (the empty `PRIVATE_KEY` case), with unit tests for the expanded-empty cases. This thread was on an earlier commit.

packages/core-mobile/app/new/features/wallets/components/AccountListItem.tsx
bogdandobritoiu: Acknowledged — leaving `height: 52` for now: the touchable is `flex: 1` so the touch target fills the row, and the 1px difference is cosmetic. Can switch to `minHeight` if squeeze shows up in practice.

atn4z7 (approved):

onghwan (approved):

B0Y3R-AVA (approved):
Re-reviewed after the latest commits (`0b9a5064`, `8ea2aa53`) — all findings from the adversarial review are addressed. Verified each fix against source:

- **WalletHeaderRow memo (perf)** — field-level compare restored (id/name/type/accounts.length + account-id set); live balance updates still flow via useWalletBalances, so nothing goes stale. ✅
- **AccountRow seam overlap** — `paddingBottom: … : 1` so CardRow's -1 overlap eats padding, not the divider; separators survive on both platforms. ✅
- **buildWalletListRows open-bottom card** — header self-closes with `cardPos: 'single'` when no body rows follow; two new unit tests cover the expanded-empty PRIVATE_KEY and mnemonic cases. ✅
- **contentInset** — confirmed intentional iOS pull-to-refresh fix, device-verified. ✅
- **AccountListItem 1px squeeze** — acknowledged nit, fine to leave as-is. ✅

TypeScript check passes; unit-test job was still running at review time (the new pure-function tests look correct). Nice work — no correctness issues, the perf regression is gone. LGTM.

PR #3938, opened 2026-07-01
CP-14641: Batch approval for recurring (Markr) swaps

ruijia1in:
iOS: 9161
Android: 9162
## Summary

Implements batch transaction approval on Core Mobile so recurring (DCA / "Markr") swaps can approve an ERC-20 `approve` + first swap-fill in a **single** approval flow, mirroring the Core browser extension. Jira: [CP-14641](https://ava-labs.atlassian.net/browse/CP-14641).

Previously a recurring first-fill that needed an allowance surfaced as multiple sequential single-tx approvals (or, when Quick-Swaps-eligible, auto-approved silently). This adds the missing **manual batch approval screen** and routes the recurring first-fill through it.

### Batch approval UI

- **New `BatchApprovalScreen`** (route `/approvalBatch`): aggregate balance-change overview → per-transaction stepper (`BatchTxStep`, with editable spend limit) → "Approve all" confirmation, with a progress-header overlay and page-aware back button. Reuses the existing Blockaid alert gate and renders the recurring-swap copy. No "Get free gas" toggle (hidden for parity with the extension — extension gasless is single-tx only).
- **`AccountNetworkCard`** — new shared component summarizing the signing account + network, used identically across the overview, each tx step, and the final confirm (previously duplicated inline).
- **`ActionSheet`** gains `headerCenterOverlay`, `renderHeaderLeft`, `showNavigationHeaderTitle`, and `centerContent` props to drive the batch screen's progress header and centered confirm content; **`ActionButtons`** `cancel` is now optional (stepper steps have no reject).
- **`getNetworkSymbol`** helper extracted into `chainIdsWithIncorrectSymbol` (type-only symbol import to avoid a circular dependency) and consumed by the approval screens.
- New `help` icon added to `k2-alpine` `Icons.Action`.

### Approval controller + signing

- **`ApprovalController.requestBatchApproval`** opens the manual batch screen for software wallets when a batch isn't auto-approved (`evaluateBatchApproval` → signed / reject / **manual**); on "Approve all" it signs each tx (applying any per-index spend-limit override) and returns the signed array.
- **Multi-chainId assertion** for batches; per-index spend-limit editing (unlimited / custom, via `useSpendLimitOverrides` + re-encoded `approve` calldata); flagged Quick-Swaps batches now route to the batch screen instead of the old per-tx fallback (dead manual-review fallback removed).
- **`selectIsBatchSigningSupported`** (MNEMONIC / SEEDLESS / PRIVATE_KEY allowlist) is now the single source of truth shared by the signer, the Quick Swaps gate, and the recurring fallback decision, so the PostHog kill-switch and wallet allowlist can't drift between consumers. `selectIsQuickSwapsActive` composes it with the flag + availability.
- **`BatchSigningUnsupportedError`** marker: hardware / WalletConnect wallets throw it from `EvmSigner.signBatch`, so the SDK's `fallbackToDefaultOnBatchFailure` re-issues each tx via the single-tx path.

### Recurring first-fill routing + gas fees

- **Recurring first-fill** now dispatches a real `eth_sendTransactionBatch` for software wallets (signs → the Fusion SDK broadcasts).
- **Batch gas fees are now filled** for the recurring first-fill. The SDK's one-click batch path (`executeFirstFill`) only spreads `maybe1559(gasSettings)` and never estimates fees, so without an explicit `maxFeePerGas` the batch txs reached `EvmSigner.signBatch` broadcast-unready and fell back to sequential approvals. Fixed by:
  - `buildFusionGasSettings` (extracted in `quickSwapsFee`) layers the EIP-1559 tier override on the base estimate-gas margin; recurring first-fills always apply it.
  - `SwapContext` exposes a `recurringGasSettings` memo carrying the tier override; `SwapScreen` forwards it + `isBatchSigningSupported` into `submitRecurringSwap` and gates the recurring "Next" button on `hasRecurringFees` so a fast tap during the cold-start window before `networkFees` load can't hard-fail the swap.
  - `submitRecurringSwap` sets `fallbackToDefaultOnBatchFailure` to `!isBatchSigningSupported` — software wallets reach the batch screen so a reject/error must abort cleanly rather than silently re-prompt sequentially; HW/WC still fall back to per-tx signing.

### Gasless intent (no-op today)

- Records the user's gasless choice onto `request.context` (`RequestContext.GASLESS_ENABLED`) via `handleBatchApprovalApprove`. The batch path only signs and hands the signed RLP to the EVM module to broadcast — there's no per-tx funding hook like the single-tx `onSigned` — so this is a harmless no-op that captures intent for when batch gasless support lands.

## Wallet-type support

Batch signing is limited to software wallets (mnemonic / seedless / private-key). Hardware (Ledger / Keystone) and WalletConnect fall back to sequential single-tx signing.

https://github.com/user-attachments/assets/90b893f8-72d3-4d96-9f6e-aca5c63a3b5b

https://github.com/user-attachments/assets/fc894aaa-18fc-4780-b662-85ce29191dca

https://github.com/user-attachments/assets/4fb203f0-ec09-4540-9850-95d097be3565

mock malicious blockaid simulation

https://github.com/user-attachments/assets/26ff10bd-8e85-49c9-a691-48dc9d552915

## Dev testing (manual QA)

On a **seed-phrase** wallet:
- [x] Schedule a recurring swap that needs an allowance → the batch approval screen appears with an aggregate **Balance Change** and no "Get free gas" toggle.
- [x] Tap **See details** → page through each transaction with Next / Previous; the "Transaction X of N" caption is correct.
- [x] On a transaction with a token approval, change the **Spend limit** (Unlimited / Custom) and confirm the value updates.
- [x] Tap **Approve all** → both transactions broadcast in order and the first fill lands; no duplicate `approve` prompt.
- [x] The recurring "Next" button stays disabled until fees load (cold-start), then enables — a fast tap doesn't hard-fail the swap.
- [x] Verify the recurring-swap copy (schedule / pause / resume / cancel) renders on the overview.
- [x] Reject from the batch screen → the request is rejected and the sheet dismisses.

On a **Ledger** wallet:
- [x] Schedule the same recurring swap → it falls back to sequential single-tx approvals (no batch screen), signs each step, and does not double-broadcast the `approve`.

Blockaid:
- [x]  Trigger a flagged batch → the warning renders and "Approve all" is gated behind the confirmation toggle.

## Notes

- Gasless on the batch path is intentionally not offered yet — funding is single-tx only; the `GASLESS_ENABLED` context capture is groundwork for future batch gasless support.
- Follow-up (optional, low priority): the `manualReviewReason` parameter threading on `signEachManually` / `signOne` is now unpopulated and could be pruned.

[CP-14641]: https://ava-labs.atlassian.net/browse/CP-14641?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/core-mobile/app/vmModule/ApprovalController/ApprovalController.ts
```diff
@@ -595,12 +597,104 @@ class ApprovalController implements VmModuleApprovalController {
...
+        onReject: (message?: string) => {
+          this.clearCancelBridge(requestId)
+          onReject({ resolve: resolve as never, message })
+        }
+      })
+
+      router.navigate({
+        pathname: '/approvalBatch',
+        params: { presentationMode: NavigationPresentationMode.FORM_SHEET }
+      })
```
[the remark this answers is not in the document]
ruijia1in: Batches are in-app only — the `eth_sendTransactionBatch` handler rejects any request not on `CORE_MOBILE_TOPIC`, so `isInAppRequest` is always true here and an external/injected batch can never reach this path. Still made the presentation mode conditional (`isInAppRequest(request) ? FORM_SHEET : undefined`) to mirror `requestApproval` so the two can't diverge if that trust boundary ever changes.

packages/core-mobile/app/new/features/approval/screens/BatchApprovalScreen/BatchApprovalScreen.tsx
```diff
@@ -0,0 +1,351 @@
...
+  const { recurringContext, isRecurringContextMalformed } =
+    useRecurringApprovalContext(params.request, rejectAndClose)
+
+  if (isRecurringContextMalformed) return null
+
+  // Defensive: the signer never produces a zero-tx batch, but guard it here so
+  // the inner screen's `signingRequests[0]` reads never hand `undefined` (cast
+  // to SigningData) to `useGasless`. Returning null before the inner mounts
+  // keeps that hook from ever running with a missing representative tx.
+  if (params.signingRequests.length === 0) return null
```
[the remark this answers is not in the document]
ruijia1in: Fixed. The zero-tx early-return now rejects + closes in an effect (mirroring the malformed-context path just above) before returning null as a render guard, so a violated invariant aborts the pending request instead of hanging behind a blank sheet.

packages/core-mobile/app/new/features/approval/screens/BatchApprovalScreen/BatchTxStep.tsx
```diff
@@ -0,0 +1,126 @@
...
+  const detailSections = useMemo(
+    () =>
+      displayData.details.map(section => ({
+        ...section,
+        items: section.items.filter(
+          item =>
+            typeof item === 'string' ||
+            !('label' in item) ||
+            item.label.toLowerCase() !== 'website'
+        )
```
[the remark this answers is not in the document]
ruijia1in: This is intentional and correct for the batch path: batches are in-app only (the `eth_sendTransactionBatch` handler rejects non-`CORE_MOBILE_TOPIC` requests), and `removeWebsiteItemIfNecessary` *hides* the Website row precisely for in-app requests — so reusing it here would produce the same result, never the "show for external dapp" branch (that branch is unreachable for batches). Keeping the inline filter also guards `!('label' in item)`, which `removeWebsiteItemIfNecessary` does not — the helper calls `item.label.toLowerCase()` on any non-string item, so swapping to it would be a robustness regression for label-less DetailItems. Left as-is.

packages/core-mobile/app/new/features/approval/screens/BatchApprovalScreen/BatchTxStep.tsx
```diff
@@ -0,0 +1,126 @@
...
+  ) => void
+}): JSX.Element => {
+  const { displayData } = signingRequest
+
+  const symbol = getNetworkSymbol(chainId)
+
+  const { spendLimits, canEdit, updateSpendLimit, hashedCustomSpend } =
+    useSpendLimits(displayData.tokenApprovals)
+
+  useEffect(() => {
```
atn4z7, line 48: **Spend-limit edit can leak onto the next transaction (wrong calldata signed)**

`BatchTxStep` is rendered without a `key`, so React reuses one instance across every step and only the `index` prop changes. `useSpendLimits` never clears `hashedCustomSpend` when `tokenApprovals` changes (its effect only calls `setSpendLimits`), so the value survives step navigation. Because `index` is a dependency of this effect, advancing to the next step re-fires it with the *previous* step's calldata:

1. Step 1 (index 0) is an `approve`; user edits the spend limit -> `hashedCustomSpend = calldata0`, `onOverride(0, calldata0)`.
2. Tap **Next** -> the same instance re-renders with `index = 1`; `hashedCustomSpend` is still `calldata0`; the effect re-fires as `onOverride(1, calldata0)`.
3. Overrides become `{0: calldata0, 1: calldata0}`, and `handleBatchApprovalApprove` replaces the swap tx's `.data` with the step-0 approve calldata, so a corrupted transaction is signed and broadcast.

This happens on the normal forward path right after using the screen's headline feature (editing a spend limit), and it isn't caught by tests: `BatchApprovalScreen.test.tsx` mocks `useSpendLimits` to `hashedCustomSpend: undefined`, and the controller test drives `onApprove` with a hand-built overrides map.

Simplest fix: give the step a stable identity per index so its hook state resets, e.g. `<BatchTxStep key={index} ... />`. Alternatives: reset `hashedCustomSpend` inside `useSpendLimits` when `tokenApprovals` changes, or drive `onOverride` from `updateSpendLimit` rather than from this effect.

atn4z7 (commented):
Nice PR.

packages/core-mobile/app/new/features/approval/screens/BatchApprovalScreen/BatchTxStep.tsx
ruijia1in: fixed.

atn4z7 (commented):
Nice PR.

packages/core-mobile/app/new/features/approval/screens/BatchApprovalScreen/BatchTxStep.tsx
```diff
@@ -0,0 +1,127 @@
...
+  ) => void
+}): JSX.Element => {
+  const { displayData } = signingRequest
+
+  const symbol = getNetworkSymbol(chainId)
+
+  const { spendLimits, canEdit, updateSpendLimit, hashedCustomSpend } =
+    useSpendLimits(displayData.tokenApprovals)
+
+useEffect(() => {
```
atn4z7: The guard added by the Copilot autofix is dedented to column 0 (it came in through the GitHub bot, bypassing the local prettier hook). This will fail `yarn lint` — reindent to 2 spaces to match the surrounding function body.

packages/core-mobile/app/new/features/approval/screens/BatchApprovalScreen/BatchApprovalScreen.tsx
```diff
@@ -0,0 +1,369 @@
...
+        {/* `key={index}` is load-bearing: BatchTxStep owns per-tx
+            `useSpendLimits` state (incl. `hashedCustomSpend`). Without a
+            per-step key React reuses ONE instance across every step, so an
+            edited spend limit's calldata survives into the next step and its
+            `onOverride` effect (which re-fires when `index` changes) would
+            register the previous step's calldata against the new index —
+            signing a corrupted tx. Keying by index remounts the step so its
+            hook state resets. (CP-14641) */}
+        <BatchTxStep
+          key={index}
```
atn4z7, line 317: Minor follow-up from the key fix (non-blocking): `overrides` persists in the parent while the remounted step resets to its default display. So editing a spend limit, advancing, then navigating back to that step shows the original limit even though the edited override is still what gets signed. It signs the user's real choice (no fund risk), but the review screen misrepresents it on a second visit. Consider seeding the step's `useSpendLimits` from the existing `overrides[index]` on mount so the display and the signed value stay in sync.

packages/core-mobile/app/new/features/approval/screens/BatchApprovalScreen/BatchTxStep.tsx
atn4z7: Confirmed fixed — `key={index}` remounts each step so `useSpendLimits` state resets, and the new regression test uses a state-faithful mock that would fail against the un-keyed version. Nice.

packages/core-mobile/app/new/features/approval/screens/BatchApprovalScreen/BatchTxStep.tsx
ruijia1in: fixed.

packages/core-mobile/app/new/features/approval/screens/BatchApprovalScreen/BatchApprovalScreen.tsx
ruijia1in: fixed.

atn4z7 (approved):

B0Y3R-AVA (approved):
dev tested

PR #3939, opened 2026-07-01
CP-14363: stuck-funds banner + CCT recovery

atn4z7:
## Description

**Ticket: [CP-14363](https://ava-labs.atlassian.net/browse/CP-14363)**

A cross-chain transfer is a two-step export + import. If the import never broadcasts (app closed mid-flow, network blip), AVAX ends up stranded in atomic memory and mobile had no way to surface or recover it. This adds a shared "stuck funds" banner across portfolio, stake, swap, and notification center that detects stranded funds and lets the user recover them, mirroring core-web's pattern. Recovery runs the Fusion `amountIn=0` import-only quote and broadcasts directly from the banner (no swap detour); the swap screen also accepts a 0 amount for the same recovery. Detection is read-only, so it never prompts hardware wallets.

## Screenshots/Videos
https://github.com/user-attachments/assets/27b28551-9486-41fb-9bc6-8c6c1b3eef75

## Testing
**iOS: 0.0.0.9070
Android: 0.0.0.9071**

- Strand funds first: start a CCT transfer (e.g. C-Chain to P-Chain) and reject the second approval (the import), so AVAX sits stranded in atomic memory.
- Portfolio: open the Portfolio tab and confirm the "Core has detected stuck funds" banner appears below the balance, above the action buttons. Expand it and confirm one row per stranded route (e.g. `0.1 AVAX / C-Chain to P-Chain`) with a Recover button.
- Tap Recover and confirm the import approval opens, approve it, and confirm the import broadcasts and the row clears once it confirms on-chain.
- Confirm the same banner renders on Stake home, the Swap screen (below the title), and the Notifications screen (header).
- Multi-route: strand two routes (e.g. C to P and C to X) and confirm two separate rows.
- Swap screen recovery: on a CCT route (AVAX on C-Chain to AVAX on P-Chain) enter 0; confirm the receive field shows the recovered amount and the submit button reads "Recover", then approve to complete the import.

## Checklist

[CP-14363]: https://ava-labs.atlassian.net/browse/CP-14363?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core-mobile/app/new/features/swap/components/StuckFundsBanner.tsx
+++ b/packages/core-mobile/app/new/features/swap/components/StuckFundsBanner.tsx
@@ -1,5 +1,11 @@
-import React, { useEffect, useState } from 'react'
-import { TouchableOpacity } from 'react-native'
+import React, { useCallback, useEffect, useState } from 'react'
+import { LayoutChangeEvent, TouchableOpacity } from 'react-native'
+import Animated, {
+  Easing,
+  useAnimatedStyle,
+  useSharedValue,
+  withTiming
+} from 'react-native-reanimated'
 import {
   Button,
   Icons,
@@ -21,6 +27,13 @@ import { routeLabel } from '../utils/stuckFundsRoutes'
 
 // Atomic AVAX amounts are denominated in nAVAX (9 decimals).
 const NAVAX_DECIMALS = 9
+
+// Expand/collapse timing, matched to the wallets screen's WalletCard so the
+// motion feels consistent across the app.
+const EXPAND_TIMING = {
+  duration: 300,
+  easing: Easing.bezier(0.25, 1, 0.5, 1)
+}
 
 // Title / amount rows: Inter Regular 16 / 22 per Figma. No k2-alpine variant
 // encodes 16/22 exactly, so we anchor on `body1` (Inter Regular, primary color)
@@ -52,6 +65,35 @@ export const StuckFundsBanner = ({
   const { recover, recoveringKey } = useStuckFundsRecovery()
   const [isFusionServiceReady] = useIsFusionServiceReady()
   const [expanded, setExpanded] = useState(false)
+
+  // Natural height of the rows, measured off an absolutely-positioned layer so
+  // it's the intrinsic content height regardless of the animated wrapper's
+  // clipped height. Kept in a shared value so it drives the animation on the UI
+  // thread. Measured up front (rows always rendered) so the first open animates.
+  const contentHeight = useSharedValue(0)
+  // 0 = collapsed, 1 = expanded — the only time-animated value.
+  const expandProgress = useSharedValue(0)
+
+  const onContentLayout = useCallback(
+    (event: LayoutChangeEvent): void => {
+      contentHeight.value = event.nativeEvent.layout.height
+    },
+    [contentHeight]
+  )
+
+  useEffect(() => {
+    expandProgress.value = withTiming(expanded ? 1 : 0, EXPAND_TIMING)
+  }, [expanded, expandProgress])
+
+  // Wrapper height animates between 0 and the measured content height; overflow
+  // is clipped so the rows are revealed/hidden as it grows/shrinks.
+  const animatedContentStyle = useAnimatedStyle(() => ({
+    height: expandProgress.value * contentHeight.value
+  }))
+
+  const animatedChevronStyle = useAnimatedStyle(() => ({
+    transform: [{ rotate: `${expandProgress.value * 180}deg` }]
+  }))
 
   // Gate on the CCT flag (recovery relies on CCT deps that aren't wired when the
   // flag is off) and hide until Fusion is ready, so Recover is never shown in an
@@ -105,15 +147,20 @@ export const StuckFundsBanner = ({
               )} stuck in atomic memory from incomplete cross-chain transfer${plural}`}
             </Text>
           </View>
-          <Icons.Navigation.ExpandMore
-            color={theme.colors.$textSecondary}
-            style={{ transform: [{ rotate: expanded ? '180deg' : '0deg' }] }}
-          />
+          <Animated.View style={animatedChevronStyle}>
+            <Icons.Navigation.ExpandMore color={theme.colors.$textSecondary} />
+          </Animated.View>
         </View>
       </TouchableOpacity>
 
-      {expanded && (
-        <View>
+      <Animated.View
+        pointerEvents={expanded ? 'auto' : 'none'}
+        style={[animatedContentStyle, { overflow: 'hidden' }]}>
+        {/* Absolutely positioned so onLayout reports the intrinsic row height,
+            independent of the wrapper's animated (clipped) height. */}
+        <View
+          onLayout={onContentLayout}
+          style={{ position: 'absolute', left: 0, right: 0, top: 0 }}>
           {routes.map(route => {
             const key = stuckRouteKey(route)
             const isRecovering = recoveringKey === key
@@ -151,7 +198,7 @@ export const StuckFundsBanner = ({
             )
           })}
         </View>
-      )}
+      </Animated.View>
     </View>
   )
 }
--- a/packages/core-mobile/app/new/features/swap/components/StuckFundsBanner.tsx
+++ b/packages/core-mobile/app/new/features/swap/components/StuckFundsBanner.tsx
@@ -14,6 +14,7 @@ import {
   stuckRouteKey,
   useStuckFundsRecovery
 } from '../hooks/useStuckFundsRecovery'
+import { useIsFusionServiceReady } from '../hooks/useZustandStore'
 import { routeLabel } from '../utils/stuckFundsRoutes'
 
 // Atomic AVAX amounts are denominated in nAVAX (9 decimals).
@@ -46,9 +47,12 @@ export const StuckFundsBanner = ({
   const { theme } = useTheme()
   const { routes, totalNAvax, hasAnyAtomics } = useStuckAtomicFunds()
   const { recover, recoveringKey } = useStuckFundsRecovery()
+  const [isFusionServiceReady] = useIsFusionServiceReady()
   const [expanded, setExpanded] = useState(false)
 
-  if (!hasAnyAtomics) {
+  // Hide until Fusion is ready so Recover is never shown in an unusable state
+  // (recovery builds a Fusion quote + transfer).
+  if (!hasAnyAtomics || !isFusionServiceReady) {
     return null
   }
 
--- a/packages/core-mobile/app/new/features/swap/components/StuckFundsBanner.tsx
+++ b/packages/core-mobile/app/new/features/swap/components/StuckFundsBanner.tsx
@@ -1,0 +1,137 @@
+import React, { useState } from 'react'
+import { TouchableOpacity } from 'react-native'
+import {
+  Button,
+  Icons,
+  Separator,
+  Text,
+  View,
+  useTheme
+} from '@avalabs/k2-alpine'
+import { TokenUnit } from '@avalabs/core-utils-sdk'
+import { useStuckAtomicFunds } from '../hooks/useStuckAtomicFunds'
+import {
+  stuckRouteKey,
+  useStuckFundsRecovery
+} from '../hooks/useStuckFundsRecovery'
+import { routeLabel } from '../utils/stuckFundsRoutes'
+
+// Atomic AVAX amounts are denominated in nAVAX (9 decimals).
+const NAVAX_DECIMALS = 9
+
+// Title / amount rows: Inter Regular 16 / 22 per Figma. No k2-alpine variant
+// encodes 16/22 exactly, so we anchor on `body1` (Inter Regular, primary color)
+// for the font family/weight and override only the two differing dimensions.
+const TITLE_TEXT_SX = { fontSize: 16, lineHeight: 22 } as const
+
+const formatAvax = (amountNAvax: bigint): string =>
+  `${new TokenUnit(amountNAvax, NAVAX_DECIMALS, 'AVAX').toDisplay()} AVAX`
+
+/**
+ * Shared banner surfacing AVAX stranded in atomic memory after an incomplete
+ * cross-chain transfer. Collapsed by default; expands to one row per stranded
+ * route, each with a Recover action that builds an import-only recovery quote
+ * and broadcasts it directly (via useStuckFundsRecovery), surfacing the standard
+ * CCT approval — no swap-screen detour, matching core-web's Recover UX.
+ *
+ * Renders nothing (and reserves no space) when there are no stranded funds, so
+ * callers should NOT wrap it in a spacing container — pass margins via `sx`
+ * instead, which apply only when the banner actually renders.
+ */
+export const StuckFundsBanner = ({
+  sx
+}: {
+  sx?: React.ComponentProps<typeof View>['sx']
+} = {}): JSX.Element | null => {
+  const { theme } = useTheme()
+  const { routes, totalNAvax, hasAnyAtomics } = useStuckAtomicFunds()
+  const { recover, recoveringKey } = useStuckFundsRecovery()
+  const [expanded, setExpanded] = useState(false)
+
+  if (!hasAnyAtomics) {
+    return null
+  }
+
+  const plural = routes.length > 1 ? 's' : ''
+
+  return (
+    <View
+      testID="stuckFundsBanner"
+      sx={{
+        backgroundColor: '$surfaceSecondary',
+        borderRadius: 12,
+        paddingHorizontal: 16,
+        ...sx
+      }}>
+      <TouchableOpacity
+        testID="stuckFundsBanner_toggle"
+        accessibilityRole="button"
+        onPress={() => setExpanded(prev => !prev)}>
+        <View
+          sx={{
+            flexDirection: 'row',
+            alignItems: 'center',
+            justifyContent: 'space-between',
+            gap: 8,
+            paddingVertical: 14
+          }}>
+          <View sx={{ flex: 1 }}>
+            <Text variant="body1" sx={TITLE_TEXT_SX}>
+              Core has detected stuck funds
+            </Text>
+            <Text variant="subtitle2" sx={{ color: '$textSecondary' }}>
+              {`You have ${formatAvax(
+                totalNAvax
+              )} stuck in atomic memory from incomplete cross-chain transfer${plural}`}
+            </Text>
+          </View>
+          <Icons.Navigation.ExpandMore
+            color={theme.colors.$textSecondary}
+            style={{ transform: [{ rotate: expanded ? '180deg' : '0deg' }] }}
+          />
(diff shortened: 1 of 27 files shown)


packages/core-mobile/app/new/features/swap/components/StuckFundsBanner.tsx
```diff
@@ -0,0 +1,141 @@
...
+} = {}): JSX.Element | null => {
+  const { theme } = useTheme()
+  const { routes, totalNAvax, hasAnyAtomics } = useStuckAtomicFunds()
+  const { recover, recoveringKey } = useStuckFundsRecovery()
+  const [isFusionServiceReady] = useIsFusionServiceReady()
+  const [expanded, setExpanded] = useState(false)
+
+  // Hide until Fusion is ready so Recover is never shown in an unusable state
+  // (recovery builds a Fusion quote + transfer).
+  if (!hasAnyAtomics || !isFusionServiceReady) {
```
B0Y3R-AVA: **Important — banner isn't gated by `isFusionAvalancheCctEnabled`, so Recover can be shown but can't succeed when the flag is off.**

This gates only on `hasAnyAtomics` + `isFusionServiceReady`. But `isFusionServiceReady` flips true even with CCT disabled, because `FusionService` always initializes wrap/unwrap (`FusionService.ts:157`) while `cctDependencies` is deliberately left `undefined` when the flag is off (`store/listeners.ts:230`). So a flag-off user still lands on Portfolio, sees the banner, and taps Recover — but the recovery quote/transfer relies on CCT deps that were never wired.

The swap-screen 0-amount path *is* flag-gated (`isAvalancheCctZeroAmountRoute` takes `isAvalancheCctEnabled`), so the two entry points are inconsistent. Suggest gating both the detection query and this banner behind the same flag — that also gives the feature a kill switch, which it currently lacks.

packages/core-mobile/app/new/features/swap/hooks/useStuckAtomicFunds.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  const { data: routes = [] } = useQuery({
+    queryKey: [
+      ReactQueryKeys.FUSION_STUCK_ATOMIC_FUNDS,
+      account,
+      isDeveloperMode,
+      xpAddresses
+    ],
+    enabled: !!account && xpAddresses.length > 0,
+    refetchInterval: STUCK_ATOMIC_FUNDS_REFETCH_INTERVAL,
```
B0Y3R-AVA, line 72: **Important — this polls 6 RPC calls / 60s for every XP user, on the default landing screen, forever.**

`getAllAtomicUTXOs` fans out to 6 `getAtomicUTXOs` calls, and Portfolio (a banner host) is the default screen, so this runs continuously for *all* users regardless of whether they've ever done a CCT. Combined with the missing flag gate (see StuckFundsBanner), that's ongoing load/battery cost for everyone. At minimum reconsider the interval; ideally gate detection behind `isFusionAvalancheCctEnabled`.

Nit: the query key embeds the whole `account` object — RQ deep-hashes so it works, but `account.id` would be a stabler key.

packages/core-mobile/app/services/wallet/AvalancheWalletService.ts
```diff
@@ -50,6 +50,51 @@ class AvalancheWalletService {
...
+    const routes: [Avalanche.ChainIDAlias, Avalanche.ChainIDAlias][] = [
+      ['P', 'C'],
+      ['P', 'X'],
+      ['C', 'P'],
+      ['C', 'X'],
+      ['X', 'C'],
+      ['X', 'P']
+    ]
+
+    return Promise.all(
```
B0Y3R-AVA: **Important — `Promise.all` fail-hides all routes when one RPC rejects.**

A single rejected route (e.g. a flaky `X→P`) rejects the whole promise, and `useStuckAtomicFunds` collapses that to `routes: []` → banner hidden. So one bad chain suppresses genuinely stuck `C→P` funds that loaded fine. Use `Promise.allSettled` and drop only the failed routes.

(Arg order `getAtomicUTXOs(dest, source)` is correct — matches the existing `('P','C')` usage. 👍)

packages/core-mobile/app/new/features/swap/screens/SwapScreen.tsx
```diff
@@ -598,6 +557,18 @@ export const SwapScreen = (): JSX.Element => {
...
+  // "enter an amount" gate is relaxed and the submit button reads "Recover".
+  const allowZeroAmount = isAvalancheCctZeroAmountRoute({
+    isAvalancheCctEnabled,
+    fromToken,
+    toToken
+  })
+
+  // Recovery flow = a CCT route with 0 explicitly entered (empty input is not
+  // treated as recovery, so the button stays "Next" until the user types 0).
+  const isCctRecovery = allowZeroAmount && debouncedFromTokenValue === 0n
```
B0Y3R-AVA, line 676: **Important — CTA relabels to "Recover" on a cleared field, and the nearby comment is inaccurate.**

The comment above claims empty input isn't treated as recovery and the button stays "Next" until the user types 0. But given the known `TokenAmountInput` behavior (empty emits `0n`), clearing the amount on any enabled CCT pair sets `isCctRecovery` true → CTA becomes "Recover" and the quote stream fires at `amountIn=0`. When nothing is stranded, that can drive the `onNoQuotesError` alert path (`SwapContext.tsx:213`) instead of staying inert.

The PR body acknowledges empty-vs-0 as a follow-up, but at minimum fix the misleading comment; better, don't enter recovery mode / suppress the no-quotes alert when the field is empty rather than an explicit typed 0.

packages/core-mobile/app/new/features/swap/hooks/useStuckFundsRecovery.ts
```diff
@@ -0,0 +1,154 @@
...
+          route,
+          account,
+          isDeveloperMode,
+          getNetwork
+        })
+        if (!params) throw new Error('Could not resolve recovery route')
+
+        // amountIn=0 yields an import-only recovery quote; take the first one.
+        const quote = await new Promise<Quote>((resolve, reject) => {
+          const cleanup = subscribeToFirstQuote(params, resolve, () =>
```
B0Y3R-AVA: **Minor — the `subscribeToFirstQuote` cleanup is discarded, so there's no cancel/unmount path (and no timeout).**

The returned cleanup fn is thrown away. If the user leaves the screen after tapping Recover but before the first `quote`/`done`/`error`, the quoter subscription keeps running off-screen and later settles against a dead hook. There's also no timeout — if the SDK stream ever stalls without a terminal event, `recoveringKey` stays set and *all* Recover buttons stay disabled until remount. Capture the cleanup and abort on unmount; consider a timeout.

packages/core-mobile/app/new/features/swap/hooks/useStuckFundsRecovery.ts
```diff
@@ -0,0 +1,154 @@
...
+ */
+export const useStuckFundsRecovery = (): {
+  recover: (route: StuckRoute) => Promise<void>
+  recoveringKey: string | null
+} => {
+  const account = useSelector(selectActiveAccount)
+  const isDeveloperMode = useSelector(selectIsDeveloperMode)
+  const transferGasMarginBps = useSelector(selectFusionTransferGasMarginBps)
+  const { getNetwork } = useNetworks()
+  const { invalidate } = useStuckAtomicFunds()
```
B0Y3R-AVA: **Minor** — this calls `useStuckAtomicFunds()` only to get `invalidate`, which registers a second `select` + refetch-interval consumer of the polling query (RQ dedupes by key, but still). Consider passing `invalidate` in as a param, or calling `queryClient.invalidateQueries({ queryKey: [ReactQueryKeys.FUSION_STUCK_ATOMIC_FUNDS] })` directly.

B0Y3R-AVA (commented):
Adversarial review (Claude + an independent Codex pass, both grounded in source). Solid feature and good test coverage on the utils. A few correctness/robustness items below — the two I'd treat as blocking are the flag-gating gap (tappable-but-broken Recover for flag-off users) and the `Promise.all` fail-hide.

packages/core-mobile/app/new/features/swap/components/StuckFundsBanner.tsx
atn4z7: Good catch. Gated the detection query on `selectIsFusionAvalancheCctEnabled` so it doubles as the kill switch, and put the same flag check on the banner render so the two entry points line up. a2d1ce190

packages/core-mobile/app/new/features/swap/hooks/useStuckAtomicFunds.ts
atn4z7: The flag gate covers this one too. The query only runs when CCT is enabled now, so flag-off users never poll. Also moved the key over to `account.id`. a2d1ce190

packages/core-mobile/app/services/wallet/AvalancheWalletService.ts
atn4z7: Switched to `Promise.allSettled` and drop only the routes that reject, so one flaky chain no longer hides the rest. a2d1ce190

packages/core-mobile/app/new/features/swap/screens/SwapScreen.tsx
atn4z7: You are right, the comment was wrong. Root cause is `TokenAmountInput` emits `0n` for both an empty field and a typed 0, so we cannot tell them apart yet. Fixed the comment to be honest about that for now. Actually distinguishing the two (and skipping the no-quotes alert when the field is empty) needs the input to expose an is-empty signal, so I am tracking that as the follow-up. a2d1ce190

packages/core-mobile/app/new/features/swap/hooks/useStuckFundsRecovery.ts
atn4z7: Done. Capture the cleanup in a ref and tear it down on unmount, plus a 30s timeout so a stalled stream cannot leave every Recover button stuck disabled until remount. a2d1ce190

packages/core-mobile/app/new/features/swap/hooks/useStuckFundsRecovery.ts
atn4z7: Pulled it out into a plain `invalidateStuckAtomicFunds()` that uses the singleton queryClient, so the recovery hook no longer subscribes to the polling query just to grab invalidate. a2d1ce190

packages/core-mobile/app/new/features/swap/screens/SwapScreen.tsx
atn4z7: https://ava-labs.atlassian.net/browse/CP-14669

packages/core-mobile/app/new/features/swap/hooks/useQuoteStreaming.ts
```diff
@@ -70,16 +74,17 @@ export function useQuoteStreaming(
...
       !toToken ||
       !toNetwork ||
-      !fromAmount ||
-      fromAmount <= 0n ||
       !fromAddress ||
-      !toAddress
+      !toAddress ||
+      fromAmount === undefined ||
+      (fromAmount <= 0n && !allowZeroAmount)
     ) {
```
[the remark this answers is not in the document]
atn4z7: Good point. Now rejecting `fromAmount < 0n` outright and only letting an explicit `0n` through when allowZeroAmount is set. e4fb9772e

packages/core-mobile/app/new/features/swap/hooks/useStuckFundsRecovery.ts
```diff
@@ -0,0 +1,187 @@
...
+          if (!cleanup) {
+            reject(new Error('Could not create recovery quote'))
+            return
+          }
+          activeCleanupRef.current = cleanup
+          timeoutId = setTimeout(
+            () => reject(new Error('Recovery quote timed out')),
+            RECOVERY_QUOTE_TIMEOUT_MS
+          )
+        })
```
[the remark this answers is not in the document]
atn4z7: Fixed. Pulled the quote wait into a helper that clears the timeout the moment the quote settles, so a slow approval no longer leaves the timer armed. e4fb9772e

packages/core-mobile/app/new/features/swap/components/StuckFundsBanner.tsx
```diff
@@ -0,0 +1,145 @@
...
+  const { recover, recoveringKey } = useStuckFundsRecovery()
+  const [isFusionServiceReady] = useIsFusionServiceReady()
+  const [expanded, setExpanded] = useState(false)
+
+  // Gate on the CCT flag (recovery relies on CCT deps that aren't wired when the
+  // flag is off) and hide until Fusion is ready, so Recover is never shown in an
+  // unusable state. Detection is already flag-gated, so this is defense-in-depth.
+  if (!isAvalancheCctEnabled || !hasAnyAtomics || !isFusionServiceReady) {
+    return null
+  }
```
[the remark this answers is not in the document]
atn4z7: Yep. Added an effect that collapses the banner whenever it goes hidden, so it re-shows collapsed after a recovery clears the routes. e4fb9772e

B0Y3R-AVA (dismissed):

packages/core-mobile/app/new/features/swap/utils/subscribeToFirstQuote.ts
```diff
@@ -0,0 +1,46 @@
...
+      } else if (event === 'done' || event === 'error') {
+        settled = true
+        onFailed()
+        unsubscribe()
+      }
+    })
+    return () => {
+      settled = true
+      unsubscribe()
+    }
```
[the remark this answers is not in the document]
atn4z7: Added `subscribeToFirstQuote.test.ts` covering all four: resolves on the first quote and unsubscribes, calls onFailed on done/error before any quote, the returned cleanup unsubscribes without firing either callback, and returns undefined when getQuoter returns null or throws. ed3019197

packages/core-mobile/app/services/wallet/AvalancheWalletService.test.ts
```diff
@@ -211,4 +211,40 @@ describe('WalletService', () => {
...
+      getAtomicUTXOs: getAtomicUTXOsMock
+    })
+
+    beforeAll(() => {
+      jest // @ts-ignore
+        .spyOn(AvalancheWalletService, 'getReadOnlySigner')
+        // @ts-ignore
+        .mockImplementation(() => atomicMockWallet())
+    })
+
```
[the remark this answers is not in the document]
atn4z7: Captured the spy and `mockRestore()` it in `afterAll` so it does not leak if tests get appended later. ed3019197

packages/core-mobile/app/new/features/swap/hooks/useStuckFundsRecovery.ts
```diff
@@ -0,0 +1,278 @@
...
+    )
+    FusionService.trackTransfer(
+      transfer,
+      () => undefined,
+      concluded => {
+        clearTimeout(timer)
+        resolve(concluded)
+      }
+    )
+  })
```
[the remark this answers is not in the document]
atn4z7: Good call. `FusionService.trackTransfer` now returns its per-transfer canceller, and `awaitTransferConclusion` calls it when the timeout fires, so the SDK-side tracking stops instead of running on in the background (and the `#trackingCancels` entry is dropped). The main swap-flow caller just ignores the new return value. 4c165d7fc

packages/core-mobile/app/new/features/swap/screens/SwapScreen.tsx
atn4z7: Update: ended up fixing this in-PR rather than deferring it. The root cause was the shared `TokenAmountInput` collapsing an empty field and a typed `0` into the same event (both `value: 0n`, `valueString: "0"`). It now emits `valueString: ""` for an empty field, `TokenInputWidget` forwards that up, and `SwapScreen` maps an empty field to `undefined`. So an empty field stays on "Next" with no quote fired (no more no-quotes alert), and only an explicit typed `0` enters recovery. 5e1e44c2e

packages/core-mobile/app/services/wallet/AvalancheWalletService.ts
```diff
@@ -50,6 +50,66 @@ class AvalancheWalletService {
...
+    // detection query, or one bad chain would hide genuinely stuck funds that
+    // loaded fine on the other routes. Drop only the routes that failed.
+    const settled = await Promise.allSettled(
+      routes.map(async ([dest, source]) => ({
+        dest,
+        source,
+        utxos: await readOnlySigner.getAtomicUTXOs(dest, source)
+      }))
+    )
+
```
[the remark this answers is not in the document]
atn4z7: Added a `Logger.warn` per rejected route (with `source->dest` + reason) before filtering, so intermittent RPC/chain failures are observable on the poll without breaking detection for the routes that loaded. c618569b0

packages/core-mobile/app/new/features/swap/services/FusionService.ts
```diff
@@ -408,6 +409,14 @@ class FusionService implements IFusionService {
...
       })
+
+    // Return a per-transfer canceller so callers that give up early (e.g. a UI
+    // timeout) can stop the SDK-side tracking instead of leaving it running
+    // until the next full `cleanup()`.
+    return () => {
+      cancel()
+      this.#trackingCancels.delete(transfer.id)
+    }
   }
```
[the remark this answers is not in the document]
atn4z7: Good catch. Updated `IFusionService.trackTransfer` to return `() => void` (and added the `onCompleted` param it was already missing), so the interface matches the implementation. c618569b0

packages/core-mobile/app/new/features/swap/hooks/useStuckAtomicFunds.ts
```diff
@@ -0,0 +1,83 @@
...
+      account?.id,
+      isDeveloperMode,
+      xpAddresses
+    ],
+    // Gate detection on the CCT flag: it's the feature's kill switch and stops
+    // the 6-RPC/60s poll for every XP user when CCT recovery isn't enabled.
+    enabled: isAvalancheCctEnabled && !!account && xpAddresses.length > 0,
+    refetchInterval: STUCK_ATOMIC_FUNDS_REFETCH_INTERVAL,
+    refetchIntervalInBackground: false,
+    queryFn: async () => {
```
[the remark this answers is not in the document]
atn4z7: Gated the query on `useIsFusionServiceReady()` too, so it no longer polls during the post-unlock init window while the banner is hidden. Lines up with the banner's hidden-until-Fusion-ready behavior. c618569b0

onghwan (approved):

B0Y3R-AVA (approved):

PR #3943, opened 2026-07-03
CP-14672: fix Android Send amount input not accepting input

atn4z7:
## Description

**Ticket: [CP-14672](https://ava-labs.atlassian.net/browse/CP-14672)**

On Android, the Send amount field was auto-focusing the moment the screen mounted, while the form-sheet and keyboard were still animating in. About 1 in 5 times that left the native input with a broken connection: the cursor blinked but keystrokes never registered, so you couldn't enter an amount until you killed and reopened the app. The fix waits for the sheet's enter transition to finish plus a short settle buffer before focusing on Android, so the input is ready when it receives focus. iOS is unchanged (still auto-focuses on mount).

The platform/timing decision lives in the Send screen; the shared `TokenUnitInput` stays generic and only exposes a `focus()` handle.

## Testing
**iOS: 0.0.0.9107
Android: 0.0.0.9108**

- On an **Android** device, fresh install. From Portfolio tap **Send**, enter a recipient (e.g. Account 2), tap **AVAX** on the token screen, and land on the "How much would you like to send?" screen. The keyboard should come up on its own after a brief moment, and typing `0.01` should register every digit.
- Repeat the flow **8–10 times** (the bug reproduced ~1 in 5). The amount field should accept input every time — no more "cursor shows but nothing types" state.
- Back out and re-enter the amount screen a few times; confirm it still auto-focuses and accepts input on each visit.
- On **iOS**, confirm the amount field still auto-focuses immediately as before (no behavior change).

## Checklist

[CP-14672]: https://ava-labs.atlassian.net/browse/CP-14672?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core-mobile/app/new/features/send/components/SendToken.tsx
+++ b/packages/core-mobile/app/new/features/send/components/SendToken.tsx
@@ -360,7 +360,11 @@ export const SendToken = ({
           onChange={setAmount}
           validateAmount={validateSendAmount}
           disabled={isSending || selectedToken === undefined}
-          autoFocus
+          // iOS auto-focuses immediately on mount. Android focuses later, once
+          // the form-sheet transition and keyboard layout have settled (see the
+          // useAfterScreenEnterTransition call above); focusing on mount there
+          // leaves a broken InputConnection ~1 in 5 times. See CP-14672.
+          autoFocus={Platform.OS === 'ios'}
           maxAmount={maxAmount}
         />
       )}
--- a/packages/core-mobile/app/new/features/send/components/SendToken.tsx
+++ b/packages/core-mobile/app/new/features/send/components/SendToken.tsx
@@ -20,6 +20,8 @@ import {
 import { ScrollScreen } from 'common/components/ScrollScreen'
 import { TRUNCATE_ADDRESS_LENGTH } from 'common/consts/text'
 import { usePrevious } from 'common/hooks/usePrevious'
+import { useAfterScreenEnterTransition } from 'common/hooks/useAfterScreenEnterTransition'
+import { Platform } from 'react-native'
 import { dismissKeyboardIfNeeded } from 'common/utils/dismissKeyboardIfNeeded'
 import { loadAvatar } from 'common/utils/loadAvatar'
 import { stripAddressPrefix } from 'common/utils/stripAddressPrefix'
@@ -31,6 +33,10 @@ import { useSelector } from 'react-redux'
 import { selectSelectedCurrency } from 'store/settings/currency'
 import { useSendContext } from '../context/sendContext'
 import { useSendSelectedToken } from '../store'
+
+// Delay (after the form-sheet transition ends) before auto-focusing the amount
+// input on Android, so the sheet/keyboard layout settles first. See CP-14672.
+const ANDROID_AMOUNT_FOCUS_BUFFER_MS = 500
 
 export const SendToken = ({
   onSend,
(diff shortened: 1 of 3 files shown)


packages/k2-alpine/src/components/TokenUnitInput/TokenUnitInput.tsx
```diff
@@ -144,11 +144,17 @@ export const TokenUnitInput = forwardRef<
...
-        })
-      }
+      // Android intentionally does not auto-focus this amount input. On the New
+      // Architecture (Fabric), inside a react-native-screens form-sheet,
+      // programmatically focusing the field while the screen/keyboard is still
+      // settling leaves a broken InputConnection ~1 in 5 times: the cursor shows
+      // but keystrokes never reach JS, so the amount can't be entered (CP-14672).
+      // A user tap always establishes a healthy connection, so on Android we let
+      // the user tap to focus. iOS is unaffected and keeps auto-focus.
+      if (!autoFocus || Platform.OS !== 'ios') return
```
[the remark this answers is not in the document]
atn4z7: Addressed in a follow-up commit (6594b0bd7). TokenUnitInput is now reverted to generic — no platform or navigation logic and no comment. The platform decision and the clarifying comment moved to the SendToken call site: `autoFocus={Platform.OS === 'ios'}` (iOS auto-focuses on mount), and Android auto-focus is driven by the screen after the enter transition via `useAfterScreenEnterTransition`. So the misleading comment this flagged no longer exists.

packages/core-mobile/app/new/features/send/components/SendToken.tsx
```diff
@@ -73,6 +79,18 @@ export const SendToken = ({
     }
   }, [prevSelectedToken, selectedToken, resetAmount])
 
+  // Android: focus the amount input only after the form-sheet enter transition
+  // has ended AND a short layout/keyboard settle buffer. Focusing during the
+  // still-settling window leaves a broken InputConnection (cursor shows but
+  // keystrokes never reach JS). iOS keeps the input's own autoFocus. See CP-14672.
+  useAfterScreenEnterTransition(
+    () => tokenUnitInputWidgetRef.current?.focus(),
```
bogdandobritoiu: On Android the focus now fires at a fixed moment (transitionEnd + 500ms), but `SendTokenUnitInputWidget` only renders once `tokenBalance` is truthy. If the balance is still loading when the callback runs, `tokenUnitInputWidgetRef.current` is `null` and the `?.focus()` silently does nothing — so on a slow load the keyboard never auto-opens (iOS doesn't have this problem since `autoFocus` fires whenever the widget eventually mounts). Could we retry the focus when `tokenBalance` becomes available, e.g. gate an effect on "transition ended && tokenBalance"?

bogdandobritoiu (dismissed):
Nice, clean fix — right layering (timing policy in the Send screen, widget stays generic) and it reuses the established `useAfterScreenEnterTransition` pattern. One edge case flagged inline.

ruijia1in (dismissed):

packages/core-mobile/app/new/features/send/components/SendToken.tsx
atn4z7: Good catch — addressed in c8e3c089d. I gated the helper's `enabled` on `tokenBalance`:

```ts
enabled: Platform.OS === 'android' && !!tokenBalance
```

Since `enabled` is one of `useAfterScreenEnterTransition`'s effect deps, when the widget mounts (`tokenBalance` becomes truthy) the helper re-arms and focuses via the fallback timer — so a late-mounting widget still auto-opens the keyboard, no separate retry effect needed.

Worth noting `tokenBalance` is a `TokenUnit` object (truthy even at 0 balance), so it's only falsy when no token is selected — i.e. exactly when the widget isn't rendered, so this can't wedge a zero-balance token. In the current flows it was already masked (tokenBalance is derived synchronously from selectedToken, and the token round-trip re-fires the helper via useFocusEffect), but this makes the intent explicit and future-proofs it. Thanks!

bogdandobritoiu (approved):

B0Y3R-AVA (approved):
dev tested

PR #3950, opened 2026-07-07
chore: remove ParaSwap leftovers, unused deps, and JSC fallback

onghwan:
## Description

**Ticket: [CP-]** (cleanup, no ticket)

Dead-code cleanup left over from the ParaSwap → Fusion swap migration, plus removal of unused dependencies and the vestigial JSC engine fallback.

**Swap feature (ParaSwap leftovers)**

- Removed the `destination: SwapSide` state from `SwapContext` and its `setDestination` calls in `SwapScreen`. This was the old ParaSwap/Jupiter exact-in vs exact-out selector; the Fusion quote stream is exact-in only (`fromAmount`) and nothing reads this state anymore.
- Removed the unreachable `handleToAmountChange` handler: the "You receive" widget is `editable={false}`, `TokenAmountInput` only fires `onChange` from user typing (never from programmatic value updates), and percentage buttons only render while the input is focused — which a non-editable input can't be. This also eliminates a latent bug where it would have overwritten the context `amount` (used as the quote stream's `fromAmount`) with the destination amount.
- Made `TokenInputWidget.onAmountChange` optional (all internal call sites were already optional-chained; SwapScreen is the only consumer).
- Removed the now-unused `svc.swap.*` Sentry op names and the `swap` span name.
- Removed the "Powered by ParaSwap" logos from k2-alpine (imports, `PartnerLogos` entries, Storybook story, and both SVG assets) — they were only referenced by the Storybook story.

**Dependencies removed (all verified unused)**

- `@paraswap/sdk` — only imported for the now-removed `SwapSide` enum
- `jsc-android` — never referenced; the gradle `jscFlavor` was a Maven Central artifact, not this npm package
- `react-native-big-list`, `react-native-chart-kit` — zero references anywhere
- `deprecated-react-native-prop-types` — no direct imports; `react-native-popover-view` still gets its own copy (2.3.0) via its own dependency declaration

**Hermes-only engine (both platforms)**

- Android: removed the `jscFlavor` declaration and the `hermesEnabled` conditional from `app/build.gradle` (now links `hermes-android` unconditionally), and removed the `hermesEnabled` property from `gradle.properties`. The React Native Gradle Plugin and expo-modules-core both default to Hermes when the property is absent (verified in their sources).
- iOS: removed the deprecated `:hermes_enabled => true` option from the Podfile (RN 0.85 marks it deprecated; Hermes is the default and JSC has moved to community support). `Podfile.lock` change is the `PODFILE CHECKSUM` only.

No dependencies introduced. No breaking changes — all removed code was unreachable or unreferenced.

## Screenshots/Videos

N/A — no visual or behavioral changes. The swap screen behavior is identical (removed code was unreachable).

## Testing
iOS: 9137
Android: 9138

**Dev Testing**

- `yarn tsc` passes for core-mobile and k2-alpine
- ESLint passes on all changed files
- Swap feature unit tests: 41 suites / 681 tests pass
- Android: `./gradlew :app:dependencies` (internalDebugRuntimeClasspath) resolves successfully with `hermes-android` present and no JSC artifacts, with the `hermesEnabled` property removed
- iOS: `pod install` succeeds; `hermes-engine` still resolved; lockfile delta is the Podfile checksum only
- Suggest verifying a full Bitrise Android + iOS build before merge (the `jsc-android`/gradle changes are the only native-build inputs touched)

**QA Testing**

- Smoke-test the Swap flow (quote fetch, amount entry, Max button, swap execution) — behavior should be unchanged.

## Checklist


atn4z7 (approved):
amazing! thank you for cleaning up!

ruijia1in (approved):

PR #3952, opened 2026-07-07
fix(swap): recurring "You receive" updates + keep recurring quote fresh

ruijia1in:
iOS: 9140
Android: 9141
## Summary

Two related fixes for the recurring-swap flow on the Swap screen (both reproduce on `main`).

### 1. "You receive" never updated in recurring mode
In recurring mode the one-time `activeQuote` is intentionally frozen — `syncDebouncedAmount` early-returns, so `setAmount` never fires and `activeQuote` never refetches. But `applyQuote` sourced the "You receive" amount **only** from `activeQuote`, so changing the pay amount with recurring on left the received amount stuck at whatever it was when the toggle flipped on.

**Fix:** `applyQuote` now sources the received amount from the recurring quote's first-fill estimate (`recurringQuote.data.amountOut`) when recurring is on, and from `activeQuote` otherwise. The decision is a new pure, unit-tested helper `resolveReceiveAmount` that preserves the one-time path's exact semantics (including CCT recovery + "keep previous during refetch"). The receive-field loading spinner now reflects `recurringQuote.isFetching` while recurring is on.

### 2. Recurring quote goes stale → "Quote expired" on submit
Unlike the one-time quote (a live `quoter.subscribe` stream), the recurring quote is a one-shot `useQuery` with `staleTime` but no polling. Sitting on the screen let it age past its server-declared `expiredAt`, so the submit pre-flight (`submitRecurringSwap`) rejected it as `QUOTE_EXPIRED` — the first Next after idling always failed.

The recurring SDK namespace only exposes a one-shot `quote()` (no `subscribe`/stream), so we can't reuse the streaming path. **Fix:** a dynamic `refetchInterval` driven by each quote's own `expiredAt` (tested helper `computeRecurringQuoteRefetchInterval`) — the next refetch is scheduled a small buffer before the current quote expires, keeping it fresh on the server's own cadence. The pre-flight expiry check remains as a backstop.

## Testing
- `resolveReceiveAmount.test.ts` — 10 cases (recurring + one-time paths)
- `useRecurringQuote.test.tsx` — 4 new cases for the interval math
- `yarn tsc` clean · eslint clean on changed files

## Notes
- No ticket number was provided; branch cut fresh off latest `main`.

## Dev testing (verify on a real device)

**Prerequisites**
- Build this branch on a real device (mainnet). Recurring swaps must be enabled for your account (feature flag not blocked) and you need an **eligible Markr pair** — e.g. **USDC ↔ AVAX on Avalanche C-Chain**.
- Have a small amount of the pay token + some AVAX for gas. (Fix #1 needs no submit/funds; fix #2 you can reject at the approval screen.)

### Fix #1 — "You receive" updates in recurring mode (no funds needed)
1. Open **Swap**. Set From = USDC, To = AVAX. Type a pay amount → confirm **You receive** shows a value (one-time mode works).
2. Turn **Make this a recurring swap** ON, then pick a **frequency** and **number of orders** (the recurring quote won't fetch until both are set).
3. Now change the pay amount a few times (e.g. 1 → 2 → 5).
   - ❌ Before fix: **You receive** stays frozen at the old number.
   - ✅ After fix: **You receive** re-computes to the recurring first-fill estimate on each change, with a brief loading spinner while it refetches.
4. Clear the amount → **You receive** clears. Toggle recurring OFF → it returns to the one-time (streaming) value.

### Fix #2 — recurring quote stays fresh (no "Quote expired")
1. With recurring ON and fully configured (frequency + orders + amount), let a recurring quote load.
2. **Leave the Swap screen open and idle for ~2–5 minutes** — do **not** background the app (backgrounding would trigger a focus refetch and mask the bug).
3. Press **Next**.
   - ❌ Before fix: first attempt shows **"Quote expired — please re-confirm and try again"** (or "Recurring swap failed"); a second Next then works.
   - ✅ After fix: **Next** goes straight to the batch-approval / signing screen with no expired error, because the quote auto-refreshed before its `expiredAt`. (You can reject at the approval screen — no need to actually schedule.)

### Regression checks
- One-time (non-recurring) swaps behave exactly as before: You-receive updates live, submit works.
- CCT recovery route (0 amount) still shows its recovery amount and "Recover" button.


B0Y3R-AVA (approved):

onghwan (approved):
dev-tested

PR #3953, opened 2026-07-07
CP-13954: map Ledger 0x6984 to blind-signing guidance for Avalanche L1 signing

B0Y3R-AVA:
## Description

**Ticket: [CP-13954](https://ava-labs.atlassian.net/browse/CP-13954)**

When a Ledger user signs an Avalanche L1 (foreign-EVM) transaction, the Avalanche Ledger app returns a bare `0x6984` status word — its blind-signing refusal path (`view_blindsign_error_show()`): it can't clear-sign a foreign-EVM tx and blind signing isn't enabled, so it rejects with no review screen. That status word is unmapped at every layer (the Avalanche app's own `ApduError` enum, `@avalabs/hw-app-avalanche`'s `LedgerError` enum, and our `LEDGER_ERROR_CODES`), so users hit a dead-end **"Ledger device: UNKNOWN_ERROR (0x6984)"** with only Retry/Cancel — matching Eunji's report on the ticket.

This maps `0x6984` (scoped to the Avalanche Ledger app) to actionable guidance: enable Blind signing in the Avalanche app, and update it via Ledger Live if the setting isn't there.

- Adds `LEDGER_ERROR_CODES.BLIND_SIGN_REQUIRED = '0x6984'` and a shared soft message `LEDGER_BLIND_SIGN_MESSAGE`.
- Maps it in both Ledger error handlers: `handleLedgerError` (throw-side) and `handleLedgerErrorAndShowAlert` (the approval alert UI the user sees).
- Gated to `LedgerAppType.AVALANCHE` so unrelated `0x6984`s keep the generic message.

No new dependencies. This is the reactive fix; a proactive Avalanche-app version gate (mirroring the existing Bitcoin `MAX_BITCOIN_APP_VERSION` pattern) is deferred pending confirmation of the minimum published Avalanche app version with Zondax/Amanda.

## Screenshots/Videos

**Before** (Orange L1 approval, from CP-13954): "Transaction failed — Ledger device: UNKNOWN_ERROR (0x6984)".
**After:** the alert reads "Enable blind signing" with the guidance message. After-screenshots (iOS/Android) to be captured from the Bitrise build during device testing.

## Testing

**Dev Testing**

- Happy path: on a Ledger wallet with the Avalanche app installed and Blind signing **off**, attempt an L1 (e.g. Orange / JUICE) send or swap → the approval alert should now read **"Enable blind signing"** with guidance, instead of "UNKNOWN_ERROR (0x6984)".
- Enable Blind signing in the Avalanche app settings on the device → the same transaction should sign successfully.
- Regression: C-Chain, X, and P signing and non-Avalanche EVM chains are unaffected (they don't reach this branch); a non-Avalanche `0x6984` still shows the generic message.
- Unit tests: `yarn test app/services/wallet/utils.test.ts app/vmModule/ApprovalController/utils.test.ts` (12 tests, covering the Avalanche-only gate on both the throw-side and UI-side handlers, including the real L1 network-resolution path).
- Trigger a Bitrise build and reference it here.
- Move the ticket into the "Testing" column on Jira.

## Checklist

- [x] I have verified the code works (unit tests; end-to-end device verification via Bitrise build + QA)

[CP-13954]: https://ava-labs.atlassian.net/browse/CP-13954?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core-mobile/app/services/wallet/utils.test.ts
+++ b/packages/core-mobile/app/services/wallet/utils.test.ts
@@ -1,7 +1,12 @@
 import { TokenUnit } from '@avalabs/core-utils-sdk'
 import { RpcMethod, TypedData, MessageTypes } from '@avalabs/vm-module-types'
 import { SignTypedDataVersion } from '@metamask/eth-sig-util'
-import { addBufferToCChainBaseFee, getEvmTypedDataVersion } from './utils'
+import { LedgerAppType, LEDGER_BLIND_SIGN_MESSAGE } from 'services/ledger/types'
+import {
+  addBufferToCChainBaseFee,
+  getEvmTypedDataVersion,
+  handleLedgerError
+} from './utils'
 
 describe('addBufferToCChainBaseFee', () => {
   it('should increase base fee by 20%', async () => {
@@ -62,5 +67,27 @@ describe('getEvmTypedDataVersion', () => {
     expect(
       getEvmTypedDataVersion(RpcMethod.SIGN_TYPED_DATA_V1, TYPED_DATA_V1)
     ).toBe(SignTypedDataVersion.V1)
+  })
+})
+
+describe('handleLedgerError', () => {
+  const err = (msg: string): Error => new Error(msg)
+
+  it('maps a bare 0x6984 from the Avalanche app to the blind-sign message', () => {
+    expect(() =>
+      handleLedgerError({
+        error: err('Ledger device: UNKNOWN_ERROR (0x6984)'),
+        appType: LedgerAppType.AVALANCHE
+      })
+    ).toThrow(LEDGER_BLIND_SIGN_MESSAGE)
+  })
+
+  it('does NOT apply the blind-sign message for 0x6984 on a non-Avalanche app', () => {
+    expect(() =>
+      handleLedgerError({
+        error: err('Ledger device: UNKNOWN_ERROR (0x6984)'),
+        appType: LedgerAppType.ETHEREUM
+      })
+    ).not.toThrow()
   })
 })
--- a/packages/core-mobile/app/services/wallet/utils.test.ts
+++ b/packages/core-mobile/app/services/wallet/utils.test.ts
@@ -1,11 +1,6 @@
 import { TokenUnit } from '@avalabs/core-utils-sdk'
-import {
-  NetworkVMType,
-  RpcMethod,
-  TypedData,
-  MessageTypes
-} from '@avalabs/vm-module-types'
-import { Network } from '@avalabs/core-chains-sdk'
+import { RpcMethod, TypedData, MessageTypes } from '@avalabs/vm-module-types'
+import { Network, NetworkVMType } from '@avalabs/core-chains-sdk'
 import { SignTypedDataVersion } from '@metamask/eth-sig-util'
 import { LedgerAppType, LEDGER_BLIND_SIGN_MESSAGE } from 'services/ledger/types'
 import {
--- a/packages/core-mobile/app/services/wallet/utils.test.ts
+++ b/packages/core-mobile/app/services/wallet/utils.test.ts
@@ -1,5 +1,11 @@
 import { TokenUnit } from '@avalabs/core-utils-sdk'
-import { RpcMethod, TypedData, MessageTypes } from '@avalabs/vm-module-types'
+import {
+  NetworkVMType,
+  RpcMethod,
+  TypedData,
+  MessageTypes
+} from '@avalabs/vm-module-types'
+import { Network } from '@avalabs/core-chains-sdk'
 import { SignTypedDataVersion } from '@metamask/eth-sig-util'
 import { LedgerAppType, LEDGER_BLIND_SIGN_MESSAGE } from 'services/ledger/types'
 import {
@@ -89,5 +95,19 @@ describe('handleLedgerError', () => {
         appType: LedgerAppType.ETHEREUM
       })
     ).not.toThrow()
+  })
+
+  it('resolves an L1 network to the Avalanche app and maps 0x6984', () => {
+    const l1Network = {
+      vmName: NetworkVMType.EVM,
+      subnetId: 'orange-subnet',
+      chainId: 999999
+    } as unknown as Network
+    expect(() =>
+      handleLedgerError({
+        error: err('Ledger device: UNKNOWN_ERROR (0x6984)'),
+        network: l1Network
+      })
+    ).toThrow(LEDGER_BLIND_SIGN_MESSAGE)
   })
 })
--- a/packages/core-mobile/app/vmModule/ApprovalController/utils.test.ts
+++ b/packages/core-mobile/app/vmModule/ApprovalController/utils.test.ts
@@ -1,6 +1,5 @@
-import { NetworkVMType } from '@avalabs/vm-module-types'
 import { showAlert } from '@avalabs/k2-alpine'
-import { Network } from '@avalabs/core-chains-sdk'
+import { Network, NetworkVMType } from '@avalabs/core-chains-sdk'
 import LedgerService from 'services/ledger/LedgerService'
 import { LedgerAppType, LEDGER_BLIND_SIGN_MESSAGE } from 'services/ledger/types'
 import { handleLedgerErrorAndShowAlert } from './utils'
@@ -52,6 +51,10 @@ describe('handleLedgerErrorAndShowAlert — 0x6984 on Avalanche L1', () => {
   })
 
   it('does NOT show blind-sign guidance for a non-Avalanche (Ethereum) app', () => {
+    // The gate requires BOTH the expected app (ledgerAppName, from the network
+    // shape below) AND the detected open app (getCurrentAppType) to be
+    // Avalanche; here both are Ethereum. The dedicated detected-app case is
+    // covered by the test below.
     mockLedger.getCurrentAppType.mockReturnValue(LedgerAppType.ETHEREUM)
 
     handleLedgerErrorAndShowAlert({
--- a/packages/core-mobile/app/vmModule/ApprovalController/utils.test.ts
+++ b/packages/core-mobile/app/vmModule/ApprovalController/utils.test.ts
@@ -35,10 +35,17 @@ beforeEach(() => {
   mockLedger.getCurrentAppType.mockReturnValue(LedgerAppType.AVALANCHE)
 })
 
-describe('handleLedgerErrorAndShowAlert — 0x6984 on Avalanche L1', () => {
-  it('shows the blind-sign guidance instead of the raw UNKNOWN_ERROR', () => {
+describe('handleLedgerErrorAndShowAlert — Avalanche blind-sign guidance', () => {
+  it('shows the blind-sign title for the transformed message the real send/swap flow delivers', () => {
+    // Real Avalanche L1 (foreign-EVM) path: signEvmTransaction's catch runs the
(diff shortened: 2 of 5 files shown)


packages/core-mobile/app/vmModule/ApprovalController/utils.test.ts
```diff
@@ -0,0 +1,69 @@
...
+    })
+
+    expect(mockShowAlert).toHaveBeenCalledTimes(1)
+    const arg = mockShowAlert.mock.calls[0][0]
+    expect(arg.title).toBe('Enable blind signing')
+    expect(arg.description).toBe(LEDGER_BLIND_SIGN_MESSAGE)
+  })
+
+  it('does NOT show blind-sign guidance for a non-Avalanche (Ethereum) app', () => {
+    mockLedger.getCurrentAppType.mockReturnValue(LedgerAppType.ETHEREUM)
```
atn4z7: Minor: this mock override is a red herring. The gate in `handleLedgerErrorAndShowAlert` reads `getLedgerAppName(network)`, not `LedgerService.getCurrentAppType()`, so the actual discriminator in this test is the `ethereumNetwork` shape (no `subnetId` -> ETHEREUM). Overriding `getCurrentAppType` here could give false confidence that app-type detection is being exercised. A one-line comment noting the network drives the gate would help.

packages/core-mobile/app/services/wallet/utils.test.ts
```diff
@@ -64,3 +75,39 @@ describe('getEvmTypedDataVersion', () => {
...
+    ).toThrow(LEDGER_BLIND_SIGN_MESSAGE)
+  })
+
+  it('does NOT apply the blind-sign message for 0x6984 on a non-Avalanche app', () => {
+    expect(() =>
+      handleLedgerError({
+        error: err('Ledger device: UNKNOWN_ERROR (0x6984)'),
+        appType: LedgerAppType.ETHEREUM
+      })
+    ).not.toThrow()
```
atn4z7, line 92: Minor: since `handleLedgerError` has no final `else`, this `.not.toThrow()` passes for any unmatched message, so it wouldn't catch a broken gate. Consider asserting it doesn't throw the blind-sign message specifically (e.g. check the thrown message when it should throw, and assert no-throw only as the fall-through) to make the intent tighter.

atn4z7 (dismissed):
Nice PR.

packages/core-mobile/app/vmModule/ApprovalController/utils.ts
```diff
@@ -71,6 +76,12 @@ export const handleLedgerErrorAndShowAlert = ({
   ) {
     description =
       'Ledger is processing another request. Please try again later.'
+  } else if (
+    lowercasedMessage.includes(LEDGER_ERROR_CODES.BLIND_SIGN_REQUIRED) &&
+    ledgerAppName === LedgerAppType.AVALANCHE
+  ) {
```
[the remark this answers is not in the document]
B0Y3R-AVA: Good catch. Fixed in 059f38304 — the branch now also gates on `detectedAppType === LedgerAppType.AVALANCHE` so the guidance only shows when the Avalanche app is actually open, not just expected from the network. Added a test covering the Avalanche-network-but-different-app-open case.

packages/core-mobile/app/vmModule/ApprovalController/utils.test.ts
```diff
@@ -0,0 +1,69 @@
+import { NetworkVMType } from '@avalabs/vm-module-types'
+import { showAlert } from '@avalabs/k2-alpine'
+import { Network } from '@avalabs/core-chains-sdk'
```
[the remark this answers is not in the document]
B0Y3R-AVA: Aligned in edfbd5b36 — the `Network` fixtures now import `NetworkVMType` from `@avalabs/core-chains-sdk` to match `Network`/`getLedgerAppName`. The two enums have identical string values today so there was no behavior divergence, but this removes the future-mismatch risk you flagged.

packages/core-mobile/app/services/wallet/utils.test.ts
```diff
@@ -1,7 +1,18 @@
 import { TokenUnit } from '@avalabs/core-utils-sdk'
-import { RpcMethod, TypedData, MessageTypes } from '@avalabs/vm-module-types'
+import {
+  NetworkVMType,
+  RpcMethod,
+  TypedData,
+  MessageTypes
+} from '@avalabs/vm-module-types'
+import { Network } from '@avalabs/core-chains-sdk'
```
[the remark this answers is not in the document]
B0Y3R-AVA: Aligned in edfbd5b36 — `NetworkVMType` now comes from `@avalabs/core-chains-sdk` (same source as `Network`/`getLedgerAppName`). Values were identical across both enums, so no runtime change; this just keeps the fixture and production enum in sync.

packages/core-mobile/app/vmModule/ApprovalController/utils.test.ts
B0Y3R-AVA: This is resolved rather than clarified: as of 059f38304 the gate now also reads `LedgerService.getCurrentAppType()` (`detectedAppType === LedgerAppType.AVALANCHE`), so the override is a real discriminator, not a red herring. Added a dedicated test (Avalanche network + Ethereum app open → raw error) that exercises it, plus a clarifying comment in edfbd5b36.

packages/core-mobile/app/services/wallet/utils.test.ts
B0Y3R-AVA: The positive test already asserts the specific message (`.toThrow(LEDGER_BLIND_SIGN_MESSAGE)`). For a non-Avalanche app the correct behavior is to not throw at all, so `.not.toThrow()` is actually stricter than `.not.toThrow(LEDGER_BLIND_SIGN_MESSAGE)` — it would fail on any throw, including a regression that surfaced the blind-sign message. Kept as-is for that reason; happy to revisit if you had a different fall-through failure mode in mind.

packages/core-mobile/app/vmModule/ApprovalController/utils.ts
```diff
@@ -71,6 +76,13 @@ export const handleLedgerErrorAndShowAlert = ({
   ) {
     description =
       'Ledger is processing another request. Please try again later.'
+  } else if (
+    lowercasedMessage.includes(LEDGER_ERROR_CODES.BLIND_SIGN_REQUIRED) &&
+    ledgerAppName === LedgerAppType.AVALANCHE &&
+    detectedAppType === LedgerAppType.AVALANCHE
+  ) {
+    title = 'Enable blind signing'
```
atn4z7, line 92: This branch (and its `detectedAppType === AVALANCHE` gate + all three tests in the new test file) is effectively dead for the real L1 send/swap flow, and the intended `"Enable blind signing"` title never appears.

Runtime path for an L1 (foreign-EVM) send/swap:
1. `onApprove` -> `ETH_SEND_TRANSACTION` -> `handlers/ethSendTransaction.ts` -> `WalletService.sign`.
2. `LedgerWallet.signEvmTransaction` catch calls `handleLedgerError({ error, network })`. For an L1 network (`getLedgerAppName -> AVALANCHE`) with a raw `0x6984`, the new **throw-side** branch fires and rewrites the message to `LEDGER_BLIND_SIGN_MESSAGE`, which contains no `0x6984`.
3. `ethSendTransaction`'s catch wraps it: `rpcErrors.internal({ message: 'Failed to sign evm transaction', data: { cause: error } })`, so `data.cause.message` is now the guidance text, not `0x6984`.
4. `ApprovalController` passes that to `handleLedgerErrorAndShowAlert`, which reads `error.data?.cause?.message` -> `LEDGER_BLIND_SIGN_MESSAGE`.
5. `lowercasedMessage.includes('0x6984')` is now **false** -> falls to `else`, so the user gets the guidance text as the description but under the default `"Transaction failed"` title, not `"Enable blind signing"`.

So the fix does reach the user (via the throw-side), but the PR's "After: alert reads 'Enable blind signing'" won't hold for the send/swap path. The three `ApprovalController` unit tests only pass because they call this handler directly with a synthetic raw `0x6984`, which never happens in production once `signEvmTransaction` has run.

Two coherent options, either is fine:
- Drop this alert-side branch and keep only the throw-side mapping (simplest; guidance already reaches the user), or
- Keep the title but match on the `LEDGER_BLIND_SIGN_MESSAGE` constant instead of `0x6984`, and add a test that drives the real chain (raw `0x6984` -> `signEvmTransaction` -> alert) so this regression can't hide.

I verified via core-vm-modules that the module itself doesn't sign here (signing runs inside the app's ApprovalController via `onApprove`), so the transformed message is genuinely what reaches this branch.

atn4z7 (commented):
Follow-up after tracing the full runtime path (incl. core-vm-modules): one reachability issue on the alert-side branch. Fix still works via the throw-side, so not blocking, but the intended title never shows for the real flow. Details inline.

packages/core-mobile/app/vmModule/ApprovalController/utils.ts
B0Y3R-AVA: Confirmed the trace — the alert-side `0x6984` branch was dead for the real send/swap flow. Went with your second option (keep the title) in 503f65a54:

- `handleLedgerErrorAndShowAlert` now matches on `LEDGER_BLIND_SIGN_MESSAGE` (the transformed message that `handleLedgerError` throws and `ethSendTransaction` re-wraps as `data.cause.message`), not `0x6984`, so the "Enable blind signing" title now shows for the real L1 flow. Dropped the alert-side app-type gate since that constant is only ever produced under the throw-side `ledgerAppName === AVALANCHE` gate, so it's inherently Avalanche-scoped.
- Rewrote the tests: the positive test now feeds the `data.cause.message === LEDGER_BLIND_SIGN_MESSAGE` shape the real flow delivers (so a regression here can't hide), and a negative test asserts a bare untransformed `0x6984` keeps default handling.

Left the throw-side (`app/services/wallet/utils.ts`) untouched since it's the actual working fix.

packages/core-mobile/app/vmModule/ApprovalController/utils.ts
```diff
@@ -18,6 +22,7 @@ export const handleLedgerErrorAndShowAlert = ({
   rpcMethod?: RpcMethod
   onRetry: () => void
   onCancel: () => void
+  // eslint-disable-next-line sonarjs/cognitive-complexity
 }): void => {
```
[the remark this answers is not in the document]
B0Y3R-AVA: Checked this against the linter and the suggestion doesn't hold for this codebase: `sonarjs/cognitive-complexity` anchors its report on the arrow-function line (`}): void => {`), not the `export const` declaration. Moving the `eslint-disable-next-line` above the `export const` makes it an unused disable (`eslint-comments/no-unused-disable`) *and* leaves the complexity error unreported/unsuppressed:

```
14:29  warning  'sonarjs/cognitive-complexity' rule is disabled but never reported  eslint-comments/no-unused-disable
26:10  error    Refactor this function to reduce its Cognitive Complexity from 16 to the 15 allowed  sonarjs/cognitive-complexity
```

So the original placement (immediately above `}): void => {`) is the correct one and is what actually applies. Kept it there and added a comment explaining why (503f65a54). `yarn eslint` on the file is now clean.

atn4z7 (approved):

ruijia1in (approved):

PR #3954, opened 2026-07-07
CP-14708: normalize swap quote-stream errors so raw JSON parse errors never reach the UI

ruijia1in:
## Ticket
[CP-14708](https://ava-labs.atlassian.net/browse/CP-14708) — 1.0.34 - Swap JSON parse error

## Problem
Swapping WETH.e → USDC surfaced a raw `JSON Parse error: ...` string in the swap screen, and the user couldn't proceed.

Root cause: the Markr quote stream (Fusion SDK) parses each SSE frame with a bare `JSON.parse(event.data)`. Any non-JSON frame — an empty SSE heartbeat, an HTML/plain-text error page injected by an intermediary proxy/CDN on a 200 response, or a truncated chunk — throws a `SyntaxError`. The quote path (unlike the execution path) rendered `error.message` **verbatim**, so the raw parser string reached the user with no message normalization and no retry mapping.

## Fix (app-side)
Normalize every quote-stream error before it reaches `SwapScreen`:
- `isResponseParseError()` — detects JSON parse failures (matches `SyntaxError` instances plus the Hermes `"JSON Parse error"` and V8/Node `"unexpected token"` / `"unexpected end of json"` / `"is not valid json"` phrasings, so it behaves the same on-device and under Jest).
- `toQuoteDisplayError()` — `FusionQuoteError`s pass through untouched (already curated); parse failures become a friendly, retryable `"Couldn't load a quote. Please try again."`; everything else is run through the existing `getSwapErrorMessage`.
- `shouldRetryWithNextQuote()` now also treats parse errors as retryable.
- `useQuoteStreaming`'s `'error'` handler now runs `toQuoteDisplayError(data)` — a raw parser string can no longer reach the UI.

This is defense-in-depth on the consumer side. The true root-cause fix (guarding the bare `JSON.parse` in the Fusion SDK's Markr quote stream so a bad frame is skipped instead of aborting the whole stream) is being handled separately in a Fusion SDK patch release; this PR is not blocked on it.

## Dev testing (manual QA)
- Open **Swap**, select **WETH.e → USDC** on C-Chain, enter `0.0001`. Confirm a quote loads and you can proceed to review/confirm the swap (verifies the normalization wrapper doesn't regress the happy path).
- Try a few other pairs and amounts and confirm quotes still stream and swaps complete as before.
- Confirm the swap error area never shows a raw `JSON Parse error: ...` string; any quote failure reads as a friendly message (e.g. "Couldn't load a quote. Please try again.").

> Note: the parse-error path is intermittent and backend/proxy-dependent, so it can't be triggered on demand in the UI. The happy-path/regression checks above are the practical manual verification; the error-normalization behavior is covered by unit tests.

[CP-14708]: https://ava-labs.atlassian.net/browse/CP-14708?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core-mobile/app/new/features/swap/utils/fusionErrors.ts
+++ b/packages/core-mobile/app/new/features/swap/utils/fusionErrors.ts
@@ -42,6 +42,16 @@ export const fusionErrors = {
       { reason: 'no-eligible-services' }
     )
   },
+  // The Fusion SDK's Markr quote stream does a bare JSON.parse on each SSE
+  // frame; a non-JSON frame (heartbeat, gateway/proxy error page, truncated
+  // chunk) throws a SyntaxError that would otherwise reach the UI as a raw
+  // "JSON Parse error: ..." string. Surface a friendly, retryable message
+  // instead. See CP-14708.
+  quoteFailed(): FusionQuoteError {
+    return new FusionQuoteError("Couldn't load a quote. Please try again.", {
+      kind: 'provider-specific'
+    })
+  },
 
   // Service / initialisation errors
   serviceNotInitialized(): FusionQuoteError {
@@ -287,11 +297,60 @@ export function isInvalidResponseError(error: unknown): boolean {
 }
 
 /**
+ * Check if error is a JSON/SSE response parse failure.
+ *
+ * The Fusion SDK's Markr `/quote` handler runs a bare `JSON.parse(e.data)` on
+ * every SSE frame, so any non-JSON frame — an empty heartbeat, a plain-text or
+ * HTML gateway/proxy error page, or a truncated chunk — throws a `SyntaxError`.
+ * These are transient/provider-specific: retry with the next quote and never
+ * show the raw parser string to the user. See CP-14708.
+ *
+ * Matches both the Hermes ("JSON Parse error: ...") and V8/Node
+ * ("Unexpected token", "Unexpected end of JSON input", "is not valid JSON")
+ * phrasings so it behaves the same on-device and under Jest.
+ */
+export function isResponseParseError(error: unknown): boolean {
+  if (error instanceof SyntaxError) return true
+  if (!(error instanceof Error)) return false
+
+  const message = error.message.toLowerCase()
+  return (
+    message.includes('json parse error') || // Hermes (on-device)
+    message.includes('unexpected token') || // V8/Node
+    message.includes('unexpected end of json') || // V8/Node (empty/truncated)
+    message.includes('is not valid json') // V8/Node (modern)
+  )
+}
+
+/**
  * Determine if we should retry with next quote
  * Only retry for specific error types
  */
 export function shouldRetryWithNextQuote(error: unknown): boolean {
-  return isGasEstimationError(error) || isInvalidResponseError(error)
+  return (
+    isGasEstimationError(error) ||
+    isInvalidResponseError(error) ||
+    isResponseParseError(error)
+  )
+}
+
+/**
+ * Normalise an error from the quote stream into a user-facing error before it
+ * is rendered on the swap screen. The quote path (unlike the execution path)
+ * renders `error.message` verbatim, so a raw SDK error — most notably the
+ * `JSON.parse` `SyntaxError` from the Markr quote stream — would otherwise be
+ * shown to the user as-is. See CP-14708.
+ *
+ * - `FusionQuoteError`s are already curated (friendly message + metadata the
+ *   swap screen relies on, e.g. `isWarning`), so they pass through untouched.
+ * - Parse failures become a friendly, retryable "couldn't load a quote".
+ * - Anything else is run through `getSwapErrorMessage` for the same
+ *   normalisation the execution path already applies.
+ */
+export function toQuoteDisplayError(error: unknown): Error {
+  if (error instanceof FusionQuoteError) return error
+  if (isResponseParseError(error)) return fusionErrors.quoteFailed()
+  return new FusionQuoteError(getSwapErrorMessage(error))
 }
 
 /**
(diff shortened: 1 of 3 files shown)


B0Y3R-AVA (dismissed):

packages/core-mobile/app/new/features/swap/utils/fusionErrors.ts
```diff
@@ -286,12 +296,61 @@ export function isInvalidResponseError(error: unknown): boolean {
...
+export function isResponseParseError(error: unknown): boolean {
+  if (error instanceof SyntaxError) return true
+  if (!(error instanceof Error)) return false
+
+  const message = error.message.toLowerCase()
+  return (
+    message.includes('json parse error') || // Hermes (on-device)
+    message.includes('unexpected token') || // V8/Node
+    message.includes('unexpected end of json') || // V8/Node (empty/truncated)
+    message.includes('is not valid json') // V8/Node (modern)
```
atn4z7, line 321: where do these come from?

atn4z7 (commented):
Nice PR.

packages/core-mobile/app/new/features/swap/utils/fusionErrors.ts
```diff
@@ -286,12 +296,61 @@ export function isInvalidResponseError(error: unknown): boolean {
...
+ * - `FusionQuoteError`s are already curated (friendly message + metadata the
+ *   swap screen relies on, e.g. `isWarning`), so they pass through untouched.
+ * - Parse failures become a friendly, retryable "couldn't load a quote".
+ * - Anything else is run through `getSwapErrorMessage` for the same
+ *   normalisation the execution path already applies.
+ */
+export function toQuoteDisplayError(error: unknown): Error {
+  if (error instanceof FusionQuoteError) return error
+  if (isResponseParseError(error)) return fusionErrors.quoteFailed()
+  return new FusionQuoteError(getSwapErrorMessage(error))
```
atn4z7, line 353: Nice fix on the quote path. One gap worth closing while we're here: this fallback routes through `getSwapErrorMessage`, which has no parse-error branch, so the same parse error can still leak to the "Swap failed" toast.

I traced it through the SDK. The execution endpoints (`/swap`, `/authorize`) go through `fetchJson`, which guards against non-JSON content-types but not a malformed or truncated body that still carries an `application/json` header. In that case `await response.json()` throws a native `SyntaxError`, it's a 200 so the retry loop doesn't catch it, and it propagates up to `handleSwapError` -> `getSwapErrorMessage`, which falls through to the raw message. So the toast shows `JSON Parse error: ...`, the exact thing this PR kills on the quote screen.

Since you already built `isResponseParseError`, could we add a branch inside `getSwapErrorMessage` so both surfaces are covered?

```ts
if (isResponseParseError(error)) {
  return 'Something went wrong completing the swap. Please try again.'
}
```

Different copy from the quote message on purpose since "Couldn't load a quote" reads oddly for an execution failure. Happy to bikeshed the wording.

If you'd rather keep this PR strictly to the quote screen, that's fine too, but let's leave a note that the execution toast can still surface a raw parse string so it's a deliberate follow-up rather than a surprise later.

packages/core-mobile/app/new/features/swap/utils/fusionErrors.ts
ruijia1in: These are the SyntaxError messages that JSON.parse() throws, which differ by JS engine.

ruijia1in: To add specifics: they're lowercased substrings of the `SyntaxError.message` thrown by `JSON.parse`, matched per engine because the wording differs:

- **`'json parse error'`** → **Hermes** (what we run on-device). Its JSON parser prefixes every failure with `"JSON Parse error: …"`, e.g. `JSON Parse error: Unexpected character: <` (Hermes `JSONLexer`). This is the one that actually bit the user in the ticket.
- **`'unexpected token'` / `'unexpected end of json'` / `'is not valid json'`** → **V8** (Node/Jest, so the unit tests hit this path). Verified on Node 22 / V8 12.4:
  - `""` → `Unexpected end of JSON input`
  - `<html>…502…` → `Unexpected token '<', "<html…" is not valid JSON`
  - `Internal Server Error` → `Unexpected token 'I', … is not valid JSON`

One thing worth calling out: the string list is a **fallback**. The real guarantee is the `error instanceof SyntaxError` check at the top of `isResponseParseError` — `JSON.parse` always throws a genuine `SyntaxError` and the SDK forwards that instance unchanged, so `instanceof` catches every parse failure on-device regardless of wording (it also catches cases the substrings miss, e.g. a truncated object → `Expected ',' or '}' …`). The substrings only matter if the error ever reaches us re-wrapped / prototype-stripped. Happy to add this as a code comment if useful.

packages/core-mobile/app/new/features/swap/utils/fusionErrors.ts
ruijia1in: Verified the trace end-to-end and it's exactly right:

- `fetchJson` only throws `"Expected application/json response."` when the content-type **isn't** JSON (fetch-utilities.js:43); on a valid `application/json` header it does `return await s.json()` (line 44), which throws a native `SyntaxError` on a malformed/truncated body.
- That `.json()` runs *after* the retry wrapper returns, so a 200 isn't retried — it propagates up to `handleSwapError` → `getSwapErrorMessage`, which had no parse branch and fell through to the raw message → `JSON Parse error: ...` in the "Swap failed" toast.

Fixed in `785ff24c3`: added an `isResponseParseError` branch to `getSwapErrorMessage`, using your suggested copy since it reads correctly for an execution failure:

```ts
if (isResponseParseError(actualError ?? error)) {
  return 'Something went wrong completing the swap. Please try again.'
}
```

Checks both the walked SDK error and the raw error, so it catches the SyntaxError whether or not it arrives wrapped. Covered by a new `getSwapErrorMessage` unit test. Both surfaces (quote screen + execution toast) are now covered in this PR.

atn4z7 (approved):

onghwan (approved):

PR #3955, opened 2026-07-07
CP-14187: Restake

onghwan:
## Description

**Ticket: [CP-14187]**

Adds **Restake** to the V2 staking flow, plus stake-type filter chips on the Stake home screen. Mirrors core-web's restake behavior (`getOnRestake` in `StakingNewTable`): only **completed** stakes can be restaked, and the original stake's **node, amount, and duration are reused as-is** — the user lands directly on the review screen with no amount/duration steps in between.

### Entry points (completed stakes only)

- **Restake button** at the bottom of a completed stake card
- **Restake CTA** at the bottom of the Stake details screen

### Flow branches (web parity)

- **Fast Stake tx** → straight to the Fast Stake confirm. The original node gets first refusal: it's re-validated against the Fast Stake criteria (active, uptime ≥ 98%, fee ≤ 2%, capacity, time remaining) via `preferredNodeId`, and auto-selection kicks in only if it no longer qualifies.
- **Delegation tx** → straight to the delegate confirm, which resolves the exact original node (`restakeNodeId`). If the validator has left the active set, an alert is shown and the user is dropped into the node picker — with the original **amount and duration prefilled** through the rest of the flow (mirrors web's `DelegationForm` `initialAmount`/`initialDurationMs` + validator-not-found search fallback).

### Also included

- **Stake-type filter chips** on the Stake home screen: `Fast stakes` and `Delegation` after All/Active/Completed (web's filter set; semantics match web — Delegation is txType-based and includes fast stakes, Fast stakes is escrow-output-based). V1 list is untouched (opt-in flag).
- **Fix**: selecting a chip no longer resets the chip row's horizontal scroll (the FlashList remount workaround was remounting the header; the scroll offset is now preserved and restored).
- **Fix**: tapping Restake on a card no longer also triggers the card's own press (raw touch-event bubbling in `BaseCard` — propagation is stopped at the button wrapper).

### Implementation notes for reviewers

- The add-stake layout seeds the shared stake amount with the minimum on entry; parent effects run **after** child effects on mount, so it would clobber the restake amount. A one-shot `beginRestakeEntry` flag makes the layout skip that seed for restake entries.
- Skipping the amount screen is safe: both confirm screens re-run the delegation-steps computation themselves via `useStakeFundingPreflight`.
- The delegate restake prefill lives in a module-level slot that persists for the modal session (like web's page-level state) and is cleared on the next non-restake flow entry.

## Screenshots/Videos

_TBD_

## Testing
iOS: 9206
Android: 9207

**Dev Testing**

Happy path — Fast Stake restake:

1. Use a wallet (Fuji) that has a **completed** Fast Stake.
2. Go to the Stake tab — the completed card shows a **Restake** button; open the card's details — a **Restake** CTA shows at the bottom.
3. Tap Restake (either entry point).
4. Verify the review screen opens directly with the **original amount** and a duration equal to the original stake's duration (end time = now + original whole-day duration).
5. Verify the node: if the original node still qualifies, its NodeID is shown; otherwise a qualifying node is auto-selected.
6. Slide to stake and verify the delegation succeeds.

Happy path — Delegation restake:

1. Use a wallet (Fuji) that has a **completed** advanced delegation.
2. Tap Restake on the card or in details.
3. Verify the review screen opens directly with the **original node, amount, and duration**.
4. Slide to stake and verify the delegation succeeds.

Edge cases:

- Completed delegation whose validator has **left the active set** → tapping Restake shows a "Node unavailable" alert; OK drops into the node picker (only nodes with enough capacity for the original amount are listed); after picking a node, the amount and duration screens open **prefilled** with the original values.
- Active or pending stakes → no Restake button/CTA anywhere.
- Filter chips: `Fast stakes` shows only fee-escrow stakes, `Delegation` shows all delegation-type stakes (including fast stakes); scrolling the chip row right and selecting a chip keeps the scroll position.
- Abandon a restake mid-flow, then start a normal Fast Stake / Delegate → amounts and durations start fresh (no prefill leakage).

**QA Testing**

- Restake must only be offered on completed stakes and must reuse the original node/amount/duration, matching core-web's behavior on the same wallet.

## Checklist

[CP-14187]: https://ava-labs.atlassian.net/browse/CP-14187?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core-mobile/app/new/features/stake/v2/hooks/useAdvancedReviewSource.ts
+++ b/packages/core-mobile/app/new/features/stake/v2/hooks/useAdvancedReviewSource.ts
@@ -100,13 +100,18 @@ export const useAdvancedReviewSource = (): StakeReviewSource => {
 
   // Restake only: the node's remaining delegation capacity must still cover
   // the original amount (the seeded shared amount). The normal flow enforces
-  // this on the amount screen, which restake skips.
+  // this on the amount screen, which restake skips. Suppressed while the
+  // validators query is (re)fetching: a warm cache serves the node instantly,
+  // and evaluating against that stale snapshot could fire the confirm route's
+  // irreversible redirect-to-picker alert for a node fresh data would pass —
+  // same settle-first treatment the not-found case gets below (the CTA is
+  // already held during the fetch via `isFetching`).
   const [stakeAmount] = useStakeAmount()
   const isRestakeNodeFull = useMemo(() => {
-    if (!isRestake || !node) return false
+    if (!isRestake || !node || isFetchingValidators) return false
     const { maxAmount } = getDelegateNodeLimits(node, isDeveloperMode)
     return stakeAmount.gt(maxAmount)
-  }, [isRestake, node, isDeveloperMode, stakeAmount])
+  }, [isRestake, node, isFetchingValidators, isDeveloperMode, stakeAmount])
 
   // Restake only: the node's own validation period must have room for at
   // least a minimum-duration stake (anchored at now + the same 1-minute
@@ -116,15 +121,16 @@ export const useAdvancedReviewSource = (): StakeReviewSource => {
   // original duration, the stake proceeds with the end time clamped to the
   // validator's — shown on the review — mirroring web's `DelegationForm`,
   // which clamps its prefilled duration to the node's remaining days.
+  // Suppressed while fetching for the same reason as the capacity gate.
   const now = useNow()
   const isRestakeNodeEnding = useMemo(() => {
-    if (!isRestake || !node) return false
+    if (!isRestake || !node || isFetchingValidators) return false
     const { maxEndDate } = getDelegateNodeLimits(node, isDeveloperMode)
     return (
       maxEndDate.getTime() - now <
       MIN_START_TIME_BUFFER_MS + getMinimumStakeDurationMs(isDeveloperMode)
     )
-  }, [isRestake, node, isDeveloperMode, now])
+  }, [isRestake, node, isFetchingValidators, isDeveloperMode, now])
 
   return useMemo<StakeReviewSource>(() => {
     const isRestakeNodeUnusable = isRestakeNodeFull || isRestakeNodeEnding
--- a/packages/core-mobile/app/new/features/stake/v2/hooks/useAdvancedReviewSource.ts
+++ b/packages/core-mobile/app/new/features/stake/v2/hooks/useAdvancedReviewSource.ts
@@ -1,3 +1,5 @@
+import { useLocalSearchParams } from 'expo-router'
+import { useNodes } from 'hooks/earn/useNodes'
 import { useMemo } from 'react'
 import { useSelector } from 'react-redux'
 import { selectIsDelegationFeeBlocked } from 'store/posthog'
@@ -12,6 +14,15 @@ import { useDelegateNodeSelection } from '../store'
 // Hoisted so the memo doesn't allocate a fresh Error on every recompute when no
 // node is selected.
 const NO_NODE_SELECTED_ERROR = new Error('No node selected')
+/**
+ * Sentinel for "the restake target validator has left the active set".
+ * Exported so the delegate confirm route can identity-match it and redirect
+ * to the node picker (web parity) instead of letting the confirm screen
+ * surface its generic no-match alert.
+ */
+export const RESTAKE_NODE_UNAVAILABLE_ERROR = new Error(
+  'The original node is no longer available for delegation'
+)
 
 /**
  * Advanced delegate data source for `StakeConfirmScreen`.
@@ -19,6 +30,13 @@ const NO_NODE_SELECTED_ERROR = new Error('No node selected')
  * Unlike Fast Stake (which resolves a validator server-side), the advanced
  * flow uses the node the user picked on the Node details screen — held in the
  * delegate node-selection store.
+ *
+ * Restake is the exception: `useRestake` navigates straight to the delegate
+ * confirm with a `restakeNodeId` param (the original stake's node), and this
+ * hook resolves it against the current validator set — `isFetching` drives
+ * the confirm screen's loading state, and a node that has left the active set
+ * surfaces as an error (the screen's no-match alert), since restaking reuses
+ * the original node verbatim rather than auto-selecting a substitute.
  *
  * The service fee mirrors Fast Stake's convenience fee, but is gated behind
  * its own `delegation-fee-enabled` flag and paid to a delegate-specific escrow
@@ -28,14 +46,38 @@ const NO_NODE_SELECTED_ERROR = new Error('No node selected')
 export const useAdvancedReviewSource = (): StakeReviewSource => {
   const nodes = useDelegateNodeSelection(state => state.nodes)
   const index = useDelegateNodeSelection(state => state.index)
-  const node = nodes[index]
+  const selectedNode = nodes[index]
+
+  // Only present on the restake path — the normal delegate flow never sets it.
+  const { restakeNodeId } = useLocalSearchParams<{ restakeNodeId?: string }>()
+  const isRestake = restakeNodeId !== undefined && restakeNodeId !== ''
+  const {
+    data: validatorsData,
+    isFetching: isFetchingValidators,
+    error: validatorsError
+  } = useNodes(isRestake)
+
+  const node = useMemo(() => {
+    if (!isRestake) return selectedNode
+    return validatorsData?.validators.find(v => v.nodeID === restakeNodeId)
+  }, [isRestake, selectedNode, validatorsData, restakeNodeId])
 
   const isDeveloperMode = useSelector(selectIsDeveloperMode)
   const isDelegationFeeBlocked = useSelector(selectIsDelegationFeeBlocked)
   const isDelegationFeeEnabled = !isDelegationFeeBlocked
 
-  return useMemo<StakeReviewSource>(
-    () => ({
+  return useMemo<StakeReviewSource>(() => {
+    let error: Error | null = null
+    if (!node) {
+      if (isRestake) {
+        error = isFetchingValidators
+          ? null
+          : validatorsError ?? RESTAKE_NODE_UNAVAILABLE_ERROR
+      } else {
+        error = NO_NODE_SELECTED_ERROR
+      }
+    }
+    return {
       validator: node
         ? {
             nodeID: node.nodeID,
@@ -43,15 +85,21 @@ export const useAdvancedReviewSource = (): StakeReviewSource => {
             delegationFee: node.delegationFee
           }
         : undefined,
-      isFetching: false,
-      error: node ? null : NO_NODE_SELECTED_ERROR,
+      isFetching: isRestake ? isFetchingValidators : false,
+      error,
       feePolicy: isDelegationFeeEnabled
         ? {
             rate: DELEGATION_FEE_RATE,
             recipientAddresses: [getDelegationFeeEscrowAddress(isDeveloperMode)]
           }
         : null
-    }),
-    [node, isDelegationFeeEnabled, isDeveloperMode]
-  )
+    }
+  }, [
+    node,
+    isRestake,
+    isFetchingValidators,
+    validatorsError,
+    isDelegationFeeEnabled,
+    isDeveloperMode
+  ])
 }
--- a/packages/core-mobile/app/new/features/stake/v2/hooks/useAdvancedReviewSource.ts
+++ b/packages/core-mobile/app/new/features/stake/v2/hooks/useAdvancedReviewSource.ts
@@ -1,8 +1,10 @@
 import { useLocalSearchParams } from 'expo-router'
 import { useNodes } from 'hooks/earn/useNodes'
 import { useStakeAmount } from 'hooks/earn/useStakeAmount'
+import { useNow } from 'hooks/time/useNow'
 import { useMemo } from 'react'
 import { useSelector } from 'react-redux'
+import { getMinimumStakeDurationMs } from 'services/earn/utils'
 import { selectIsDelegationFeeBlocked } from 'store/posthog'
 import { selectIsDeveloperMode } from 'store/settings/advanced'
 import {
@@ -12,6 +14,10 @@ import {
 import { StakeReviewSource } from '../types'
 import { useDelegateNodeSelection } from '../store'
 import { getDelegateNodeLimits } from './useSelectedDelegateNodeLimits'
+
+// Matches `useValidateStakingEndTime`'s min-start-time buffer (submission is
+// anchored at now + 1 minute).
+const MIN_START_TIME_BUFFER_MS = 60 * 1000
 
 // Hoisted so the memo doesn't allocate a fresh Error on every recompute when no
 // node is selected.
@@ -35,6 +41,19 @@ export const RESTAKE_NODE_UNAVAILABLE_ERROR = new Error(
  */
 export const RESTAKE_NODE_FULL_ERROR = new Error(
   'The original node no longer has enough delegation capacity'
+)
+/**
+ * Sentinel for "the restake target validator is still active but its own
+ * validation period ends too soon to host even a minimum-duration stake".
+ * The normal flow catches this on the duration screen (`exceedsNodeEndTime`);
+ * restake skips that screen, and `useValidateStakingEndTime` clamps the end
+ * time to the validator's *after* its min-duration adjustment — so without
+ * this gate a sub-minimum stake could reach the chain (which rejects it after
+ * the cross-chain import has already run). Surfaced to the confirm route the
+ * same way as the other restake sentinels.
+ */
+export const RESTAKE_NODE_ENDING_ERROR = new Error(
+  "The original node's remaining validation period is too short"
 )
 
 /**
@@ -89,7 +108,26 @@ export const useAdvancedReviewSource = (): StakeReviewSource => {
     return stakeAmount.gt(maxAmount)
   }, [isRestake, node, isDeveloperMode, stakeAmount])
 
+  // Restake only: the node's own validation period must have room for at
+  // least a minimum-duration stake (anchored at now + the same 1-minute
+  // buffer `useValidateStakingEndTime` uses for the start time). The normal
+  // flow enforces this on the duration screen (`exceedsNodeEndTime`), which
+  // restake skips. When the remaining time covers the minimum but not the
+  // original duration, the stake proceeds with the end time clamped to the
+  // validator's — shown on the review — mirroring web's `DelegationForm`,
+  // which clamps its prefilled duration to the node's remaining days.
+  const now = useNow()
+  const isRestakeNodeEnding = useMemo(() => {
+    if (!isRestake || !node) return false
+    const { maxEndDate } = getDelegateNodeLimits(node, isDeveloperMode)
+    return (
+      maxEndDate.getTime() - now <
+      MIN_START_TIME_BUFFER_MS + getMinimumStakeDurationMs(isDeveloperMode)
+    )
+  }, [isRestake, node, isDeveloperMode, now])
+
   return useMemo<StakeReviewSource>(() => {
+    const isRestakeNodeUnusable = isRestakeNodeFull || isRestakeNodeEnding
     let error: Error | null = null
     if (!node) {
       if (isRestake) {
@@ -99,14 +137,17 @@ export const useAdvancedReviewSource = (): StakeReviewSource => {
       } else {
         error = NO_NODE_SELECTED_ERROR
       }
+    } else if (isRestakeNodeEnding) {
+      error = RESTAKE_NODE_ENDING_ERROR
     } else if (isRestakeNodeFull) {
       error = RESTAKE_NODE_FULL_ERROR
     }
     return {
-      // A capacity-exhausted restake node reads as "no validator" so the
-      // confirm screen can't proceed with it even if the redirect is slow.
+      // A capacity-exhausted / soon-ending restake node reads as "no
+      // validator" so the confirm screen can't proceed with it even if the
+      // redirect is slow.
       validator:
-        node && !isRestakeNodeFull
+        node && !isRestakeNodeUnusable
           ? {
               nodeID: node.nodeID,
               endTime: node.endTime,
@@ -126,6 +167,7 @@ export const useAdvancedReviewSource = (): StakeReviewSource => {
     node,
     isRestake,
     isRestakeNodeFull,
+    isRestakeNodeEnding,
     isFetchingValidators,
     validatorsError,
     isDelegationFeeEnabled,
--- a/packages/core-mobile/app/new/features/stake/v2/hooks/useAdvancedReviewSource.ts
+++ b/packages/core-mobile/app/new/features/stake/v2/hooks/useAdvancedReviewSource.ts
@@ -1,8 +1,9 @@
+import { useQueryClient } from '@tanstack/react-query'
 import { useLocalSearchParams } from 'expo-router'
 import { useNodes } from 'hooks/earn/useNodes'
 import { useStakeAmount } from 'hooks/earn/useStakeAmount'
 import { useNow } from 'hooks/time/useNow'
-import { useMemo } from 'react'
+import { useEffect, useMemo } from 'react'
 import { useSelector } from 'react-redux'
 import { getMinimumStakeDurationMs } from 'services/earn/utils'
 import { selectIsDelegationFeeBlocked } from 'store/posthog'
@@ -18,6 +19,16 @@ import { getDelegateNodeLimits } from './useSelectedDelegateNodeLimits'
 // Matches `useValidateStakingEndTime`'s min-start-time buffer (submission is
 // anchored at now + 1 minute).
 const MIN_START_TIME_BUFFER_MS = 60 * 1000
+
+// Covers the wall-clock drift between this render-time check and the actual
+// submission: `computeDelegateDates` re-anchors the start at submit time
+// (+1 min) and, for minimum-duration stakes, extends the end date by the
+// elapsed time — so a node that clears the gate by only a few seconds can
+// still produce a delegation the chain rejects (past the validator's end, or
+// under the minimum duration from the fresh start) after the C→P import has
+// already run. Five minutes comfortably exceeds any realistic review dwell
+// time; a validator that close to its end is a bad restake target anyway.
+const SUBMIT_DRIFT_MARGIN_MS = 5 * 60 * 1000
 
 // Hoisted so the memo doesn't allocate a fresh Error on every recompute when no
 // node is selected.
@@ -55,6 +66,49 @@ export const RESTAKE_NODE_FULL_ERROR = new Error(
 export const RESTAKE_NODE_ENDING_ERROR = new Error(
   "The original node's remaining validation period is too short"
 )
+/**
+ * Sentinel for "the validators query itself failed" (network / RPC), so the
+ * confirm route can surface a connection-appropriate message instead of the
+ * confirm screen's generic "no node matches your requirements" alert — which
+ * describes a filtering miss, not a dropped connection.
+ */
+export const RESTAKE_NODES_FETCH_FAILED_ERROR = new Error(
+  'Unable to load the validator list for restake'
+)
+
+/**
+ * Maps the source's resolution state to the error surfaced to the confirm
+ * screen / route. Module-level so the branch chain stays out of the hook's
+ * cognitive-complexity budget. A failed fetch means we couldn't even check
+ * the node, so it gets its own sentinel (connection-appropriate copy)
+ * instead of "node unavailable" (which implies we checked).
+ */
+const pickSourceError = ({
+  hasNode,
+  isRestake,
+  isFetchingValidators,
+  validatorsError,
+  isRestakeNodeEnding,
+  isRestakeNodeFull
+}: {
+  hasNode: boolean
+  isRestake: boolean
+  isFetchingValidators: boolean
+  validatorsError: Error | null
+  isRestakeNodeEnding: boolean
+  isRestakeNodeFull: boolean
+}): Error | null => {
+  if (!hasNode) {
+    if (!isRestake) return NO_NODE_SELECTED_ERROR
+    if (isFetchingValidators) return null
+    return validatorsError
+      ? RESTAKE_NODES_FETCH_FAILED_ERROR
+      : RESTAKE_NODE_UNAVAILABLE_ERROR
+  }
+  if (isRestakeNodeEnding) return RESTAKE_NODE_ENDING_ERROR
+  if (isRestakeNodeFull) return RESTAKE_NODE_FULL_ERROR
+  return null
+}
 
 /**
  * Advanced delegate data source for `StakeConfirmScreen`.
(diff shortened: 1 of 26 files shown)


packages/core-mobile/app/new/routes/(signedIn)/(modals)/addStakeV2/_layout.tsx
```diff
@@ -32,8 +33,18 @@ export default function StakeLayoutV2(): JSX.Element {
...
   // `minStakeAmount`) doesn't stomp on the user's typed value — only
   // the value at modal entry seeds the input.
+  //
+  // Restake entries skip the seed entirely: `useRestake` pre-seeds the
+  // original stake's amount before navigating, and this effect runs
+  // *after* any child screen's (parent effects fire last on mount), so
+  // it would otherwise overwrite the restake amount with the minimum.
   const initialAmountRef = useRef(minStakeAmount)
   useEffect(() => {
+    if (takeRestakeEntry()) return
```
B0Y3R-AVA, line 43: I think the restake amount/duration can leak into a normal stake if you don't close the modal in between.

This effect is the only thing that clears `restakePrefill` and resets `stakeAmount`, and it only runs once when the modal mounts. On a restake entry it early-returns before either happens. Since `useRestake` sets `restakePrefill` on every delegate restake (not just the node-gone case), and `initialRouteName="index"` means a deep link to `/addStakeV2/delegate/confirm` keeps the chooser under it, backing out of confirm lands you back on the chooser with the modal still open.

So the repro is: restake a completed delegation whose node is still active, back out to the chooser, then start a normal Delegate (or Fast Stake) without closing the modal. By the time you hit the amount screen, `getRestakePrefill() !== null && !stakeAmount.isZero()` is still true, so it prefills the previous stake's amount, and duration.tsx prefills the previous duration. The fast-stake dial gets seeded with the leftover amount instead of the minimum too. Only fully closing and reopening the modal clears it.

Could we tie the reset to actually entering a sub-flow rather than to the modal mount? e.g. reset when the chooser regains focus, or only honor the prefill when there's a restake param in the current navigation instead of trusting the module-level flag that's still hanging around.

packages/core-mobile/app/new/features/stake/v2/hooks/useRestake.ts
```diff
@@ -0,0 +1,118 @@
...
+        // steps open prefilled with the original stake's values.
+        const config = getStakingConfig(isDeveloperMode)
+        applyDefaultDelegateFilters({
+          minFeePercent:
+            Number(config.MinDelegationFee) / PERMILLION_PER_PERCENT,
+          minStakeDays: Math.floor(
+            Number(config.MinStakeDuration) / SECONDS_PER_DAY
+          )
+        })
+        setRestakePrefill({ durationDays: params.durationDays })
```
B0Y3R-AVA: Looks like the delegate restake skips the node capacity check.

In the normal flow, `DelegateAmountScreen` blocks Next when the amount is over the node's remaining capacity (`stakeAmount.gt(maxAmount)`). Restake goes straight to confirm and skips that screen, and confirm only preflights the balance funding, not the node's delegation capacity. So if the original validator is still active but its remaining capacity has dropped below the original amount since the first stake, you can get all the way to submit on something the normal flow would've stopped, and it just fails on chain.

The fast-stake path doesn't have this issue since it filters `minDelegationCapacity` server-side in `useFastStakeNode`. It's only the delegate side. Maybe re-check the resolved node's remaining capacity against the amount in `useAdvancedReviewSource` and surface it the same way as `RESTAKE_NODE_UNAVAILABLE_ERROR`?

Somewhat related: the node comes from `useNodes`, which has a 5 minute `staleTime`, so the availability/capacity read can be up to 5 minutes stale.

packages/core-mobile/app/new/routes/(signedIn)/(modals)/addStakeV2/_layout.tsx
onghwan: Good catch! Fixed in 13c155875 — `startFastStake`/`startDelegate` now clear restake leftovers and re-seed the minimum amount at flow start. FWIW the chooser does not seem reachable beneath a restake-opened confirm today (no `unstable_settings`, so the modal mounts with confirm as the only route and back dismisses the whole modal), but the module-level flag is now guarded regardless of router behavior.

packages/core-mobile/app/new/features/stake/v2/hooks/useRestake.ts
onghwan: Good catch! Fixed in 13c155875 — `useAdvancedReviewSource` now checks the resolved node's remaining capacity (same `getDelegateNodeLimits` math the amount screen uses) and surfaces it as `RESTAKE_NODE_FULL_ERROR`, which goes through the same alert + picker fallback with web's capacity copy. The 5-min `staleTime` is backstopped by this check plus the picker's own capacity gate.

packages/core-mobile/app/new/features/stake/v2/store.ts
```diff
@@ -167,3 +167,54 @@ export const setDelegateNodeSelection = (
...
+ * the next non-restake modal entry (see the add-stake layout), a
+ * chooser-initiated flow start, or the next restake (overwrite).
+ */
+let restakePrefill: { durationDays: number } | null = null
+
+export const setRestakePrefill = (prefill: { durationDays: number }): void => {
+  restakePrefill = prefill
+}
+
+/** Peeks the active restake prefill (null outside a delegate restake). */
```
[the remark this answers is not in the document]
onghwan: Good catch! Fixed in d3e480b54.

packages/core-mobile/app/new/features/stake/v2/hooks/useRestake.ts
```diff
@@ -0,0 +1,121 @@
...
+        // as the amount/duration prefill, and `useStartStaking` treats its
+        // presence as "restake leftovers pending" — a chooser-initiated flow
+        // clears it and re-seeds the amount so nothing leaks (overwritten by
+        // the next restake, cleared by the layout on a normal modal entry).
+        setRestakePrefill({ durationDays: params.durationDays })
+        // Same end-time derivation web uses: now + the original stake's
+        // whole-day duration.
+        const stakeEndTime = getUnixTime(
+          addDays(new Date(), params.durationDays)
+        )
```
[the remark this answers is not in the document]
onghwan: Good catch on the DST edge! Fixed in d3e480b54 — aligned with core-web instead of switching to second-based arithmetic: `FastStakeReview` derives this as `dayjs().add(days, "day").add(1, "hour")`, so the mobile path now keeps the calendar-day derivation for parity and adds the same +1h slack, which absorbs the spring-forward case (and the confirm-side end-time validation clamps the rest).

packages/core-mobile/app/new/features/stake/v2/screens/StakeHomeScreen.tsx
```diff
@@ -75,6 +81,28 @@ export const StakeHomeScreen = (): JSX.Element => {
...
+    []
+  )
+  const restoreChipScroll = useCallback(() => {
+    if (chipScrollOffsetX.current > 0) {
+      chipScrollRef.current?.scrollTo({
+        x: chipScrollOffsetX.current,
+        animated: false
+      })
+    }
+  }, [])
```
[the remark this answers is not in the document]
onghwan: The ref contract is explicit in dripsy itself: its `ScrollView` is `forwardRef<rScrollView, DripsyScrollViewProps>` where `rScrollView` is React Native's `ScrollView` (dripsy/src/core/components/ScrollView.tsx), and k2's styled wrapper preserves that. So `useRef<RNScrollView>` matches the declared contract with no cast, and if the underlying type ever diverged it would surface as a compile error at the `ref={...}` site rather than a silent runtime mismatch. Casting at the imperative call site would actually weaken that check, so keeping as-is.

packages/core-mobile/app/new/features/stake/v2/screens/StakeDurationScreen.tsx
```diff
@@ -106,13 +117,32 @@ const StakeDurationScreen = ({
...
+    const index = durationsWithDays.findIndex(
+      duration => duration.numberOfDays === initialDurationDays
+    )
+    return index >= 0 ? index : undefined
+  })
+  const [initialCustomEndDate] = useState<UTCDate | undefined>(() =>
+    initialDurationDays !== undefined && initialPresetIndex === undefined
+      ? new UTCDate(addHours(addDays(now, initialDurationDays), 1).getTime())
+      : undefined
+  )
```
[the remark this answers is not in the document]
onghwan: This mirrors core-web exactly — `DelegationForm` computes the same prefill as `dayjs().add(days, "day").add(1, "hour")` (calendar days + 1h slack), and that slack is what absorbs the DST spring-forward case; a fall-back just adds an extra hour, which is harmless. Switching to UTC-ms arithmetic would diverge from web for the same net result, so keeping parity.

packages/core-mobile/app/new/features/stake/v2/hooks/useRestake.ts
```diff
@@ -0,0 +1,124 @@
...
+
+      return () => {
+        beginRestakeEntry()
+        useStakeAmount.setState(new TokenUnit(params.amountNAvax, 9, 'AVAX'))
+        // Set for BOTH branches: the delegate node-gone fallback consumes it
+        // as the amount/duration prefill, and `useStartStaking` treats its
+        // presence as "restake leftovers pending" — a chooser-initiated flow
+        // clears it and re-seeds the amount so nothing leaks (overwritten by
+        // the next restake, cleared by the layout on a normal modal entry).
+        setRestakePrefill({ durationDays: params.durationDays })
```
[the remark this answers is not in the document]
onghwan: Good catch! Added in 66b64345b — `useRestake.test.ts` covers the undefined cases (not completed / missing params / neither fast stake nor delegation) and both branches: navigation URL (incl. `stakeEndTime` + node param) plus the seeded stores (`useStakeAmount`, restake prefill, entry flag, node-selection reset, default filters).

packages/core-mobile/app/new/routes/(signedIn)/(modals)/addStakeV2/_layout.tsx
```diff
@@ -32,8 +33,18 @@ export default function StakeLayoutV2(): JSX.Element {
...
+    // Non-restake entry: drop any prefill left behind by a restake that was
+    // abandoned mid-flow, so a fresh Fast Stake / Delegate session doesn't
+    // inherit the old stake's amount or duration.
+    clearRestakePrefill()
     setStakeAmount(initialAmountRef.current)
   }, [setStakeAmount])
```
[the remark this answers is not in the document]
onghwan: `takeRestakeEntry` and `clearRestakePrefill` are module-level imports, not component-scope values, so `react-hooks/exhaustive-deps` does not require them in the deps array (the rule only tracks values that can change between renders). Verified by running ESLint on this file with the repo config — it passes with zero warnings, so CI stays clean as-is.

packages/core-mobile/app/new/features/stake/v2/hooks/useAdvancedReviewSource.ts
```diff
@@ -28,30 +59,76 @@ const NO_NODE_SELECTED_ERROR = new Error('No node selected')
...
-      isFetching: false,
-      error: node ? null : NO_NODE_SELECTED_ERROR,
+  // Restake only: the node's remaining delegation capacity must still cover
+  // the original amount (the seeded shared amount). The normal flow enforces
+  // this on the amount screen, which restake skips.
+  const [stakeAmount] = useStakeAmount()
+  const isRestakeNodeFull = useMemo(() => {
+    if (!isRestake || !node) return false
+    const { maxAmount } = getDelegateNodeLimits(node, isDeveloperMode)
+    return stakeAmount.gt(maxAmount)
```
B0Y3R-AVA, line 181: I think we're checking the node's capacity here but not its remaining time, and I can't convince myself that's safe.

Restake reuses the original duration and jumps straight to confirm, so we skip the duration screen and never run its `exceedsNodeEndTime` gate. If the original validator is still active but getting close to the end of its own validation period, `useValidateStakingEndTime` clamps our end time down to the validator's end time, and that clamp happens *after* the min-duration check with nothing re-checking it afterwards.

So two cases:
- Validator has less than MinStakeDuration left: we clamp to a sub-minimum duration, the CTA stays enabled (nothing in `isSubmitDisabled` looks at duration), the user slides, and the chain rejects it. Since we've already done the C→P import by that point, that's a stuck-funds situation.
- Validator has more than min but less than the original duration: the stake just quietly locks for less time than we told the user we'd reuse. The confirm screen does show the shorter duration, so this one's less bad, but it still isn't really "reuse the original duration."

This feels like the exact mirror of the capacity check we just added. Could we do the same move here: if the node's remaining time can't cover at least the network minimum (ideally the original duration), treat it as unavailable and drop into the picker the same way `RESTAKE_NODE_FULL_ERROR` does?

packages/core-mobile/app/new/features/stake/v2/hooks/useAdvancedReviewSource.ts
```diff
@@ -28,30 +59,76 @@ const NO_NODE_SELECTED_ERROR = new Error('No node selected')
...
+  const selectedNode = nodes[index]
+
+  // Only present on the restake path — the normal delegate flow never sets it.
+  const { restakeNodeId } = useLocalSearchParams<{ restakeNodeId?: string }>()
+  const isRestake = restakeNodeId !== undefined && restakeNodeId !== ''
+  const {
+    data: validatorsData,
+    isFetching: isFetchingValidators,
+    error: validatorsError
+  } = useNodes(isRestake)
```
B0Y3R-AVA, line 144: Smaller, related thing: `useNodes` shares the `['nodes']` cache with a 5 min staleTime, and the confirm CTA only waits on `validator` being defined, not on `isFetchingValidators`. So right after someone's been in the node picker, a restake can resolve the original node from stale data and let confirm proceed even if it went inactive or full in the meantime. Nothing's at risk since the chain re-validates at submit, but it does widen the window where these gates can wave through something stale (and it'd apply to a remaining-time check too, if we add one). Fine as a follow-up imo, just noting it.

B0Y3R-AVA (commented):
Took another pass focused on the restake path. The leftover-leakage clearing, the capacity check, and the DST end-time handling all hold up under tracing, and the new tests look good. One thing I think we're still missing though, plus a smaller related note. Details inline.

packages/core-mobile/app/new/features/stake/v2/hooks/useAdvancedReviewSource.ts
onghwan: Good catch! Fixed in ba00d49a5 — added `RESTAKE_NODE_ENDING_ERROR`, mirroring the capacity gate: if the node's end time can't fit `getMinimumStakeDurationMs` (anchored at now + the same 1-min buffer `useValidateStakingEndTime` uses), the confirm route drops into the picker with its own alert copy. For the middle case (remaining ≥ min but < original duration) the stake proceeds with the clamped end time shown on the review — that matches web, where `DelegationForm` clamps its prefilled days to `nodeRemainingDays`.

packages/core-mobile/app/new/features/stake/v2/hooks/useAdvancedReviewSource.ts
onghwan: Good catch! Also fixed in ba00d49a5 rather than deferring — `isSubmitDisabled` now includes `isFetchingValidator`, so the CTA stays held during a background validator refetch and only enables after the eligibility gates have re-run on fresh data. Applies to both flows; the only cost is a brief disable while the refetch settles.

packages/core-mobile/app/new/features/stake/v2/hooks/useAdvancedReviewSource.ts
B0Y3R-AVA: Verified the fix in ba00d49a5. The `RESTAKE_NODE_ENDING_ERROR` boundary (`nodeEnd - now < 60s + minStakeDuration`) lines up exactly with the clamp in `useValidateStakingEndTime`, so the sub-minimum case now drops into the picker before confirm. Letting the min-but-shorter-than-original case proceed with the clamped end shown on the review is a fair web-parity call. Good stuff.

packages/core-mobile/app/new/features/stake/v2/hooks/useAdvancedReviewSource.ts
B0Y3R-AVA: Nice, thanks for taking this one now instead of deferring. Adding `isFetchingValidator` to `isSubmitDisabled` closes the stale-cache window for both gates. One cheap follow-up if you feel like it: a unit test around the ending/full eligibility gates, since that boundary is what stands between a near-expiry validator and an on-chain rejection. Not blocking.

B0Y3R-AVA (dismissed):
Re-reviewed after the latest commit. Both things I flagged are addressed and I verified the fixes against the source (the remaining-time gate boundary matches the clamp in useValidateStakingEndTime, and the CTA now holds during a background validator refetch). The rest of the restake path held up under a fairly adversarial pass. LGTM.

packages/core-mobile/app/new/features/stake/v2/hooks/useAdvancedReviewSource.ts
```diff
@@ -28,30 +78,99 @@ const NO_NODE_SELECTED_ERROR = new Error('No node selected')
...
-      error: node ? null : NO_NODE_SELECTED_ERROR,
+  // Restake only: the node's remaining delegation capacity must still cover
+  // the original amount (the seeded shared amount). The normal flow enforces
+  // this on the amount screen, which restake skips.
+  const [stakeAmount] = useStakeAmount()
+  const isRestakeNodeFull = useMemo(() => {
+    if (!isRestake || !node) return false
+    const { maxAmount } = getDelegateNodeLimits(node, isDeveloperMode)
+    return stakeAmount.gt(maxAmount)
+  }, [isRestake, node, isDeveloperMode, stakeAmount])
```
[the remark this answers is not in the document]
onghwan: Good catch! Fixed in 79a111ca7 — the capacity gate now returns false while `isFetchingValidators` is true and re-evaluates once the fetch settles, matching the settle-first treatment the not-found case already had. The CTA stays held during the window via `source.isFetching`.

packages/core-mobile/app/new/features/stake/v2/hooks/useAdvancedReviewSource.ts
```diff
@@ -28,30 +78,99 @@ const NO_NODE_SELECTED_ERROR = new Error('No node selected')
...
+  // which clamps its prefilled duration to the node's remaining days.
+  const now = useNow()
+  const isRestakeNodeEnding = useMemo(() => {
+    if (!isRestake || !node) return false
+    const { maxEndDate } = getDelegateNodeLimits(node, isDeveloperMode)
+    return (
+      maxEndDate.getTime() - now <
+      MIN_START_TIME_BUFFER_MS + getMinimumStakeDurationMs(isDeveloperMode)
+    )
+  }, [isRestake, node, isDeveloperMode, now])
```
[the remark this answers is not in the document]
onghwan: Good catch! Fixed in 79a111ca7 — same suppression applied to the minimum-time-remaining gate: it skips evaluation while `isFetchingValidators` is true and re-runs on fresh data.

B0Y3R-AVA (dismissed):

ruijia1in (dismissed):

packages/core-mobile/app/new/features/stake/v2/hooks/useAdvancedReviewSource.ts
```diff
@@ -28,30 +78,105 @@ const NO_NODE_SELECTED_ERROR = new Error('No node selected')
...
+  // validator's — shown on the review — mirroring web's `DelegationForm`,
+  // which clamps its prefilled duration to the node's remaining days.
+  // Suppressed while fetching for the same reason as the capacity gate.
+  const now = useNow()
+  const isRestakeNodeEnding = useMemo(() => {
+    if (!isRestake || !node || isFetchingValidators) return false
+    const { maxEndDate } = getDelegateNodeLimits(node, isDeveloperMode)
+    return (
+      maxEndDate.getTime() - now <
+      MIN_START_TIME_BUFFER_MS + getMinimumStakeDurationMs(isDeveloperMode)
```
B0Y3R-AVA: This gate blocks when the validator's remaining time is under `buffer + minDuration`, but there's a second time shift at submit that it doesn't account for. For a minimum-duration stake, `computeDelegateDates` (useDelegation.ts:54-61) pushes the end date out by however long passed between render and the slide. So if the original stake was around min duration and the node ends just above this boundary, `validatedStakingEndTime` gets clamped down to the validator's end time and then extended right back past it. The chain rejects the DELEGATE, and by then the C→P import has already run, which is the funds-stuck case.

In the normal flow the duration screen's `exceedsNodeEndTime` check is what saves the user here, but restake skips that screen so this gate is the only thing left. I'd widen it to cover the elapsed extension (block at `<= buffer + minDuration + margin`), or clamp the seeded duration so a min-duration restake can't land on a node that's about to end.

Narrow window, but it's exactly the case this gate was added to prevent.

packages/core-mobile/app/new/features/stake/v2/screens/StakeConfirmScreen.tsx
```diff
@@ -852,6 +852,13 @@ const StakeConfirmScreen = ({
         isSubmitting={isIssueDelegationPending}
         isSubmitDisabled={
           !validator ||
+          // Also hold the CTA during a background validator refetch — the
+          // source can serve a cached validator immediately (`useNodes` has a
+          // 5-min staleTime), and the eligibility gates (restake capacity /
+          // remaining time, fast-stake requalification) only re-run once
+          // fresh data lands. Without this, a stale-cache restake could be
+          // submitted inside that window.
+          isFetchingValidator ||
```
B0Y3R-AVA, line 863: One caveat on this guard: `useNodes` has a 5-minute staleTime, so on a warm cache there's no refetch at all and `isFetchingValidator` stays false the whole time. The capacity/time gates then run against a snapshot that could be up to 5 minutes old, and this line only holds the CTA while an actual refetch is in flight. Behaviorally that's the same as the rest of the flow and the chain is the real backstop, so I'm not worried about correctness. It's the comment I'd tweak: it reads like it closes the stale-cache window, and it doesn't quite. Either soften the wording or kick off a refetch when the screen is entered via restake.

packages/core-mobile/app/new/features/stake/v2/hooks/useAdvancedReviewSource.ts
```diff
@@ -28,30 +78,105 @@ const NO_NODE_SELECTED_ERROR = new Error('No node selected')
...
+  }, [isRestake, node, isFetchingValidators, isDeveloperMode, now])
+
+  return useMemo<StakeReviewSource>(() => {
+    const isRestakeNodeUnusable = isRestakeNodeFull || isRestakeNodeEnding
+    let error: Error | null = null
+    if (!node) {
+      if (isRestake) {
+        error = isFetchingValidators
+          ? null
+          : validatorsError ?? RESTAKE_NODE_UNAVAILABLE_ERROR
```
B0Y3R-AVA: When the validators query fails outright (network or RPC), `node` is undefined and this returns the raw `validatorsError`, which isn't one of the three sentinels. So the confirm route won't redirect to the picker, and `StakeConfirmScreen` shows its "Core was unable to find a node that matches your requirements" alert with Start over → dismissAll. That copy describes a filtering miss, not a dropped connection. The other flows do the same thing so it isn't a regression, but hitting it on airplane mode at the restake entrypoint is confusing. Could route a fetch error to the picker redirect as well, or give it its own retry path.

packages/core-mobile/app/new/routes/(signedIn)/(modals)/addStakeV2/delegate/confirm.tsx
```diff
@@ -1,14 +1,76 @@
...
+  useEffect(() => {
+    if (!isRestakeNodeUnusable) return
+    const nodeId = truncateNodeId(restakeNodeId ?? '')
+    // Capacity copy mirrors core-web's `ReviewDelegationTx` toast.
+    const description = isRestakeNodeFull
+      ? `Validator ${nodeId} is no longer eligible for staking. The capacity has been reached. Please select a different one.`
+      : isRestakeNodeEnding
+      ? `Validator ${nodeId} does not have enough time remaining in its validation period for a new stake. Please select a different one.`
+      : `Validator ${nodeId} is no longer available for delegation. Please select a different one.`
+    showAlert({
```
B0Y3R-AVA, line 62: No visible-latch here the way `StakeConfirmScreen` has one. If a background refetch flips the classification between FULL and ENDING while the alert is already up, the effect deps change and you can stack a second "Node unavailable" alert on top. Very narrow to hit, but a simple boolean guard would rule it out.

packages/core-mobile/app/new/features/stake/v2/utils/getRestakeParams.ts
```diff
@@ -0,0 +1,40 @@
...
+ */
+export const getRestakeParams = (
+  tx: PChainTransaction
+): RestakeParams | undefined => {
+  const { nodeId, startTimestamp, endTimestamp } = tx
+  if (!nodeId || startTimestamp === undefined || endTimestamp === undefined) {
+    return undefined
+  }
+
+  const durationDays = Math.round(
```
B0Y3R-AVA, line 28: Worth flagging that rounding to whole days can drift from the real on-chain duration. Non-minimum stakes aren't day-aligned since the start moves forward at submit, so a 29d11h stake comes back as roughly 29d. That's fine if it's deliberately matching web's whole-day reuse, and the comment says it is. Just want to confirm web actually rounds the same way rather than carrying the exact duration, otherwise restake quietly shaves a few hours off the stake.

packages/core-mobile/app/new/features/stake/v2/screens/StakeHomeScreen.tsx
```diff
@@ -75,6 +81,28 @@ export const StakeHomeScreen = (): JSX.Element => {
...
+  const chipScrollOffsetX = useRef(0)
+  const handleChipScroll = useCallback(
+    (event: NativeSyntheticEvent<NativeScrollEvent>) => {
+      chipScrollOffsetX.current = event.nativeEvent.contentOffset.x
+    },
+    []
+  )
+  const restoreChipScroll = useCallback(() => {
+    if (chipScrollOffsetX.current > 0) {
+      chipScrollRef.current?.scrollTo({
```
B0Y3R-AVA, line 99: `chipScrollRef` is typed as `RNScrollView` but it's attached to k2's `ScrollView`, which is `styled(DripsyScrollView)`. Dripsy usually forwards refs so this probably reaches the native `scrollTo`, but nothing else in the app calls `scrollTo` on this primitive imperatively, so I couldn't confirm it from precedent. If the ref doesn't forward, this no-ops silently and the offset just isn't restored (no crash). Can you sanity-check on a device that the scroll position actually survives a chip tap?

B0Y3R-AVA (commented):
Ran an adversarial pass over this with a second model reviewing in parallel. Good news first: no fund-loss and nothing I'd call a blocking correctness bug. The parts I expected to be shaky actually hold up. I traced the module-level restake state and it can't leak into a normal flow, the funding preflight can't strand the CTA, and the duration screen behaves identically to before on the non-restake path.

Left six inline notes below. Only the first one (the remaining-time gate in `useAdvancedReviewSource`) is something I'd want sorted before merge, since it can let a min-duration restake through to a validator that's about to drop off. The rest are polish or things to sanity-check on device.

packages/core-mobile/app/new/features/stake/v2/utils/getRestakeParams.ts
onghwan: Confirmed it matches web: both consumers of the restake state round to whole days the same way — `FastStakeReview.page.tsx` does `dayjs().add(Math.round(durationMs / MILLISECONDS_PER_DAY), "day")` and `DelegationForm.tsx` computes `durationDays = Math.round(initialDurationMs / MILLISECONDS_PER_DAY)` — so the whole-day reuse here is deliberate parity, not drift. That said, QA has already raised a closely related issue about selected vs actual staking duration ([CP-14723](https://ava-labs.atlassian.net/browse/CP-14723)), so duration handling will be revisited in a follow-up PR under that ticket.

packages/core-mobile/app/new/features/stake/v2/screens/StakeHomeScreen.tsx
onghwan: The ref does reach the native ScrollView: dripsy's `ScrollView` is `forwardRef<rScrollView, DripsyScrollViewProps>` where `rScrollView` is React Native's `ScrollView` (dripsy/src/core/components/ScrollView.tsx), and k2's `styled()` wrapper preserves that ref. Also sanity-checked on a physical device — the chip row keeps its scroll position after tapping a chip (scrolled to the right, selected Delegation, no snap-back).

packages/core-mobile/app/new/features/stake/v2/hooks/useAdvancedReviewSource.ts
onghwan: Good catch! Fixed in 4fd1767e5 — the gate now blocks at `buffer + minDuration + SUBMIT_DRIFT_MARGIN_MS` (5 min), covering both failure directions of the submit-time re-anchor in `computeDelegateDates` (the min-duration extension overshooting the validator's end, and the fresh start undershooting the minimum). A validator that close to its end is a bad restake target anyway, so the wider gate just routes those to the picker.

packages/core-mobile/app/new/features/stake/v2/screens/StakeConfirmScreen.tsx
onghwan: Good catch! Went with both suggestions in 4fd1767e5 — restake entry now invalidates the nodes cache on mount (`useAdvancedReviewSource`), so a fetch is always in flight when this screen opens via restake and the gates re-evaluate against fresh data with the CTA held meanwhile. The comment is also rewritten to state exactly that (and that outside a fetch this is not a staleness guarantee — the chain stays the backstop).

packages/core-mobile/app/new/features/stake/v2/hooks/useAdvancedReviewSource.ts
onghwan: Good catch! Fixed in 4fd1767e5 — a failed validators query now surfaces as its own sentinel (`RESTAKE_NODES_FETCH_FAILED_ERROR`) and the confirm route shows connection-appropriate copy ("We couldn't load the validator list. Please check your connection and try again.") before dropping into the picker, whose own error + retry surface handles a still-failing query.

packages/core-mobile/app/new/routes/(signedIn)/(modals)/addStakeV2/delegate/confirm.tsx
onghwan: Good catch! Fixed in 4fd1767e5 — added a one-shot `hasShownAlertRef` latch so a classification flip during a background refetch can't stack a second alert.

B0Y3R-AVA (dismissed):
Approving. Re-checked the six things I flagged against head (4fd1767e5) and they're all genuinely handled, not just patched over:

- The remaining-time gate now carries a 5-min submit-drift margin, so the min-duration-restake-to-a-near-expiry-validator case can't reach the chain and get rejected after the import. This was the one I actually cared about.
- Restake entry invalidates the nodes cache on mount, so the eligibility gates run on fresh data with the CTA held while it settles. Closed for both the capacity and ending gates.
- Failed validator fetch gets its own sentinel + connection-appropriate copy instead of the generic no-match alert.
- One-shot latch stops the alert stacking on a classification flip.
- Duration rounding is deliberate web parity (confirmed both consumers), with CP-14723 tracking the broader duration work.
- Dripsy ScrollView ref forwarding checked out and you verified the scroll-restore on device.

Nice work on the turnaround. Ship it.

B0Y3R-AVA (approved):

ruijia1in (approved):

PR #3963, opened 2026-07-08
CP-14711: Restore whatwg-fetch as global fetch over broken expo/fetch

onghwan:
## Description

**Ticket: [CP-14711](https://ava-labs.atlassian.net/browse/CP-14711)**

Fixes the "Stake Failed — JSON Parse error: Unexpected character" errors in production ([Sentry issue](https://ava-labs.sentry.io/issues/7560604997/)) by restoring React Native's whatwg-fetch as the global fetch implementation.

**Root cause:** The Expo SDK 56 / RN 0.85 upgrade silently swapped `globalThis.fetch` — Expo's winter runtime installs `expo/fetch` as the default global (`expo/src/winter/runtime.native.ts`). Its new Android response-body handling (gzip/brotli/zstd decompression added in expo/expo#45458) can corrupt payloads, which surfaces as raw JSON parse errors from avalanchejs P-Chain RPC calls during stake delegation. The Sentry issue's first-seen release matches the first RN 0.85 build, and the stack trace points at `expo/src/winter/fetch/FetchResponse.ts` on `index.android.bundle`.

**Fix:**

- Restore `global.fetch = whatwg-fetch` in `index.js` — after the import graph (where the winter runtime installs its fetch) and before `setupDeBankCaching()` (which wraps global fetch). This returns all global-fetch consumers (avalanchejs RPC, vm-modules, third-party libs) to the exact implementation they used before the upgrade.
- Declare `whatwg-fetch@3.6.20` as a direct dependency (previously transitive via react-native) and pin it in root `resolutions`, collapsing the pre-existing 3.6.19/3.6.20 split into a single copy shared with RN's own polyfill.

Explicit `expo/fetch` imports (swap quote streaming via `appCheckStreamingFetch` / `fetchAdapter`) are intentionally untouched — they need ReadableStream support that whatwg-fetch lacks.

**Removal condition:** revert this (index.js assignment + direct dependency + resolutions entry) once the upstream fix (expo/expo#46398, currently unreleased for SDK 56) ships in an SDK 56 patch, or when we upgrade to Expo 57 which includes it.

Pure-JS change — no pod install / native rebuild required.

## Screenshots/Videos

N/A (no UI change). Verified on device via temporary logging:

```
fetch before restore, is whatwg? false   ← winter fetch was active
fetch after restore, is whatwg? true     ← whatwg-fetch restored
```

## Testing

1. Reload the app (Metro restart is enough; no native rebuild)
2. Confirm normal app networking works (portfolio balances, stake list)
3. Run a stake confirm flow — the "JSON Parse error: Unexpected character" alert should no longer occur
4. Confirm swap quote streaming still works (it uses explicit expo/fetch, unaffected)

Note: the corruption is intermittent (Android chunk/compression timing), so the definitive verification is the Sentry issue's event rate dropping to zero on releases containing this fix.

- Smoke test staking (delegate, claim) on Android and iOS
- Smoke test swap including quote streaming
- General networking smoke (portfolio, watchlist, browser)

## Checklist

[CP-14711]: https://ava-labs.atlassian.net/browse/CP-14711?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,8 @@
     "react-native-is-edge-to-edge": "1.2.1",
     "react-native-reanimated": "4.5.1",
     "react-native-worklets": "0.10.1",
-    "react-native-ble-plx": "3.5.1"
+    "react-native-ble-plx": "3.5.1",
+    "whatwg-fetch": "3.6.20"
   },
   "engines": {
     "node": ">=20.18.0",
--- a/packages/core-mobile/index.js
+++ b/packages/core-mobile/index.js
@@ -3,6 +3,7 @@ import { enableFreeze } from 'react-native-screens'
 enableFreeze(true)
 
 import { AppRegistry, LogBox, Platform, UIManager } from 'react-native'
+import { fetch as whatwgFetch } from 'whatwg-fetch'
 import './polyfills'
 import Big from 'big.js'
 import FCMService from 'services/fcm/FCMService'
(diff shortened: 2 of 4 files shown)


atn4z7 (approved):
good finding!!

ruijia1in (approved):

PR #3971, opened 2026-07-10
CP-14726: keep swap form scroll position when editing recurring setup on Android

ruijia1in:
## Description

**Ticket: [CP-14726](https://ava-labs.atlassian.net/browse/CP-14726)**

Keeps the Swap form's scroll position stable while editing the recurring-swap setup on **Android**, so the form no longer snaps back to the "You pay" input.

The custom frequency / number-of-orders controls open a `react-native-dialog` text-input alert (autofocused → keyboard). On Android that alert renders **inline in the swap window** (iOS isolates it in a `FullWindowOverlay`), so the Swap screen's `KeyboardAwareScrollView` reacts to its keyboard and scroll-to-focused behavior and snaps the form to the top — hiding the recurring setup the user was just editing.

Two complementary fixes, both Android-only (iOS is unaffected — it never sees the inline alert's keyboard):

1. **`disableScrollOnKeyboardHide` on the Swap `ScrollScreen`** — suppresses the auto-scroll-to-initial-position when the alert's keyboard hides. Scoped to `Platform.OS === 'android' && recurring.isRecurring` so the plain one-time swap flow (which has no such dialog) keeps its normal restore-on-dismiss behavior.
2. **Blur the focused input before opening the recurring prompts** (`Keyboard.dismiss()` in `promptCustomFrequencyValue` / `promptCustomOrders`) — covers the edge case where the keyboard was already up the whole time (e.g. "You pay" focused): the snap there comes from a keyboard frame/focus change (not a hide), so `disableScrollOnKeyboardHide` alone doesn't catch it. Blurring first leaves `KeyboardAwareScrollView` with no child to scroll back to.

## Screenshots/Videos

_To be added._

## Testing

**QA Testing (Android)**

1. Open Swap, focus **You pay**, enter an amount (keyboard up).
2. Turn on recurring and scroll down to the recurring setup **without dismissing the keyboard**.
3. Tap the custom frequency chip → pick a unit → the number prompt opens; enter a value and Save.
4. Tap the custom number-of-orders chip → enter a value and Save.
5. **Expected:** the form stays put at the recurring setup throughout — it does **not** snap back to the "You pay" input, during or after each prompt.
6. Also verify the general flow (keyboard down before opening a prompt) still behaves — no snap on dismiss.
7. Also verify a **plain one-time swap** (recurring off): focus "You pay", enter an amount, dismiss the keyboard — the form still restores to its normal position (this fix does not change that flow).

**iOS**

- Unchanged: the prompts are isolated in a `FullWindowOverlay`, so scroll position is unaffected.

## Build Numbers

`build-apps-internal` pipeline (iOS + Android) — build **#9226** (commit `515705c`)
→ https://app.bitrise.io/app/7d7ca5af7066e290/pipelines/61bcee5d-7df8-4c32-bb70-33ccce8ca436

- iOS: 9226
- Android: 9227

## Checklist

[CP-14726]: https://ava-labs.atlassian.net/browse/CP-14726?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core-mobile/app/new/features/recurringSwap/components/RecurringDetailsRows.test.tsx
+++ b/packages/core-mobile/app/new/features/recurringSwap/components/RecurringDetailsRows.test.tsx
@@ -113,6 +113,7 @@ jest.mock('../contexts/RecurringSwapContext', () => ({
 
 // Import after mocks are registered.
 
+import { Keyboard, Platform } from 'react-native'
 import {
   RecurringDetailsRows,
   ordersChipsForToken
@@ -268,6 +269,70 @@ describe('<RecurringDetailsRows />', () => {
 
     expect(containsText(json, 'Unlimited')).toBe(true)
     expect(containsText(json, 'Estimated total spend')).toBe(false)
+  })
+})
+
+describe('recurring prompt keyboard handling', () => {
+  const originalOS = Platform.OS
+
+  const setPlatform = (os: 'ios' | 'android'): void => {
+    Object.defineProperty(Platform, 'OS', { configurable: true, value: os })
+  }
+
+  // Invoke the "custom" orders chip, which routes to `promptCustomOrders`
+  // (the guarded `Keyboard.dismiss()` site). The chips only mount once the
+  // "Number of orders" row is expanded, so toggle it open first.
+  const triggerCustomOrders = async (): Promise<void> => {
+    let instance!: renderer.ReactTestRenderer
+    await act(async () => {
+      instance = renderer.create(<RecurringDetailsRows {...baseProps} />)
+    })
+    await act(async () => {
+      const row = instance.root
+        .findAllByProps({ testID: 'recurring_row__orders' })
+        .find(node => typeof node.props.onPress === 'function')
+      expect(row).toBeDefined()
+      row?.props.onPress()
+    })
+    const chips = instance.root
+      .findAllByProps({ testID: 'orders_chips' })
+      .find(node => typeof node.props.onSelect === 'function')
+    expect(chips).toBeDefined()
+    await act(async () => {
+      chips?.props.onSelect('custom')
+    })
+  }
+
+  afterEach(() => {
+    Object.defineProperty(Platform, 'OS', {
+      configurable: true,
+      value: originalOS
+    })
+    jest.restoreAllMocks()
+  })
+
+  it('dismisses the keyboard before opening the prompt on Android', async () => {
+    setPlatform('android')
+    const dismissSpy = jest
+      .spyOn(Keyboard, 'dismiss')
+      .mockImplementation(() => undefined)
+
+    await triggerCustomOrders()
+
+    // Blurs the swap form's focused input so its KeyboardAwareScrollView has no
+    // child to snap back to when the inline alert opens. (CP-14726)
+    expect(dismissSpy).toHaveBeenCalledTimes(1)
+  })
+
+  it('does not dismiss the keyboard on iOS (prompt is isolated in a FullWindowOverlay)', async () => {
+    setPlatform('ios')
+    const dismissSpy = jest
+      .spyOn(Keyboard, 'dismiss')
+      .mockImplementation(() => undefined)
+
+    await triggerCustomOrders()
+
+    expect(dismissSpy).not.toHaveBeenCalled()
   })
 })
 
--- a/packages/core-mobile/app/new/features/recurringSwap/components/RecurringDetailsRows.test.tsx
+++ b/packages/core-mobile/app/new/features/recurringSwap/components/RecurringDetailsRows.test.tsx
@@ -303,6 +303,41 @@ describe('recurring prompt keyboard handling', () => {
     })
   }
 
+  // Drive the Android custom-frequency path: expand the row, pick the custom
+  // chip (opens the Android unit picker), then select a unit — which routes to
+  // `promptCustomFrequencyValue` (the other guarded `Keyboard.dismiss()` site).
+  const triggerCustomFrequencyAndroid = async (): Promise<void> => {
+    let instance!: renderer.ReactTestRenderer
+    await act(async () => {
+      instance = renderer.create(<RecurringDetailsRows {...baseProps} />)
+    })
+    await act(async () => {
+      const row = instance.root
+        .findAllByProps({ testID: 'recurring_row__frequency' })
+        .find(node => typeof node.props.onPress === 'function')
+      expect(row).toBeDefined()
+      row?.props.onPress()
+    })
+    await act(async () => {
+      const chips = instance.root
+        .findAllByProps({ testID: 'frequency_chips' })
+        .find(node => typeof node.props.onSelect === 'function')
+      expect(chips).toBeDefined()
+      chips?.props.onSelect('custom')
+    })
+    // The Android unit-picker dialog is the element exposing both onSelect and
+    // onDismiss; selecting a unit routes to the value prompt.
+    await act(async () => {
+      const dialog = instance.root.findAll(
+        node =>
+          typeof node.props.onSelect === 'function' &&
+          typeof node.props.onDismiss === 'function'
+      )[0]
+      expect(dialog).toBeDefined()
+      dialog?.props.onSelect('day')
+    })
+  }
+
   afterEach(() => {
     Object.defineProperty(Platform, 'OS', {
       configurable: true,
@@ -333,6 +368,17 @@ describe('recurring prompt keyboard handling', () => {
     await triggerCustomOrders()
 
     expect(dismissSpy).not.toHaveBeenCalled()
+  })
+
+  it('dismisses the keyboard on the Android custom-frequency path (unit picker → value prompt)', async () => {
+    setPlatform('android')
+    const dismissSpy = jest
+      .spyOn(Keyboard, 'dismiss')
+      .mockImplementation(() => undefined)
+
+    await triggerCustomFrequencyAndroid()
+
+    expect(dismissSpy).toHaveBeenCalledTimes(1)
   })
 })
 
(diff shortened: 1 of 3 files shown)


packages/core-mobile/app/new/features/recurringSwap/components/RecurringDetailsRows.tsx
```diff
@@ -238,6 +238,13 @@ export function RecurringDetailsRows({
 
   const promptCustomFrequencyValue = useCallback(
     (unit: FrequencyUnit) => {
+      // Android: this alert renders inline in the swap window and autofocuses
+      // its own input. If "You pay" is still focused with the keyboard up, the
+      // swap KeyboardAwareScrollView snaps back to it on the keyboard frame
+      // change (a show/refresh, not a hide — so `disableScrollOnKeyboardHide`
+      // on SwapScreen doesn't cover it). Blur first so KAS has no child to
+      // scroll to; the alert then shows its own keyboard. (CP-14726)
+      if (Platform.OS === 'android') Keyboard.dismiss()
```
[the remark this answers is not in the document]
ruijia1in: Added tests in `RecurringDetailsRows.test.tsx` — expanding the "Number of orders" row and triggering the custom chip, asserting `Keyboard.dismiss` is called once on Android and not on iOS. Done in 6878fa3.

packages/core-mobile/app/new/features/recurringSwap/components/RecurringDetailsRows.test.tsx
```diff
@@ -271,6 +272,70 @@ describe('<RecurringDetailsRows />', () => {
...
+    setPlatform('ios')
+    const dismissSpy = jest
+      .spyOn(Keyboard, 'dismiss')
+      .mockImplementation(() => undefined)
+
+    await triggerCustomOrders()
+
+    expect(dismissSpy).not.toHaveBeenCalled()
+  })
+})
```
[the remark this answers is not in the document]
ruijia1in: Added a test for the frequency flow too: it drives the full Android custom-frequency path (expand row → custom chip → unit picker → select unit → value prompt) and asserts `Keyboard.dismiss` fires once. Done in 061dc5c.

packages/core-mobile/app/new/features/swap/screens/SwapScreen.tsx
```diff
@@ -1628,6 +1628,15 @@ export const SwapScreen = (): JSX.Element => {
...
       shouldAvoidKeyboard
+      // Android only: the custom frequency / number-of-orders inputs open a
+      // react-native-dialog text-input alert (autofocused → keyboard). On
+      // Android that alert renders inline in this window (iOS isolates it in a
+      // FullWindowOverlay), so this KeyboardAwareScrollView sees the keyboard
+      // hide and, by default, auto-scrolls back to its initial position —
+      // snapping the form to the top and hiding the recurring setup the user
+      // was just editing. Keeping the position on keyboard-hide fixes that
+      // without affecting iOS (which never sees the alert's keyboard). (CP-14726)
+      disableScrollOnKeyboardHide={Platform.OS === 'android'}
```
B0Y3R-AVA: **[design / blast-radius]** This disables scroll-restore-on-keyboard-hide for the **entire** `SwapScreen` on Android, for the whole screen lifetime — not just the recurring-edit flow. The "You pay" field auto-focuses on entry (`useAfterScreenEnterTransition` → `fromTokenInputRef.focus()`), so even a plain one-time swap will now keep its shifted position when that keyboard dismisses, instead of restoring. Low harm (short form; focus-scroll via `bottomOffset` is unaffected), but it's broader than the recurring bug it's fixing.

Suggest either (a) QA the ordinary non-recurring keyboard-dismiss path on Android, or (b) scope it to the actual case, e.g. `disableScrollOnKeyboardHide={Platform.OS === 'android' && recurring.isRecurring}`.

(For the record: this lever is *not* redundant with the `Keyboard.dismiss()` fix — it correctly suppresses the restore-snap that fires when the prompt's keyboard hides on Save/Cancel.)

packages/core-mobile/app/new/features/recurringSwap/components/RecurringDetailsRows.tsx
B0Y3R-AVA, line 247: **[minor / consistency]** The repo already has `dismissKeyboardIfNeeded()` (`common/utils/dismissKeyboardIfNeeded.tsx`), which on Android awaits `keyboardDidHide` (500ms fallback) precisely because Android dismissal isn't instantaneous. This uses a bare `Keyboard.dismiss()` that doesn't wait.

I think it's fine as-is — `Keyboard.dismiss()` blurs the focused input synchronously, and the mechanism here ("leave KAS no focused child to scroll to") depends on the blur, not on the hide animation finishing. But two things to watch in QA: a possible keyboard hide→reshow flicker on the "You pay already focused" path, and consistency with the established util. Reuse `dismissKeyboardIfNeeded` if flicker shows up. Applies to both call sites (here and `promptCustomOrders`).

B0Y3R-AVA (commented):
Adversarial review (Claude + an independent Codex pass, grounded against `react-native-keyboard-controller@1.21.12` source). Overall this is a correct, well-commented fix — I verified the prop is genuinely forwarded through `ScrollScreen` and typechecks, and that iOS is untouched. Two inline notes below; only the first is worth a decision before merge. Nice explanatory comments throughout.

packages/core-mobile/app/new/features/swap/screens/SwapScreen.tsx
ruijia1in: Scoped it to the actual case: `disableScrollOnKeyboardHide={Platform.OS === 'android' && recurring.isRecurring}`. The custom frequency / number-of-orders dialogs that trigger the restore-snap only render while recurring is on, so this keeps the plain one-time swap's normal restore-on-dismiss (incl. the auto-focused "You pay" path) untouched. Comment updated to explain the narrower scope.

packages/core-mobile/app/new/features/recurringSwap/components/RecurringDetailsRows.tsx
ruijia1in: Keeping the bare `Keyboard.dismiss()`. As you noted, the mechanism here depends on the synchronous blur (leave KAS no focused child to scroll to), not on the hide animation finishing, so awaiting `keyboardDidHide` via `dismissKeyboardIfNeeded` isn't needed. Will keep an eye on the hide→reshow flicker on the "You pay already focused" path during QA and switch to the util if it shows up.

packages/core-mobile/app/new/features/swap/screens/SwapScreen.tsx
B0Y3R-AVA: Confirmed — this resolves the blast-radius concern. Verified `recurring = useRecurringSwapContext()` is in scope here (`SwapScreen.tsx:466`), and since the frequency/orders dialogs only mount while recurring is on, `recurring.isRecurring` is still `true` at the moment their keyboard hides on Save/Cancel — so the restore-snap is still suppressed exactly where it needs to be, while plain one-time swaps keep their normal restore-on-dismiss. Updated comment reads well too. 👍

B0Y3R-AVA (dismissed):

B0Y3R-AVA (commented):
Adversarial review (Claude skeptic + independent Codex pass, grounded against PR-head source + the shipped keyboard-controller 1.21.12). Functionally correct/merge-safe; the notes are about inert code and misleading rationale, not a runtime bug.

packages/core-mobile/app/new/features/swap/screens/SwapScreen.tsx
```diff
@@ -1730,6 +1730,19 @@ export const SwapScreen = (): JSX.Element => {
...
+      // inputs (rendered only while recurring is on) open a react-native-dialog
+      // text-input alert (autofocused → keyboard). On Android that alert
+      // renders inline in this window (iOS isolates it in a FullWindowOverlay),
+      // so this KeyboardAwareScrollView sees the keyboard hide and, by default,
+      // auto-scrolls back to its initial position — snapping the form to the top
+      // and hiding the recurring setup the user was just editing. Keeping the
+      // position on keyboard-hide fixes that without affecting iOS (which never
+      // sees the alert's keyboard) or the plain one-time swap flow, which has no
+      // such dialog and should keep its normal restore-on-dismiss. (CP-14726)
+      disableScrollOnKeyboardHide={
```
B0Y3R-AVA, line 1743: **`disableScrollOnKeyboardHide` is inert here — it can't change behavior on any platform.**

This same `<ScrollScreen>` sets `enabled={Platform.OS !== 'android'}` above, so KeyboardAwareScrollView is disabled on Android. In `react-native-keyboard-controller@1.21.12`, `maybeScroll` returns `0` immediately when `!enabled`, and `disableScrollOnKeyboardHide`'s only effect is gating that `maybeScroll` on hide (`KeyboardAwareScrollView/index.js:129` and `:346`). So:

- **Android:** KAS is `enabled=false` → `maybeScroll` is a no-op → this prop gates nothing.
- **iOS:** the `Platform.OS === 'android'` conjunct is false → prop is `false` = unchanged default.

The actual snap this PR fixes comes from SwapScreen's *own* `keyboardDidShow` listener (gated on `isYouPayFocusedRef`), not from KAS restore-on-hide — so the comment's rationale is inaccurate. The effective fix is the `Keyboard.dismiss()` in RecurringDetailsRows (it blurs "You pay" → clears that ref → the listener early-returns). Recommend removing this prop (and its comment) to avoid a false safety-net, or rewriting the comment to state it's a defensive no-op on Android.

packages/core-mobile/app/new/features/recurringSwap/components/RecurringDetailsRows.tsx
B0Y3R-AVA, line 247: This is the fix that actually works — but the rationale (here and on the SwapScreen prop) misattributes the snap to KAS restore-on-hide. KAS is disabled on Android (`enabled={Platform.OS !== 'android'}`), so the snap is driven by SwapScreen's `keyboardDidShow` handler gated on `isYouPayFocusedRef`. `Keyboard.dismiss()` blurs "You pay" → `handleYouPayBlur` clears the ref → the handler ignores the alert's autofocus keyboard show. Consider tightening the comment to reflect that mechanism. (Also worth an on-device double-check that the blur reliably lands before the alert's keyboard show — a theoretical ordering concern, likely fine given your QA.)

PR #3980, opened 2026-07-14
CP-14756: fix P-chain burned-amount validation (never ran; unhandled rejection on every P-chain tx)

B0Y3R-AVA:
## Description

**Ticket: [CP-14756](https://ava-labs.atlassian.net/browse/CP-14756)**

Fixes the largest error issue in the core-react-native Sentry project ([CORE-REACT-NATIVE-8Z2](https://ava-labs.sentry.io/issues/CORE-REACT-NATIVE-8Z2), 19.6k events / 1,134 users since Dec 2025) — and more importantly, resurrects a dead safety check.

**The bug:** `validateFee` threw `Missing evm fee data` whenever `evmBaseFeeInNAvax` was `undefined`. The three P-chain call sites (`createExportPTx`, `createImportPTx`, `createAddDelegatorTx`) never passed it — for P-chain txs there is no EVM base fee — so validation threw unconditionally. And because all five call sites invoked it fire-and-forget (`shouldValidateBurnedAmount && this.validateFee({...})`), the rejection was unobserved: every P-chain export/import/delegation logged an unhandled rejection to Sentry, and burned-amount validation **never actually ran** on any path (even the C-chain sites could not block a bad tx, since nothing awaited the result).

**The fix**, following the SDK's own canonical per-VM baseFee sourcing (the compiled validate helper in `@avalabs/core-wallets-sdk`):

- `validateFee` now branches on `unsignedTx.getVM()`:
  - `EVM` (C-chain atomic txs) → requires `evmBaseFeeInNAvax`, as before.
  - `PVM` → uses the new optional `pChainFeePriceInNAvax` (call sites pass `feeState?.price` — the same fee state the tx builder used), falling back to a fresh `provider.getApiP().getFeeState().price` when absent. If that fallback fetch fails, validation is **skipped with a warning** (fail-open): a transient RPC failure must not block staking/claim actions, and substituting a zero price would make any real burn look excessive.
  - other VMs → `baseFee: 0n` (the static validation path ignores it).
- All five call sites now `await` inside `if (shouldValidateBurnedAmount) { ... }`, so an excessive-burn tx fails creation instead of throwing into the void. This matches the intent already pinned by the existing test `'should throw if failed to validate fee'`, which only passed before because its mock threw synchronously.
- **Fee-estimation call sites opt out**: the delegation planner's dummy txs (`computeDelegationSteps/utils.ts` — literally `NodeID-1` and dummy amounts) and `useClaimFees`' preflight `createExportPTx` calls now pass `shouldValidateBurnedAmount: false`, matching the existing convention at `useClaimFees.ts` (import estimation). These txs are never submitted, and a validation rejection inside estimation would be misclassified by the planner as a balance failure. Real issuance paths (`EarnService`, `exportP.ts`, `importP.ts`) keep validation enforced.

No dependencies introduced. Behavior notes:

- Fee validation is now actually enforced on issuance — the existing 50% `feeTolerance` and the fact that validation uses the same `feeState` the builder burned against keep legitimate txs passing.
- The 50% tolerance is generous for P-chain (fees there are deterministic `price × gas`), so this catches catastrophic overburn bugs rather than small ones. Deliberately not tightened in the same PR that revives a dead check — a per-VM tolerance is a possible follow-up.

## Screenshots/Videos

N/A — no UI changes.

## Testing

- `yarn jest app/services/wallet/AvalancheWalletService.test.ts` — 17 tests, including new coverage: the fee-failure test now uses an **async** rejection (proves the `await`), `feeState.price` forwarding, and five direct `validateFee` cases (EVM missing-fee throw, EVM fee used, P-chain caller price used, P-chain fetch fallback, excessive-burn throw).
- `yarn jest app/services/earn app/hooks/earn` — 84 tests pass (delegation flows, claim fees, import/export P).
- On-device happy path: stake on Fuji (Stake → Add Delegation), claim rewards, and confirm no `Missing evm fee data` events in Sentry Spotlight; txs proceed normally.

- Regression pass on staking flows (delegate, claim, cross-chain transfer C↔P). Expected: identical UX; the only new user-visible behavior is that a transaction whose burned fee exceeds the expected fee by >50% now fails with "Excessive burn amount" instead of being submitted.

## Checklist

[CP-14756]: https://ava-labs.atlassian.net/browse/CP-14756?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core-mobile/app/hooks/earn/useClaimFees.ts
+++ b/packages/core-mobile/app/hooks/earn/useClaimFees.ts
@@ -201,6 +201,8 @@ const getExportPFee = async ({
       isTestnet: Boolean(avaxXPNetwork.isTestnet),
       destinationAddress: activeAccount.addressPVM,
       destinationChain: 'C',
+      // fee-estimation only — this tx is never submitted
+      shouldValidateBurnedAmount: false,
       feeState,
       xpAddresses
     })
@@ -245,6 +247,8 @@ const getExportPFee = async ({
       isTestnet: Boolean(avaxXPNetwork.isTestnet),
       destinationAddress: activeAccount.addressPVM,
       destinationChain: 'C',
+      // fee-estimation only — this tx is never submitted
+      shouldValidateBurnedAmount: false,
       feeState,
       xpAddresses
     })
--- a/packages/core-mobile/app/services/earn/computeDelegationSteps/utils.ts
+++ b/packages/core-mobile/app/services/earn/computeDelegationSteps/utils.ts
@@ -57,7 +57,7 @@ export const getDelegationFee = async ({
     endDate: getUnixTime(new Date()) + 60 * 60 * 24 * 30,
     rewardAddress,
     isTestnet,
-    shouldValidateBurnedAmount: true,
+    shouldValidateBurnedAmount: false,
     feeState,
     pFeeAdjustmentThreshold,
     xpAddresses
@@ -264,7 +264,7 @@ export const getImportPFee = async ({
     isTestnet,
     sourceChain: 'C',
     destinationAddress,
-    shouldValidateBurnedAmount: true,
+    shouldValidateBurnedAmount: false,
     feeState,
     xpAddresses
   })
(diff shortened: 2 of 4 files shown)


atn4z7 (approved):
wow, good catch!!

bogdandobritoiu (approved):
dev tested

PR #3983, opened 2026-07-14
CP-13903: [Settings] Add toggle to filter out small UTXOs

B0Y3R-AVA:
## Description

**Ticket: [CP-13903](https://ava-labs.atlassian.net/browse/CP-13903)**

Adds a "Filter out small UTXOs" setting (Advanced settings, default ON) that excludes P-Chain UTXOs smaller than 0.002 AVAX (2,000,000 nAVAX), mirroring the implementation already shipped in Core web and Core extension:

- **Balance layer:** the existing Balance-API field `filterOutDustUtxos` (previously hard-coded `false`) is now driven by the setting, so portfolio totals, token detail, send "available", claimable balance, and staking decision math all respect the toggle server-side. The flag is folded into the React Query balance keys so toggling refetches immediately. Account-discovery requests intentionally keep `false` so dust-only accounts are still detected.
- **Tx-building layer:** a shared `filterOutSmallUtxos` util (threshold `2_000_000n`, AVAX-asset-only, inclusive-keep at the boundary) is applied in P-Chain send tx building (after the existing tx-size cap, matching extension order) and in the swap/CCT `getUtxos` callback. `getAtomicUtxos`, staking/delegation, claim (export/import), and fee simulation are deliberately unfiltered (extension parity; avoids stranding dust in atomic memory and keeps fee-sim consistent with real builds).
- **Rollout:** everything is gated behind a new PostHog flag `filter-small-utxos` (default off = kill switch). Flag off ⇒ behavior byte-identical to today and the settings row is hidden. Persisted via redux migration 30 (VERSION 29→30).
- Analytics: `FilterSmallUtxosToggled { isEnabled }`.

Design doc: `docs/superpowers/specs/2026-07-14-cp13903-filter-small-utxos-design.md` (local, uncommitted).

**Known/accepted limitations (adversarially reviewed, accepted by design):**
- The VM-module fallback balance path (used only when the streaming Balance API fails) does not support the dust flag, so during an API outage a toggle-on user sees dust-inclusive balances while send still filters locally. Documented in the spec as out of scope; self-heals when the API recovers.
- Toggling the setting moves the balance query key; cache-only readers now fall back to the previous flag variant's cache for the seconds until refetch (`getCachedBalancesWithFlagFallback`), rather than transiently seeing empty data.
- Pre-existing send-flow races surfaced during review were ticketed separately: [CP-14762](https://ava-labs.atlassian.net/browse/CP-14762), [CP-14763](https://ava-labs.atlassian.net/browse/CP-14763).

No new dependencies. Not breaking; migration 30 default-fills the new setting.

## Screenshots/Videos

TODO before marking ready: iOS/Android screenshots of the Advanced settings toggle (Figma: Core-Mobile-Redesign-2025, node 16584-11892).

## Testing

iOS: 9260
Android: 9261

Happy path:
1. Enable the `filter-small-utxos` PostHog flag (and `everything`).
2. Settings → Advanced settings → verify "Filter out small UTXOs" row appears above Quick swaps, toggle ON by default, caption warns totals may be inaccurate.
3. On a wallet with P-Chain dust (<0.002 AVAX UTXOs): toggle OFF → P-Chain balance increases after refetch; toggle ON → decreases.
4. P-Chain send: with toggle ON, build/send succeeds and consumes no dust UTXOs; max send matches displayed available.

Edge cases:
- Flag OFF in PostHog: row hidden, balances identical to production behavior regardless of the persisted setting.
- Upgrade path: install a build from `main`, then this branch — persisted settings migrate (VERSION 30) with the setting defaulting ON.
- Toggle flip while the Meld off-ramp or a connected dapp (`wallet_getNetworkState`) is active: no empty token lists (cache fallback covers the refetch window).
- Cross-chain swap (CCT) with dust: dust excluded from source UTXO set; stuck-funds import unaffected.

- AC1: toggle exists in Advanced settings and persists across app restarts.
- AC2: warning caption shown: "Improves loading performance by removing UTXOs with a value less than 0.002 AVAX from the wallet. Total balances may be inaccurate."
- AC3: with toggle ON, dust value is excluded from portfolio P-Chain balance, send available/max, claimable balance, staking flows, and swap; with toggle OFF everything includes dust.
- Regression: account import/discovery still finds accounts that hold only dust; staking/claim flows unaffected; testnet mode behaves consistently.

## Checklist

[CP-13903]: https://ava-labs.atlassian.net/browse/CP-13903?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CP-14762]: https://ava-labs.atlassian.net/browse/CP-14762?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core-mobile/app/new/features/portfolio/hooks/useAccountBalances.ts
+++ b/packages/core-mobile/app/new/features/portfolio/hooks/useAccountBalances.ts
@@ -12,6 +12,7 @@ import { useXPAddresses } from 'hooks/useXPAddresses/useXPAddresses'
 import { selectWalletById } from 'store/wallet/slice'
 import { getXpubXPIfAvailable } from 'utils/getAddressesFromXpubXP/getAddressesFromXpubXP'
 import { useOnlineStatus } from 'common/hooks/useOnlineStatus'
+import { selectIsFilterSmallUtxosActive } from 'store/settings/advanced/filterSmallUtxosActive'
 import * as store from '../store'
 
 /**
@@ -26,14 +27,19 @@ const staleTime = 20_000
  */
 const refetchInterval = __DEV__ ? 30_000 : 5_000
 
-export const balanceKey = (account: Account | undefined, network?: Network[]) =>
+export const balanceKey = (
+  account: Account | undefined,
+  network: Network[] | undefined,
+  filterOutDustUtxos: boolean
+) =>
   [
     ReactQueryKeys.ACCOUNT_BALANCE,
     account?.id,
     network
       ?.map(n => n.chainId)
       .sort()
-      .join(',')
+      .join(','),
+    filterOutDustUtxos
   ] as const
 
 /**
@@ -64,6 +70,7 @@ export function useAccountBalances(
   const currency = useSelector(selectSelectedCurrency)
   const { xpAddresses } = useXPAddresses(account)
   const wallet = useSelector(selectWalletById(account?.walletId ?? ''))
+  const filterOutDustUtxos = useSelector(selectIsFilterSmallUtxosActive)
 
   const isNotReady = !account || enabledNetworks.length === 0 || !wallet
 
@@ -77,7 +84,7 @@ export function useAccountBalances(
     refetch: refetchFn
   } = useQuery({
     // eslint-disable-next-line @tanstack/query/exhaustive-deps
-    queryKey: balanceKey(account, enabledNetworks),
+    queryKey: balanceKey(account, enabledNetworks, filterOutDustUtxos),
     enabled,
     refetchInterval: options?.refetchInterval ?? refetchInterval,
     staleTime,
@@ -96,9 +103,10 @@ export function useAccountBalances(
         currency: currency.toLowerCase(),
         xpAddresses,
         xpub,
+        filterOutDustUtxos,
         onBalanceLoaded: balance => {
           queryClient.setQueryData(
-            balanceKey(account, enabledNetworks),
+            balanceKey(account, enabledNetworks, filterOutDustUtxos),
             (prev: AdjustedNormalizedBalancesForAccount[] | undefined) => {
               if (!prev) return [balance]
               const filtered = prev.filter(p => p.chainId !== balance.chainId)
--- a/packages/core-mobile/app/new/features/portfolio/hooks/useAccountBalances.ts
+++ b/packages/core-mobile/app/new/features/portfolio/hooks/useAccountBalances.ts
@@ -1,4 +1,4 @@
-import { useQuery, useQueryClient } from '@tanstack/react-query'
+import { QueryClient, useQuery, useQueryClient } from '@tanstack/react-query'
 import { ReactQueryKeys } from 'consts/reactQueryKeys'
 import { useCallback, useMemo } from 'react'
 import { useSelector } from 'react-redux'
@@ -41,6 +41,37 @@ export const balanceKey = (
       .join(','),
     filterOutDustUtxos
   ] as const
+
+/**
+ * Cache read with flag fallback. When the small-UTXO filter setting (or its
+ * PostHog gate) flips, `balanceKey` changes and the new key stays empty until
+ * the next refetch lands. Cache-only readers (meld off-ramp,
+ * wallet_getNetworkState) should serve the previous flag variant's data for
+ * that window rather than nothing — momentarily stale dust totals beat an
+ * empty token list. An exact-key hit (including a legitimately empty array
+ * from a completed fetch) always wins; the fallback is read-only and never
+ * written back.
+ */
+export const getCachedBalancesWithFlagFallback = ({
+  client,
+  account,
+  networks,
+  filterOutDustUtxos
+}: {
+  client: QueryClient
+  account: Account | undefined
+  networks: Network[] | undefined
+  filterOutDustUtxos: boolean
+}): AdjustedNormalizedBalancesForAccount[] | undefined => {
+  const exact = client.getQueryData(
+    balanceKey(account, networks, filterOutDustUtxos)
+  ) as AdjustedNormalizedBalancesForAccount[] | undefined
+  if (exact !== undefined) return exact
+
+  return client.getQueryData(
+    balanceKey(account, networks, !filterOutDustUtxos)
+  ) as AdjustedNormalizedBalancesForAccount[] | undefined
+}
 
 /**
  * Fetches balances for the specified account across all enabled networks (C-Chain, X-Chain, P-Chain, other EVMs, BTC, SOL, etc.)
--- a/packages/core-mobile/app/new/features/portfolio/hooks/useAccountBalances.ts
+++ b/packages/core-mobile/app/new/features/portfolio/hooks/useAccountBalances.ts
@@ -1,4 +1,8 @@
-import { QueryClient, useQuery, useQueryClient } from '@tanstack/react-query'
+import {
+  type QueryClient,
+  useQuery,
+  useQueryClient
+} from '@tanstack/react-query'
 import { ReactQueryKeys } from 'consts/reactQueryKeys'
 import { useCallback, useMemo } from 'react'
 import { useSelector } from 'react-redux'
--- a/packages/core-mobile/app/new/features/swap/hooks/useMaxSwapAmount/useSpendableXpBalance.ts
+++ b/packages/core-mobile/app/new/features/swap/hooks/useMaxSwapAmount/useSpendableXpBalance.ts
@@ -57,12 +57,18 @@ export const useSpendableXpBalance = ({
     fromToken?.type === TokenType.NATIVE &&
     chain !== undefined
 
+  // P needs a loaded feeState: getMaximumUtxoSet builds txs to measure
+  // size, and an undefined feeState makes every build fail — the "capped"
+  // set comes back empty and a bogus 0 balance would be cached.
+  const feeState = chain === 'P' ? getFeeState() : undefined
+
   const enabled =
     isSpendableBalanceRequired &&
     activeAccount !== undefined &&
     fromNetwork !== undefined &&
     chain !== undefined &&
-    xpAddresses.length > 0
+    xpAddresses.length > 0 &&
+    (chain !== 'P' || feeState !== undefined)
 
   const { data } = useQuery({
     // chain and isTestnet are derived from chainId, and account identity is
@@ -81,7 +87,7 @@ export const useSpendableXpBalance = ({
             account: activeAccount,
             isTestnet: Boolean(fromNetwork.isTestnet),
             xpAddresses,
-            feeState: chain === 'P' ? getFeeState() : undefined,
+            feeState,
             filterSmallUtxos: true
           })
       : skipToken,
--- a/packages/core-mobile/app/new/features/swap/hooks/useMaxSwapAmount/useSpendableXpBalance.ts
+++ b/packages/core-mobile/app/new/features/swap/hooks/useMaxSwapAmount/useSpendableXpBalance.ts
@@ -1,0 +1,99 @@
+import { skipToken, useQuery } from '@tanstack/react-query'
+import { TokenType } from '@avalabs/vm-module-types'
+import { Network } from '@avalabs/core-chains-sdk'
+import { useSelector } from 'react-redux'
+import { ReactQueryKeys } from 'consts/reactQueryKeys'
+import { useGetFeeState } from 'hooks/earn/useGetFeeState'
+import { useXPAddresses } from 'hooks/useXPAddresses/useXPAddresses'
+import AvalancheWalletService from 'services/wallet/AvalancheWalletService'
+import { selectActiveAccount } from 'store/account'
+import { selectIsFilterSmallUtxosActive } from 'store/settings/advanced/filterSmallUtxosActive'
+import type { LocalTokenWithBalance } from 'store/balance'
+import { isPChain, isXChain } from 'utils/network/isAvalancheNetwork'
+
+/**
+ * CP-13903: spendable balance for a native X-/P-Chain swap source, derived
+ * from the dust-filtered UTXO set. Needed because the displayed balance can
+ * stay dust-inclusive (the Balance API only honors `filterOutDustUtxos` on
+ * P-Chain, and the VM-module fallback never filters), so a Max derived from
+ * it would exceed what the CCT `getUtxos` callback offers the SDK.
+ *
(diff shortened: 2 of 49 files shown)


packages/core-mobile/app/new/features/swap/services/cct/createCctCallbacks.ts
```diff
@@ -182,8 +188,19 @@ export const createCctCallbacks = (deps: CctCallbackDeps): CctCallbacks => {
   }
 
   const getUtxos: CctCallbacks['getUtxos'] = async chainAlias => {
+    // Read once, before any await: the asset id must describe the same
+    // network the signer was built for, even if developer mode flips
+    // while the UTXO fetch is in flight.
+    const isTestnet = deps.getIsDeveloperMode()
     const signer = await getReadOnlySigner()
-    return signer.getUTXOs(chainAlias)
+    const utxoSet = await signer.getUTXOs(chainAlias)
```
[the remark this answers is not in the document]
B0Y3R-AVA: Good call — the two reads were provably in the same synchronous frame today (the internal read executes before `getReadOnlySigner`'s first await), so they couldn't diverge, but that invariant was incidental rather than structural. Fixed as suggested in 0fc3c8575: the local `getReadOnlySigner` now takes `isTestnet` (defaulting to a fresh read for the other callers), and `getUtxos` threads its hoisted value through both the signer and `getAvaxAssetId`.

packages/core-mobile/app/new/features/portfolio/hooks/useAccountBalances.ts
```diff
@@ -1,4 +1,4 @@
-import { useQuery, useQueryClient } from '@tanstack/react-query'
+import { QueryClient, useQuery, useQueryClient } from '@tanstack/react-query'
```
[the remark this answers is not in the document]
B0Y3R-AVA: Done in b45db07a5 — switched to an inline type-only import (`import { type QueryClient, ... }`), matching the existing convention (e.g. `services/balance/types.ts`). No runtime/bundle delta in this case since the same import statement already carries `useQuery`/`useQueryClient` as values, but agreed it's the right hygiene.

atn4z7 (commented):
Nice PR.

packages/core-mobile/app/services/balance/BalanceService.ts
```diff
@@ -347,7 +349,8 @@ export class BalanceService {
       networks: supportedNetworks,
       accounts: [account],
       xpAddressesByAccountId: new Map([[account.id, xpAddresses]]),
-      xpubByAccountId: new Map([[account.id, xpub]])
+      xpubByAccountId: new Map([[account.id, xpub]]),
+      filterOutDustUtxos
```
atn4z7, line 353: There's a seam between how the displayed balance/Max is filtered vs how the spend set is filtered. Max comes from the balance (Send: `selectedToken.available`; Swap: `fromToken.balance`), while the build set is filtered separately client-side (`createSendPTx`, swap `getUtxos`). Two situations make them diverge with the toggle ON:

1. **X-Chain**: display-filtering rides on the Balance-API `filterOutDustUtxos` flag threaded in here; if Glacier only honors it for P-Chain and not the AVM/X xpub path, the X balance/Max stays dust-inclusive while `getUtxos` drops dust.
2. **Any chain during a Balance-API outage**: the VM-module fallback (`getBalancesForAccountViaVmModules`, called at L390) doesn't pass this flag, and the avalanche-module's `getBalances` has no dust option, so P- and X-Chain balances come back dust-inclusive while send/swap keep filtering locally.

In both, displayed Max ends up larger than the filtered spendable set, so tapping Max builds an over-spend and the tx fails with the original "insufficient balance..." error. Case 2 also hits P-Chain, which is otherwise consistent. core-web/extension avoid this by filtering the displayed balance from the same UTXO set they spend from. Worth deciding whether to filter the fallback/X balance client-side too, or cap Max at the filtered set, so display and spend can't diverge.

packages/core-mobile/app/services/wallet/AvalancheWalletService.ts
```diff
@@ -299,12 +301,18 @@ class AvalancheWalletService {
...
     // P-chain has a tx size limit of 64KB
     let utxoSet = await readOnlySigner.getUTXOs('P')
-    const filteredUtxos = Avalanche.getMaximumUtxoSet({
+    let filteredUtxos = Avalanche.getMaximumUtxoSet({
       wallet: readOnlySigner,
       utxos: utxoSet.getUTXOs(),
       sizeSupportedTx: Avalanche.SizeSupportedTx.BaseP,
       feeState
     })
+    if (filterSmallUtxos === true) {
```
atn4z7: `createSendPTx` filters here, but `createSendXTx` below takes the same `CreateSendPTxParams` (which now carries `filterSmallUtxos`) and never applies it, so a plain X-Chain send always builds from the full UTXO set. Since `baseTX` spends exactly the set we pass it (no re-fetch), this is a real behavior gap, not just a naming one: with the toggle ON, P-Chain send filters dust but X-Chain send doesn't. Not a correctness risk (worst case the send just includes dust it could have excluded), but is leaving X-Chain send unfiltered intentional/extension parity?

packages/core-mobile/app/services/balance/BalanceService.ts
B0Y3R-AVA: Verified both cases and fixed by deriving Max from the spend set instead of the displayed balance:

1. **X-Chain**: confirmed — the generated Balance API types document `filterOutDustUtxos` on X/P request items as "Only supported on P-chain", so the displayed X balance stays dust-inclusive no matter what we send.
2. **VM-module fallback**: confirmed unfiltered (avalanche-module's `getBalances` has no dust parameter).

Rather than filtering the fallback/X balances client-side (mobile doesn't have the UTXO set at balance-mapping time — Glacier returns aggregates), this now follows the extension's `getMaxUtxoSet` pattern: Max and send validation are derived from the same filtered UTXO set the builders spend.

- `AvalancheWalletService.getSpendableAvaxBalance` computes available balance from the shared `getSendUtxoSet` (size-cap for P + dust filter) — the exact set `createSendPTx`/`createSendXTx` consume.
- `usePVMSend`/`useAVMSend` use it for Max **and** thread it into `validate` as `spendableBalance`, so a manually typed over-amount errors up front instead of failing at build.
- Swap Max for native X/P sources is capped the same way via `useSpendableXpBalance` (dust-filtered, matching the CCT `getUtxos` callback). While the spendable balance loads, Max stays disabled rather than falling back to the displayed balance.

Displayed balances themselves still come from the Balance API (same as extension) — only Max/validation are capped, so display and spend can no longer diverge in either the X case or the outage-fallback case.

packages/core-mobile/app/services/wallet/AvalancheWalletService.ts
B0Y3R-AVA: Not intentional — extension filters X-chain sends too (`buildXChainSendTx` → `getMaxUtxoSet` applies the dust threshold on X as well as P), so this was a parity gap. Fixed:

- `createSendXTx` now builds from the shared `getSendUtxoSet` helper (dust filter applied when the gate is on; no `getMaximumUtxoSet` sizing pass on X, matching extension), with the flag threaded through `useAVMSend` → `utils/avm/send.ts`. Also gave X its own `CreateSendXTxParams` type (P's minus `feeState`).
- To avoid introducing the display/spend seam from your other comment onto X send, X Max and validation now derive from the same filtered set (see the reply there).

Pinned by tests in `AvalancheWalletService.sendTx.test.ts` (filtered vs unfiltered spend set) and `avm/validate.test.ts`.

packages/core-mobile/app/new/features/swap/hooks/useMaxSwapAmount/useSpendableXpBalance.ts
```diff
@@ -0,0 +1,99 @@
...
+      ? () =>
+          AvalancheWalletService.getSpendableAvaxBalance({
+            chain,
+            account: activeAccount,
+            isTestnet: Boolean(fromNetwork.isTestnet),
+            xpAddresses,
+            feeState: chain === 'P' ? getFeeState() : undefined,
+            filterSmallUtxos: true
+          })
+      : skipToken,
```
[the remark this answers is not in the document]
B0Y3R-AVA: Partly right — the queryFn did depend on feeState outside the key, but the failure direction is inverted: an undefined feeState doesn't inflate the balance, it zeroes it. `getMaximumUtxoSet` builds transactions to measure size; with no feeState every build throws, the cost function treats each prefix as over-limit, and the "capped" set comes back empty — so a bogus 0 would be cached and Max would stick disabled. Fixed in 0458f6f65 by hoisting `feeState` and gating `enabled` on it for P (same guard added to `usePVMSend`'s fetch effect, which re-fires when defaults load since `getFeeState`'s identity changes).

Deliberately did NOT add the fee price to the query key: the P spendable value is a strict subset of the CCT spend set regardless of feeState (CCT's `getUtxos` never applies the BaseP size cap), so fee drift can only make Max more conservative — and `staleTime: 0` already refetches per mount. Keying by price would just churn the cache on every fee tick.

packages/core-mobile/app/services/wallet/AvalancheWalletService.ts
```diff
@@ -275,6 +277,83 @@ class AvalancheWalletService {
...
+        utxos: filteredUtxos,
+        sizeSupportedTx: Avalanche.SizeSupportedTx.BaseP,
+        feeState
+      })
+    }
+    if (filterSmallUtxos === true) {
+      filteredUtxos = filterOutSmallUtxos(
+        filteredUtxos,
+        getAvaxAssetId(isTestnet)
+      )
```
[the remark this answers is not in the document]
B0Y3R-AVA: Added in 0458f6f65: `createSendPTx` tests pin that `getMaximumUtxoSet` receives the full UTXO set and the dust filter runs on its output (order), plus the flag-off case (size-capped set spent untouched), and a `getSpendableAvaxBalance` P-chain test asserting dust is excluded from the capped balance. `getMaximumUtxoSet` is mocked as a pass-through since the real implementation builds transactions to binary-search the size limit, which a mocked signer can't support — the order/filter behavior is what these tests pin.

packages/core-mobile/app/new/features/swap/hooks/useMaxSwapAmount/useSpendableXpBalance.ts
```diff
@@ -0,0 +1,105 @@
...
+    fromNetwork !== undefined &&
+    chain !== undefined &&
+    xpAddresses.length > 0 &&
+    (chain !== 'P' || feeState !== undefined)
+
+  const { data } = useQuery({
+    // chain and isTestnet are derived from chainId, and account identity is
+    // its id — the key below fully determines the fetch inputs.
+    // eslint-disable-next-line @tanstack/query/exhaustive-deps
+    queryKey: [
```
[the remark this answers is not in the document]
B0Y3R-AVA: Fixed in 0571c15e6 — the hook now returns `spendableBalance: isFetching ? undefined : data`, so Max stays disabled while a refetch is in flight instead of briefly serving the cached pre-spend value. One correction to the framing: this window couldn't cause an actual over-spend — the CCT `getUtxos` callback re-filters the live UTXO set at build time, so the worst case was a late "insufficient funds" failure after tapping Max, not funds movement. Gating on freshness removes that UX window too.

packages/core-mobile/app/store/settings/advanced/filterSmallUtxosActive.ts
```diff
@@ -0,0 +1,15 @@
+import { RootState } from 'store/types'
```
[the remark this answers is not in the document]
B0Y3R-AVA: Applied in cfa8784b1. For the record there's no enforced consistent-type-imports convention here (36 of 37 RootState imports under store/ are value-form, and lint passes either way — TS erases it at transpile), but type-only is harmless and matches the earlier QueryClient cleanup on this branch.

packages/core-mobile/app/services/balance/BalanceService.ts
atn4z7: Confirmed resolved. Max now derives from the client-side dust-filtered spendable set (`getSpendableAvaxBalance`), so it no longer depends on Glacier honoring the flag for X-Chain or on the VM-module fallback filtering, and it stays disabled until that lands so it can't fall back to the dust-inclusive balance. Covers P-send, X-send, and swap. Resolving.

packages/core-mobile/app/services/wallet/AvalancheWalletService.ts
atn4z7: Confirmed resolved — `createSendXTx` now routes through the shared `getSendUtxoSet({ chain: 'X', filterSmallUtxos })`, and X-send Max/validation derive from the filtered spendable balance. Resolving.

packages/core-mobile/app/new/features/swap/hooks/useMaxSwapAmount/utils.ts
```diff
@@ -105,24 +105,33 @@ export const computeMaxAmount = ({
...
+  /**
+   * CP-13903: dust-filtered spendable balance for native X/P sources.
+   * When set it replaces the displayed balance, which can include dust
+   * the CCT spend set excludes.
+   */
+  spendableBalance?: bigint
 }): bigint | undefined => {
   if (!fromToken) return undefined
 
+  const balance = spendableBalance ?? fromToken.balance
```
atn4z7: Heads up: main just merged #3988 (CP-14788), which also rewrote `computeMaxAmount` to use `getSwappableBalance(fromToken)` (excludes P/X staked/locked), so this line conflicts. They stack rather than compete: `spendableBalance` (from `getAssetBalance(utxoSet).available`) already excludes staked/locked *and* dust, so it's a subset of swappable. Suggest resolving as `const balance = spendableBalance ?? getSwappableBalance(fromToken)` (fall back to swappable, not `fromToken.balance`) and using `balance` throughout — that keeps #3988's staked exclusion when the filter is off/non-native/loading, and your dust+fallback fix when it's on. `store/posthog/slice.test.ts` also has a trivial import/`describe` conflict, keep both sides.

ruijia1in (approved):
dev-tested

atn4z7 (approved):

PR #3984, opened 2026-07-15
CP-13659: integrate perps-sdk for Hyperliquid trading

0xstt:
# CP-13659: wire perps-sdk

## Summary

Replaces the mocked perpetuals feature with a live integration against
`@avalabs/perps-sdk` (Hyperliquid). This wires real market data, account state,
and the full trade lifecycle (open / close / manage / deposit / withdraw) into
the existing perpetuals UI, and removes all placeholder/mock data.

Hyperliquid is mainnet-only in Core (no testnet path).

### Data & SDK integration

- Live market universe + asset contexts via `metaAndAssetCtxs`, including HIP-3
  (builder-deployed) markets merged in with namespaced `dex:TICKER` symbols.
- Delisted markets filtered out of the tradeable universe (`filterDelistedFromMetaAndAssetCtxs`).
- Live account state: clearinghouse (balance, positions, margin), open orders,
  user fills, and per-coin active-asset leverage — seeded over REST and kept
  live over WebSocket.
- Live mids feed with a best-effort connection monitor (heartbeat + reconnect
  nudge) since the SDK exposes no socket-level events.

### Trading flows

- Place order (market/limit), close (market/limit), and manage (TP/SL) wired to
  the SDK exchange client.
- Leverage sourced from Hyperliquid per coin (no fabricated defaults); the
  leverage screen writes to HL and re-reads the applied value.
- Enable-trading gate (agent key + Markr builder fee + unified account) shared
  across all order-signing actions.
- Agent (session) key: generated, approved once via the master wallet, verified
  against HL, and persisted in secure storage for silent L1 signing.
- Markr fixed-route deposit/withdraw between Avalanche C-Chain USDC and
  Hyperliquid perp USDC via `FusionService`.

iOS - 9340
Android - 9341


packages/core-mobile/app/new/features/trade/perpetuals/screens/PerpetualsCloseScreen.tsx
```diff
@@ -91,101 +94,114 @@ const ProfitText = ({ value }: { value: number | undefined }): JSX.Element => {
...
+  const { closePosition, busy } = usePerpsPositionActions()
+  const { requireTradingEnabled, enableTradingModal } =
+    usePerpsEnableTradingGate()
 
   const [receive, setReceive] = useState(positionValue / 2)
   const fraction = positionValue > 0 ? receive / positionValue : 0
   const estimatedProfit = fraction * totalPnl
 
+  const szi = signedSizeContracts(positionValue, price, side)
+  const sizeToClose = price > 0 ? receive / price : 0
```
bogdandobritoiu: 🔴 **Blocking: close orders throw in the SDK for nearly every position — unquantized size**

`sizeToClose` / `szi` are raw divisions and `closePosition`/`limitClose` pass `sz` to the SDK as-is. The SDK only quantizes *prices* — `sz` goes straight to `floatToWire`, which throws on any value with float residue beyond 8dp:

```
floatToWire(320/64321.55) → Error: floatToWire causes rounding for 0.004975004489164206
```

So in practice almost every market **and** limit close throws before reaching Hyperliquid and the user gets the raw `floatToWire causes rounding for 0.0049…` message in a snackbar — positions can't be closed. The open flow already handles this with `roundSizeToSzDecimals` (its doc comment even warns about this exact throw); the close path needs the same treatment.

While here: for a 100% close, consider carrying the real `szi` through the route instead of re-deriving `value / price` — once floor-quantization is added, the FP round-trip can floor one lot low and leave dust (reduce-only protects against overshoot, not undershoot).

packages/core-mobile/app/new/features/trade/perpetuals/utils/toPosition.ts
```diff
@@ -0,0 +1,252 @@
...
+      : undefined
+
+  const { takeProfit, stopLoss } = extractPositionTriggers(
+    position.coin,
+    orders
+  )
+
+  return {
+    id: position.coin,
+    symbol: tickerOfCoin(position.coin),
```
bogdandobritoiu, line 189: 🔴 **Blocking: the dex-stripped `symbol` is used to route HIP-3 close/manage — wrong market**

`symbol` strips the `dex:` namespace, but `usePositionActions` builds the close/manage routes from `position.symbol` (not `position.id`, which keeps the full coin). Closing/managing a HIP-3 position (`somedex:BTC`) therefore submits against **main-dex BTC**: if the user also holds a same-ticker main-dex position on the same side, *that* position gets closed (wrong market, real money); otherwise the reduce-only order is rejected and the HIP-3 position can't be closed at all. Same for TP/SL via manage.

The SDK routes `dex:TICKER` correctly (per-dex mids, `builderPerpAssetId`), so switching `usePositionActions` to `position.id` should be a complete fix.

Downstream of the same stripping: `PerpsCoinLogo` fetches the bare-ticker SVG for HIP-3 positions and `DexBadge` never renders (`dexOfCoin(bareTicker)` is always `''`).

packages/core-mobile/app/new/features/trade/perpetuals/screens/PerpetualsManageScreen.tsx
```diff
@@ -79,17 +83,48 @@ export const PerpetualsManageScreen = (): JSX.Element => {
...
-    // UI-only: simulate the update. SDK wiring (updateLeverage /
-    // setPositionTpSl) lands in a follow-up.
+    // Setting TP/SL signs an L1 order, so ensure trading is set up first.
+    if (!requireTradingEnabled()) {
+      return
+    }
     setSubmitting(true)
     try {
-      await new Promise(resolve => setTimeout(resolve, 1200))
+      const tpChanged = effectiveTakeProfitPrice !== initialTakeProfitPrice
```
bogdandobritoiu: 🟠 **TP/SL edits never cancel the existing trigger orders**

`setPositionTpSl` only *places* new reduce-only triggers — it doesn't cancel or modify existing ones, and nothing in the feature ever issues a cancel. Consequences:

- Editing SL $90 → $80 leaves the $90 order resting, so the position still closes at $90 — the edit is silently ineffective.
- Changing only the TP re-sends the unchanged SL too, duplicating it.
- Clearing both throws in the SDK (`setPositionTpSl requires takeProfitPx and/or stopLossPx`) → "Failed to update TP/SL" — there's currently no way to remove a TP/SL at all.

The manage flow needs to cancel the position's existing trigger orders before placing replacements (and allow the both-cleared case to be cancel-only).

packages/core-mobile/app/new/features/trade/perpetuals/hooks/useSeededPlaceOrderProps.ts
```diff
@@ -50,37 +49,67 @@ const toPositive = (value: string | undefined): number | undefined => {
...
+  const liveMaxLeverage = universe?.maxLeverage
+  const liveMarkPrice = parseHlNumber(assetCtx?.markPx)
+  // The manage flow carries the position's own entry price; only the open flow
+  // (no `entry` param) should adopt the live mark price.
+  const isOpenFlow = params.entry === undefined
+
+  return {
+    ...resolved,
+    entryPrice:
+      isOpenFlow && liveMarkPrice > 0 ? liveMarkPrice : resolved.entryPrice,
```
bogdandobritoiu, line 101: 🟠 **Manage-flow TP/SL is validated against the historical entry price, not live mark**

In the manage flow `entryPrice` stays the position's original entry, and `isTriggerValid` compares against it. Long entered at $100, mark now $120: TP $110 passes validation but is already crossed — HL will either fire it as a market trigger immediately (closing the position the moment the user taps "Update position", fees + slippage, no warning) or reject the already-crossed trigger. Either way, validating against stale entry lets through a trigger that can't do what the user intended.

The open flow is fine (entry there *is* live mark). Manage should validate triggers against live mark instead. Related: the trigger screen's subtitle says "% above/below **current price**" while computing the % from entry.

packages/core-mobile/app/new/features/trade/perpetuals/screens/PerpetualsLeverageScreen.tsx
```diff
@@ -2,28 +2,67 @@ import { Button, LeverageGauge, Text, View } from '@avalabs/k2-alpine'
...
   const isUnchanged = draftLeverage === leverage
 
-  const handleConfirm = useCallback(() => {
-    setLeverage(draftLeverage)
+  // Commit leverage to Hyperliquid here (not at order submit) so the order
+  // itself never triggers a "decrease leverage" margin rejection. After it
+  // succeeds, read the leverage back from HL so local state reflects the
+  // actual on-chain value rather than assuming the draft was applied.
+  const handleConfirm = useCallback(async () => {
+    const ok = await updateLeverage(coin, draftLeverage, true)
```
bogdandobritoiu, line 51: 🟠 **Liquidation estimate ignores maintenance margin — up to ~2× too optimistic (and isolated-vs-cross mismatch)**

`estimateLiquidationPrice` is `entry ± entry/leverage` — the zero-maintenance-margin bound. With HL's maintenance margin (≈ notional / (2·maxLeverage)), a 10× long at $100 on a 10×-max coin actually liquidates around $94.7, not the displayed $90 — about half the promised buffer at max leverage. Both pre-trade surfaces show it with definitive wording ("Liquidated at $X"), so users will get liquidated well before the price the app told them.

Suggest adding the maintenance-margin term, or at minimum "est." wording. Also: this call hardcodes `isCross: true` while the caption's formula is the isolated-margin shape — cross liquidation depends on whole-account equity and can be further off. Worth reconciling.

packages/core-mobile/app/new/features/trade/perpetuals/hooks/useMarkrPartnerInfo.ts
```diff
@@ -0,0 +1,84 @@
...
+  })
+}
+
+/**
+ * Markr's `fee` is already in Hyperliquid's `builder.f` unit (tenths of
+ * basis points), so this is an identity round/clamp — provided to keep call
+ * sites explicit about the unit they expect and to guard against NaN /
+ * negative payloads.
+ */
+export function markrFeeToHyperliquidTenthsBps(fee: number): number {
```
bogdandobritoiu, line 89: 🟠 **Builder fee is taken verbatim from the remote endpoint — no client-side cap**

The zod schema only requires `fee` to be nonnegative, this helper only clamps NaN/negative, and the SDK's `HYPERLIQUID_PERPS_BUILDER_FEE_MAX_TENTHS_BPS` constant is never enforced on the approval path. A misconfigured (or compromised) response of e.g. `fee: 100` (0.1% — HL's max) would flow straight into the user's approval sheet and then attach to every fill.

`Math.min(fee, EXPECTED_MAX_TENTHS_BPS)` here is a one-line defense that bounds the blast radius to something we chose.

packages/core-mobile/app/new/features/trade/perpetuals/hooks/usePerpsClearinghouse.ts
```diff
@@ -0,0 +1,124 @@
...
+    )
+    return () => {
+      cancelled = true
+      unsub()
+    }
+  }, [manager, userAddress, wsResubscribeNonce])
+
+  const clearinghouse = liveClearinghouse ?? data?.clearinghouse
+
+  return {
```
bogdandobritoiu, line 205: 🟠 **A failed clearinghouse fetch renders as "$0 balance" everywhere**

The hook exposes `isLoading` but no error, and consumers fall back with `?? 0` — so a terminal `/info` failure shows "Total account value $0.00", "Available to withdraw $0.00", a place-order dial capped at $0, and the details screen steering a funded user to "Slide to deposit". An HL API outage becomes indistinguishable from having no funds, with no error state and no retry affordance. (`PerpsApiDownState` only covers manager-*init* failure, not this.)

Suggest exposing `isError` (the hook already returns `refetch`) and gating the balance-driven screens on it. The WS `clearinghouseState` subscription can partially cover a REST-only failure, but a full outage still renders as $0.

packages/core-mobile/app/new/features/trade/perpetuals/utils/orderExecution.ts
```diff
@@ -0,0 +1,211 @@
...
+  type ExchangePlaceOrderResponse,
+  type OrderErrorContext,
+  type PerpsManager
+} from '@avalabs/perps-sdk'
+import { showSnackbar } from 'common/utils/toast'
+import Logger from 'utils/Logger'
+
+/** Dump HL / SDK error fields that `Logger.error(msg, err)` often collapses. */
+export const logHyperliquidError = (label: string, error: unknown): void => {
+  // Prefer Logger.info — Logger.error uses console.groupCollapsed and Metro
```
bogdandobritoiu: 🟠 **Order failures produce zero production telemetry**

All real-money failure diagnostics (order/close/cancel rejects, TP/SL failures, submit throws, agent-approval failures) flow through `Logger.info` — but production sets the log level to `ERROR`, making `Logger.info` a no-op, and only `Logger.error` reaches Sentry. There's no analytics event on the failure paths either.

Net effect: a user reports "my order failed" and there is no log, no Sentry event, nothing server-side to debug with — for the highest-stakes flows in the app. Keeping `info` for the verbose dev dumps is fine, but the failure paths should also emit `Logger.error` (or an explicit Sentry capture / analytics event).

bogdandobritoiu (commented):
Reviewed the full diff (data layer, trading flows, signing/agent-key handling, financial math, error handling, UI). Overall the SDK wiring and architecture look solid — per-order HL errors are handled, entry+TP/SL placement is atomic, agent-key handling is clean, and deposit/withdraw decimal conversions check out.

Posting the 8 issues that affect real-money correctness or leave users blind on failure. The first two (close-order size quantization, HIP-3 coin routing) look blocking to me — closing positions is broken in both cases. Verified each of these against the branch and `@avalabs/perps-sdk@0.3.0` internals, including an empirical repro of the `floatToWire` throw.

packages/core-mobile/app/new/features/trade/perpetuals/hooks/usePerpsLiveMids.ts
```diff
@@ -0,0 +1,188 @@
...
+
+/**
+ * Opens the shared `allMids` feed for the lifetime of the mounting screen:
+ * seeds from REST then streams updates over WebSocket into the live-mid store.
+ * Covers the native (main) dex plus every builder-deployed HIP-3 dex — each
+ * builder dex is a separate `allMids` subscription whose keys are namespaced
+ * (`dex:TICKER`). Holds no React state, so ticks never re-render the host
+ * screen — only the subscribed rows update.
+ */
+export const usePerpsLiveMidsFeed = (): void => {
```
atn4z7, line 237: 🟠 **The mids feed opens a new socket every time it mounts**

The manager is ref-counted, so no matter how many screens use perps you only ever get one WS + one agent boot. The mids feed doesn't get that treatment. `usePerpsLiveMidsFeed` spins up a fresh `createPerpsWsClient()` plus a `getPerpDexs` + per-dex `getAllMids` fan-out on each mount, and it runs on both the list screen and the search screen.

The stack keeps the list mounted under the search screen when you push it, so for that whole overlap you've got two `allMids` sockets, the REST discovery running twice, and duplicate per-dex subs all writing into the same store. Every tick fires `setLiveMids` twice and re-notifies every visible row twice. It cleans itself up once the covered screen unmounts, but search is exactly when the list is doing the most work.

The store, `activeFeeds`, and `lastMidsActivityAt` are already module-level singletons, so the socket can be ref-counted the same way the manager is: first mount connects and subscribes, later mounts just bump a counter and reuse it, last one out tears down. Back to a single socket no matter how many perps screens are stacked.

While you're in here, two small things in the same path:
- `setLiveMids` does `mids = { ...mids, ...next }`, so it clones the whole map (hundreds of coins) several times a second. Nothing reads `mids` by identity (the snapshot returns a plain number), so you can mutate in place and just call the listeners.
- `buildCoinCategoryIndex` runs inside `usePerpetualMarkets` and then again in the screen off the same `usePerpCategories()` source. It's memoized so it's not per-tick, but it's the full ~340-line index built twice. Exposing it from the hook would let the screen reuse it.

packages/core-mobile/app/new/features/trade/perpetuals/hooks/usePerpsClearinghouse.ts
```diff
@@ -0,0 +1,124 @@
...
+  useEffect(() => {
+    setLiveClearinghouse(data?.clearinghouse)
+  }, [data?.clearinghouse])
+
+  useEffect(() => {
+    if (manager === null || userAddress === undefined) {
+      return
+    }
+    let cancelled = false
+    const unsub = manager.ws.subscribe(
```
atn4z7, line 111: 🟠 **Question: does the SDK de-dupe identical WS subscriptions?**

React Query shares the clearinghouse REST fetch across every `usePerpsClearinghouse` by key, but the WS side doesn't get shared the same way. Each instance calls `manager.ws.subscribe({ clearinghouseState })` and holds its own `liveClearinghouse`, so two consumers means two subscriptions, two parse calls per tick, two re-renders. `usePerpsOpenOrders` is the same, and the HIP-3 hooks do it once per builder dex per consumer.

It's fine right now since it's usually one consumer per screen, but it grows quietly the moment something like a summary widget and a positions list are on screen together. If `ws.subscribe` already collapses identical `(channel, user, dex)` tuples internally then ignore this. If it doesn't, might be worth lifting these live overrides into the provider or a shared store like the mids feed already does.

packages/core-mobile/app/new/features/trade/perpetuals/hooks/usePerpsOpenOrders.ts
```diff
@@ -0,0 +1,126 @@
...
+
+/**
+ * Module-level stale-while-revalidate cache keyed by user. Survives screen
+ * remounts (tab switches, navigation, fast refresh) so a returning screen shows
+ * the last known orders immediately instead of blanking to a loading state and
+ * blinking derived UI (e.g. position TP/SL) back to a placeholder. A user is
+ * only "loading" until its first fetch settles once per app session; after that
+ * the cached list is shown and only replaced when a newer fetch returns.
+ */
+const ordersCache = new Map<Address, readonly InfoOrderStatusWire[]>()
```
atn4z7: 🟡 **Trading caches stick around after sign-out**

`ordersCache` / `loadedUsers` here, `stickyTriggers` in `usePerpsPositionsView`, and `hip3OrdersCache` are all keyed by address so they stay correct across account switches. Nothing clears them on logout though, so one account's orders and positions sit in module memory for the rest of the app's life. It's not an unbounded leak, but it's real-money data outliving the session. A `clear()` hung off the sign-out path would cover it.

packages/core-mobile/app/new/features/trade/perpetuals/screens/PerpetualsScreen.tsx
```diff
@@ -50,37 +62,79 @@ export const PerpetualsScreen = ({
...
       case 'Trending':
       default:
-        return PERP_MARKETS_MOCK
+        return categoryFilteredMarkets
     }
-  }, [selectedFilter])
+  }, [selectedFilter, categoryFilteredMarkets])
 
-  const renderItem: ListRenderItem<PerpetualMarket> = useCallback(
+  const renderItem: ListRenderItem<PerpMarketView> = useCallback(
```
atn4z7, line 137: 🟡 **Inline `onPress` cancels out the row memo on list re-renders**

`renderItem` builds a fresh `onPress={() => router.navigate(...)}` each render, so `React.memo` on `PerpetualListItem` never hits when the list re-renders (filter, sort, refetch). Doesn't touch the price-tick path since those go through `useLiveMid` and skip the parent, so this is low stakes. If you pass `symbol` down and navigate from a `useCallback`, the memo starts working again.

atn4z7 (commented):
Nice PR! Really solid work, the SDK wiring and the lazy ref-counted manager setup are clean.

Bogdan's already covered the correctness and real-money side thoroughly, so I won't touch that. I just left a few inline suggestions on the architecture/performance side, mostly around WebSocket lifecycle and some caching. Thanks for putting this together!

packages/core-mobile/app/new/features/trade/perpetuals/utils/toPosition.ts
0xstt: The issue has been resolved.

packages/core-mobile/app/new/features/trade/perpetuals/screens/PerpetualsCloseScreen.tsx
0xstt: The issue has been resolved.

packages/core-mobile/app/new/features/trade/perpetuals/screens/PerpetualsManageScreen.tsx
0xstt: The issue has been resolved.

packages/core-mobile/app/new/features/trade/perpetuals/hooks/useSeededPlaceOrderProps.ts
0xstt: The issue has been resolved.

packages/core-mobile/app/new/features/trade/perpetuals/hooks/useMarkrPartnerInfo.ts
0xstt: The issue has been resolved.

packages/core-mobile/app/new/features/trade/perpetuals/hooks/usePerpsClearinghouse.ts
0xstt: The issue has been resolved.

packages/core-mobile/app/new/features/trade/perpetuals/utils/orderExecution.ts
0xstt: Order/close/cancel/TP-SL/approval failures emit Logger.error for Sentry.

packages/core-mobile/app/new/features/trade/perpetuals/hooks/usePerpsClearinghouse.ts
bogdandobritoiu: Checked the SDK source (`@avalabs/perps-sdk` `dist/ws/ws-client.js`): **yes, it de-dupes**. `subscribe` keys each subscription by `JSON.stringify(deepSortKeys(subscription))` and ref-counts the entry — the wire `{method: "subscribe"}` is only sent when the refCount goes 0→1, and the unsubscribe only goes out when it drops back to 0. Since all of these hooks share the single provider-owned `manager.ws`, identical `(channel, user, dex)` tuples collapse to one server subscription no matter how many consumers mount.

What the SDK does *not* collapse is the handler fan-out: each hook instance still registers its own callback, so N consumers run N parse calls per tick. That is one payload parse each, so I think this is fine as-is — lifting the live overrides into a shared store (like the mids feed) is a reasonable fast-follow if a summary widget and a positions list ever share a screen.

(Note this does not cover the mids-feed thread — that path creates its own `createPerpsWsClient()` per mount instead of using `manager.ws`, so the per-client refCount never kicks in there.)

packages/core-mobile/app/new/features/trade/perpetuals/hooks/usePerpsLiveMids.ts
0xstt: Addressed. `usePerpsLiveMidsFeed` now uses a module-level ref-counted feed: the first consumer creates the socket and subscriptions, overlapping screens reuse them, and the final consumer disconnects. Reconnect nonce changes replace the shared feed only once.

`setLiveMids` now mutates the existing map with `Object.assign` before notifying listeners, avoiding repeated full-map clones.

I also exposed `categoryIndex` from `usePerpetualMarkets`, so `PerpetualsScreen` no longer rebuilds it separately. Added tests covering shared socket reuse, final-consumer teardown, and reconnection.

packages/core-mobile/app/new/features/trade/perpetuals/hooks/usePerpsOpenOrders.ts
0xstt: Addressed. Added a centralized `clearPerpsSessionCaches` function and wired it into the logout cleanup path. It clears main-dex orders, HIP-3 orders, HIP-3 positions, seeded-scope markers, and sticky trigger caches.

Main-dex order caching also uses a generation guard so an in-flight request cannot repopulate the cache after logout. Sticky TP/SL values are now keyed by both account address and coin, preventing data from crossing account switches.

packages/core-mobile/app/new/features/trade/perpetuals/screens/PerpetualsScreen.tsx
0xstt: Addressed. Both list screens now pass a stable `handleMarketPress(symbol) `callback. `PerpetualListItem` creates its row-specific handler with `useCallback`, so filtering, sorting, or parent refetches no longer invalidate `React.memo` solely because of a newly allocated inline callback.

bogdandobritoiu (approved):
Amazing work!  Great job!

ruijia1in (approved):
🔥

PR #3988, opened 2026-07-15
CP-14788: exclude staked amount from P/X-chain swap balance

ruijia1in:
## Description

**Ticket: [CP-14788](https://ava-labs.atlassian.net/browse/CP-14788)**

The swap flow treated `fromToken.balance` as the swappable amount everywhere. For P-chain (PVM) and X-chain (AVM) AVAX, that total **includes staked/locked funds**, which can't be swapped — so the swap form displayed the full balance, seeded the %/Max quick-amounts from it, validated against it, and the token pickers listed it. That let users try to swap an amount that isn't actually swappable (it would fail on-chain).

The unlocked, swappable portion is already exposed as `token.available` (P-chain = `unlockedUnstaked`, X-chain = `unlocked`) — the send flow validates against it too. This PR makes swap use it consistently.

**Changes**
- Added `getSwappableBalance` (bigint) and `getSwappableBalanceDisplayValue` (formatted string) helpers — return `available` / `availableDisplayValue` for PVM/AVM, and fall back to `balance` / `balanceDisplayValue` for every other token type (EVM, ERC20, BTC, SVM, SPL). No behavior change for non-Avalanche-primary tokens.
- Applied them to:
  - The "You pay" displayed balance and its percentage/Max quick-amount buttons.
  - The `exceedsBalance` amount validation.
  - The Max computation (`useMaxSwapAmount`) and the pre-quote seed amount.
  - The fee-validation balance ceilings (`useFeeValidation`) so staked funds aren't counted as covering `amount + fee`.
  - Both "Select a token" lists (from + to).

No new dependencies. Not a breaking change.

## Screenshots/Videos

_To be added — P-AVAX swap form + token picker before/after on iOS and Android._

## Testing

**Dev Testing**

Requires a wallet holding **P-chain AVAX with an active stake** (so total balance > available balance).

- Open Swap and select **P-AVAX** as the "You pay" token. The displayed balance should show only the **available (unstaked)** amount, not the full balance.
- Open Swap **from the P-chain token-detail screen** (the token is preselected via nav params, not the picker). The "You pay" balance must still show the **available** amount — this path rebuilds the token and previously showed the full balance.
- Tap **Max** (and the 25/50/75% buttons). The filled amount should be based on the available balance, never the staked portion.
- Try to type an amount above the available balance → an "insufficient balance" error should appear and the swap should be blocked.
- Enter an amount that fits within available but whose amount + fee would exceed available → it should be blocked (not silently allowed to fail on-chain).
- Open the "Select a token" screen (both the "from" and "to" pickers) → P-AVAX (and X-AVAX) rows should show the **available** amount, not the total.
- Sanity-check a normal C-chain token (e.g. AVAX or USDC) still shows and swaps against its full balance as before.

Build on Bitrise: https://app.bitrise.io/app/7d7ca5af7066e290/pipelines/ee87e185-2e43-41f4-b151-1768d3ba303c

## Build Numbers

- iOS: 9287
- Android: 9288

## Checklist

[CP-14788]: https://ava-labs.atlassian.net/browse/CP-14788?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core-mobile/app/new/features/swap/screens/SelectFromSwapTokenScreen.tsx
+++ b/packages/core-mobile/app/new/features/swap/screens/SelectFromSwapTokenScreen.tsx
@@ -24,6 +24,7 @@ import { selectActiveAccount } from 'store/account'
 import { selectIsDeveloperMode } from 'store/settings/advanced'
 import { useFilteredSwapTokens } from '../hooks/useFilteredSwapTokens'
 import { tokenMatchesSearch } from '../utils/tokenMatchesSearch'
+import { getSwappableBalanceDisplayValue } from '../utils/getSwappableBalance'
 
 /**
  * Token selection screen for the "from" side of a swap.
@@ -149,7 +150,7 @@ export const SelectFromSwapTokenScreen = ({
                   {item.name}
                 </Text>
                 <Text variant="subtitle2">
-                  {item.balanceDisplayValue} {item.symbol}
+                  {getSwappableBalanceDisplayValue(item)} {item.symbol}
                 </Text>
               </View>
             </View>
--- a/packages/core-mobile/app/new/features/swap/screens/SwapScreen.tsx
+++ b/packages/core-mobile/app/new/features/swap/screens/SwapScreen.tsx
@@ -1072,7 +1072,7 @@ export const SwapScreen = (): JSX.Element => {
           disabled={isSwapping}
           editable={false}
           amount={toTokenValue}
-          balance={toToken?.balance}
+          balance={toToken ? getSwappableBalance(toToken) : undefined}
           shouldShowBalance={true}
           title="You receive"
           token={
--- a/packages/core-mobile/app/new/features/swap/screens/SwapScreen.tsx
+++ b/packages/core-mobile/app/new/features/swap/screens/SwapScreen.tsx
@@ -120,6 +120,7 @@ import { useFeeValidation } from '../hooks/useFeeValidation'
 import { useAutoAdvanceOnFeeValidationError } from '../hooks/useAutoAdvanceOnFeeValidationError'
 import { getTokenKey } from '../utils/tokenKey'
 import { resolveReceiveAmount } from '../utils/resolveReceiveAmount'
+import { getSwappableBalance } from '../utils/getSwappableBalance'
 
 // Disable KeyboardAwareScrollView's built-in scroll-into-view. The swap card is
 // short, so KAS clamps its scroll to the maximum and yanks the card up past the
@@ -284,7 +285,7 @@ function computeValidationError({
   if (
     debouncedFromTokenValue !== undefined &&
     fromToken !== undefined &&
-    debouncedFromTokenValue > fromToken.balance
+    debouncedFromTokenValue > getSwappableBalance(fromToken)
   ) {
     return fusionErrors.exceedsBalance()
   }
@@ -1016,7 +1017,7 @@ export const SwapScreen = (): JSX.Element => {
           disabled={isSwapping}
           editable={!isSwapping}
           amount={fromTokenValue}
-          balance={fromToken?.balance}
+          balance={fromToken ? getSwappableBalance(fromToken) : undefined}
           shouldShowBalance={true}
           title="You pay"
           token={
--- a/packages/core-mobile/app/new/features/swap/utils/getSwappableBalance.test.ts
+++ b/packages/core-mobile/app/new/features/swap/utils/getSwappableBalance.test.ts
@@ -1,0 +1,89 @@
+import { TokenType } from '@avalabs/vm-module-types'
+import type { LocalTokenWithBalance } from 'store/balance'
+import {
+  getSwappableBalance,
+  getSwappableBalanceDisplayValue
+} from './getSwappableBalance'
+
+describe('getSwappableBalance', () => {
+  it('returns balance for a native EVM token (no available field)', () => {
+    const token = {
+      type: TokenType.NATIVE,
+      balance: 5000n
+    } as LocalTokenWithBalance
+    expect(getSwappableBalance(token)).toBe(5000n)
+  })
+
+  it('returns balance for an ERC20 token', () => {
+    const token = {
+      type: TokenType.ERC20,
+      balance: 1234n
+    } as LocalTokenWithBalance
+    expect(getSwappableBalance(token)).toBe(1234n)
+  })
+
+  it('returns available (not balance) for a P-chain token with staked funds', () => {
+    // balance includes staked; available is the unlocked/unstaked portion
+    const token = {
+      type: TokenType.NATIVE,
+      balance: 10000n,
+      available: 3000n
+    } as unknown as LocalTokenWithBalance
+    expect(getSwappableBalance(token)).toBe(3000n)
+  })
+
+  it('returns available for an X-chain token with locked funds', () => {
+    const token = {
+      type: TokenType.NATIVE,
+      balance: 8000n,
+      available: 5000n
+    } as unknown as LocalTokenWithBalance
+    expect(getSwappableBalance(token)).toBe(5000n)
+  })
+
+  it('returns available of 0n when everything is staked/locked', () => {
+    const token = {
+      type: TokenType.NATIVE,
+      balance: 10000n,
+      available: 0n
+    } as unknown as LocalTokenWithBalance
+    expect(getSwappableBalance(token)).toBe(0n)
+  })
+
+  it('falls back to balance when available is present but not a bigint', () => {
+    const token = {
+      type: TokenType.NATIVE,
+      balance: 7000n,
+      available: undefined
+    } as unknown as LocalTokenWithBalance
+    expect(getSwappableBalance(token)).toBe(7000n)
+  })
+})
+
+describe('getSwappableBalanceDisplayValue', () => {
+  it('returns balanceDisplayValue for a token without availableDisplayValue', () => {
+    const token = {
+      type: TokenType.ERC20,
+      balanceDisplayValue: '12.5'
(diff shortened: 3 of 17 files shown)


packages/core-mobile/app/new/features/swap/screens/SwapScreen.tsx
```diff
@@ -284,7 +285,7 @@ function computeValidationError({
   if (
     debouncedFromTokenValue !== undefined &&
     fromToken !== undefined &&
-    debouncedFromTokenValue > fromToken.balance
+    debouncedFromTokenValue > getSwappableBalance(fromToken)
   ) {
```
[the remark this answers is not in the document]
ruijia1in: Good catch — fixed in b3468da7d. `computeRecurringBalanceError` now uses `getSwappableBalance(fromToken)` for the native principal+fee check and for the ERC-20 `principalShort` / native-fee checks, so a recurring schedule can't be funded from P/X-chain staked/locked balance. It's a no-op for the EVM/ERC-20 sources and the C-chain native gas token this path actually handles (they have no `available` field), and added unit tests covering the staked-native case.

packages/core-mobile/app/new/features/swap/utils/getSwappableBalance.ts
```diff
@@ -0,0 +1,40 @@
...
+ * `available` field (for P-chain, the unlocked/unstaked portion) — prefer it so
+ * the swap form's displayed balance, its percentage/Max quick-amounts, and the
+ * balance validation all reflect only what the user can actually swap. This
+ * mirrors the send flow, which already validates P-chain against `available`
+ * (see app/new/common/hooks/send/utils/pvm/validate.ts).
+ *
+ * Every other token type (EVM, ERC20, BTC, SVM, SPL) has no `available` field,
+ * so `balance` is the swappable amount. (CP-14788)
+ */
+export const getSwappableBalance = (token: LocalTokenWithBalance): bigint => {
```
B0Y3R-AVA, line 17: **The swap token mappers drop `available`, so this helper silently falls back to the full staked-inclusive balance on the deeplink/asset-detail entry path.**

`mapApiTokenToLocal` and `mapSdkAssetToLocal` rebuild the token copying only `balance`/`balanceInCurrency`/`priceInCurrency` — `available`/`availableDisplayValue` never survive. When swap is opened with `initialTokenIdFrom` (asset-detail "Swap" button, deeplinks), the from-token is resolved via `resolveInitialToken` → `buildLocalToken` → `mapApiTokenToLocal`, so a P/X AVAX from-token arrives without `available` and every call site in this PR (Max, %, validation, fee validation) quietly reverts to the total balance — the exact bug this PR fixes persists on that path. It currently only works when the user picks the token in the from-picker, whose tokens come raw from the portfolio store.

Same omission in `mapSdkAssetToLocal` makes the `SelectToSwapTokenScreen` display change a no-op for SDK-fetched assets.

Suggested fix: thread `available` + `availableDisplayValue` (and the `InCurrency` variants for consistency) through `mapApiTokenToLocal`, `mapSdkAssetToLocal`, and `buildLocalToken` when `balanceData` has them.

packages/core-mobile/app/new/features/swap/screens/SwapScreen.tsx
```diff
@@ -1016,7 +1017,7 @@ export const SwapScreen = (): JSX.Element => {
           disabled={isSwapping}
           editable={!isSwapping}
           amount={fromTokenValue}
-          balance={fromToken?.balance}
+          balance={fromToken ? getSwappableBalance(fromToken) : undefined}
```
B0Y3R-AVA, line 1020: The "You receive" widget ~50 lines below (`balance={toToken?.balance}`, line ~1074) was left on the raw balance. `toToken` can be native P/X AVAX via the CCT route, so a user swapping **into** P/X AVAX with staked holdings sees the staked-inclusive total here while `SelectToSwapTokenScreen` (patched in this PR) shows the smaller available amount for the same token — same field, two different values depending on the screen. One-line fix mirroring this line: `balance={toToken ? getSwappableBalance(toToken) : undefined}`.

packages/core-mobile/app/new/features/swap/screens/SelectFromSwapTokenScreen.tsx
```diff
@@ -149,7 +150,7 @@ export const SelectFromSwapTokenScreen = ({
                   {item.name}
                 </Text>
                 <Text variant="subtitle2">
-                  {item.balanceDisplayValue} {item.symbol}
+                  {getSwappableBalanceDisplayValue(item)} {item.symbol}
```
B0Y3R-AVA, line 153: Minor, possibly follow-up material: `useFilteredSwapTokens` still gates the zero-balance filter on `token.balance > 0n` and sorts by `balanceInCurrency` (staked-inclusive). A fully-staked P/X AVAX position (`available = 0n`) now renders "0 AVAX" here yet still appears as a selectable, high-sorted row — and any amount the user types immediately fails validation. Worth either filtering/sorting on the swappable amount too, or excluding `available === 0n` rows.

packages/core-mobile/app/new/features/swap/utils/getSwappableBalance.test.ts
```diff
@@ -0,0 +1,89 @@
...
+  it('returns available of 0n when everything is staked/locked', () => {
+    const token = {
+      type: TokenType.NATIVE,
+      balance: 10000n,
+      available: 0n
+    } as unknown as LocalTokenWithBalance
+    expect(getSwappableBalance(token)).toBe(0n)
+  })
+
+  it('falls back to balance when available is present but not a bigint', () => {
```
B0Y3R-AVA: Nit: this test is named "available is present but not a bigint" but sets `available: undefined`, which exercises the key-present/value-undefined case rather than a non-bigint value (e.g. a stringified `"3000"`). Harmless — the P/X mapping sets `available` and `balance` together atomically so the non-bigint case looks unreachable in practice — but the title overstates what's proven. Either rename it or add a second case with a string value.

B0Y3R-AVA (commented):
Ran an adversarial review on this (Claude + Codex cross-model, findings verified against `origin/main` source before posting). The approach is sound — `available` really is only on `TokenWithBalancePVM`/`TokenWithBalanceAVM` in the pinned `@avalabs/vm-module-types`, it maps to `unlockedUnstaked`/`unlocked` (excludes atomic memory), and it matches how send validates on main. Two things worth fixing before merge though (inline comments): the swap token mappers drop `available` so the fix silently no-ops for the `initialTokenIdFrom` entry path, and the "You receive" widget still shows the staked-inclusive balance. Two smaller notes inline as well.

packages/core-mobile/app/new/features/swap/utils/getSwappableBalance.ts
ruijia1in: Already addressed in 3970f309 (pushed just before this review landed). Added `buildAvailableFields` and wired it into both `mapApiTokenToLocal` and `mapSdkAssetToLocal`, so `available` / `availableDisplayValue` (+ the `InCurrency` variants and `balancePerType`) now survive the rebuild. `buildLocalToken` delegates to `mapApiTokenToLocal`, so the `initialTokenIdFrom` / deeplink path is covered transitively. Verified the from-token from the asset-detail entry now carries `available`. Resolving.

packages/core-mobile/app/new/features/swap/screens/SwapScreen.tsx
ruijia1in: Fixed in 3824e464 — `balance={toToken ? getSwappableBalance(toToken) : undefined}`, exactly as suggested. The "You receive" widget and `SelectToSwapTokenScreen` now show the same swappable value for a P/X AVAX destination.

packages/core-mobile/app/new/features/swap/screens/SelectFromSwapTokenScreen.tsx
ruijia1in: Partially fixed in 3824e464. `useFilteredSwapTokens` now applies the `hideZeroBalance` filter against `getSwappableBalance(token)`, so a fully-staked P/X AVAX position (`available === 0n`) is excluded from the swap-from picker instead of showing as a selectable "0 AVAX" row — that removes the main issue (the unswappable, high-sorted row). No-op for other token types and unaffected on the to-picker (which passes `hideZeroBalance: false`).

Left the currency sort (`balanceInCurrency`) as-is for now: after the filter change the only residual is that a *partially*-staked position sorts by its total value rather than its swappable value — purely cosmetic ordering. Happy to follow up with a `swappableBalanceInCurrency` sort key if you'd prefer it in this PR.

packages/core-mobile/app/new/features/swap/utils/getSwappableBalance.test.ts
ruijia1in: Fixed in 3824e464 — renamed the undefined case to "available is the key but the value is undefined" and added a second case with a string value (`available: '3000'`) that exercises the actual non-bigint branch of the `typeof … === 'bigint'` guard.

B0Y3R-AVA (approved):
All review findings addressed — verified the fixes: `buildAvailableFields` correctly threads `available*`/`balancePerType` through both mappers (covers the `initialTokenIdFrom` path), "You receive" now uses the swappable balance, the from-picker hides fully-staked positions, and the test cases now prove what they claim. Sort-by-fiat residue is fine as a follow-up. LGTM.

atn4z7 (approved):
dev tested

PR #3989, opened 2026-07-16
CP-14760: show C-Chain atomic import/export in activity history

ruijia1in:
## Description

**Ticket: [CP-14760](https://ava-labs.atlassian.net/browse/CP-14760)**

C-Chain atomic AVAX import/export transactions were missing from the C-Chain activity tab and the AVAX token detail history.

**Root cause:** C-Chain activity is served entirely by `@avalabs/evm-module`, which only queries Glacier's regular EVM `listTransactions` endpoint. C-Chain *atomic* import/export txns don't live there — they're on Glacier's primary-network endpoint (`CChainImportTransaction` / `CChainExportTransaction`), which the activity pipeline never called. (This was a pre-existing data-source gap, not a regression from the recurring-swap activity work.)

**Change:** additionally fetch C-Chain atomic transactions from Glacier's primary-network endpoint (using the 0x C-Chain address) and merge them into the C-Chain activity list.

- `GlacierService.listCChainAtomicTransactions` — wraps `primaryNetwork.listLatestPrimaryNetworkTransactions` (blockchainId `c-chain`).
- `convertCChainAtomicTransaction` — maps `CChainExportTransaction` (`evmInputs`) / `CChainImportTransaction` (`evmOutputs`) into the internal `Transaction` shape, tagged `ImportTx`/`ExportTx`; sums all legs of the asset; timestamps normalized seconds→ms.
- `ActivityService.getActivities` — fetches atomic txns **concurrently** with the EVM history (C-Chain, first page only), dedupes by hash, sorts by timestamp desc. Best-effort: a failure never breaks the EVM history.
- UI: the C-Chain/EVM activity row now renders "… AVAX imported / exported" with the correct icon and red/green status (mirrors the existing X/P-chain rendering).

**Known v1 limitations (follow-ups):** atomic txns are merged on the first 100-item page only (the recent-activity view fetches a single page, so both affected surfaces are covered); later-page/beyond-100 atomic history is out of scope for v1.

## Screenshots/Videos

_To be added from the build below (AVAX token detail Activity + Activity tab filtered to C-Chain, showing import/export rows on iOS and Android)._

## Testing

**Dev Testing**

On a **mainnet** account that has performed a C-Chain export **and** import of AVAX:

1. Open **AVAX token detail → Activity**. Verify the export row shows "`<amount>` AVAX exported" and the import row shows "`<amount>` AVAX imported", each with the correct icon and a red (export) / green (import) amount.
2. Open the **Activity tab** and filter to **C-Chain (Avalanche)**. Verify the same import/export rows appear, interleaved by date with normal C-Chain transactions (correct dates, not 1970).
3. In the AVAX token detail, use the **Imported** and **Exported** filter tabs and confirm the respective rows are listed.
4. Confirm normal C-Chain sends / receives / swaps still render unchanged (no regression).
5. Open a different network's token detail (e.g. ETH) and confirm no behavior change.

## Build Numbers

- iOS: 9332
- Android: 9333

Pipeline: https://app.bitrise.io/app/7d7ca5af7066e290/pipelines/ac30032d-d7fe-48e2-84ff-f0693a34a29a

## Checklist

[CP-14760]: https://ava-labs.atlassian.net/browse/CP-14760?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/core-mobile/app/services/activity/utils/convertCChainAtomicTransaction.ts
```diff
@@ -0,0 +1,91 @@
...
+  CChainExportTransaction,
+  CChainImportTransaction,
+  PChainTransactionType
+} from '@avalabs/glacier-sdk'
+import { TokenUnit } from '@avalabs/core-utils-sdk'
+import { Transaction } from 'store/transaction'
+
+const isExport = (
+  tx: CChainImportTransaction | CChainExportTransaction
+): tx is CChainExportTransaction => tx.txType === 'ExportTx'
```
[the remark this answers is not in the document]
ruijia1in: Good catch — replaced the string literal with the enum constant in 5021652b9. I used `CChainExportTransaction.txType.EXPORT_TX` (rather than `PChainTransactionType.EXPORT_TX`) because `tx.txType` here is the Glacier C-Chain atomic enum; comparing it against the vm-module `PChainTransactionType` is a cross-enum comparison TypeScript rejects as non-overlapping. This resolves the drift concern while keeping tsc happy.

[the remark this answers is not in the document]
ruijia1in: Good catch — replaced the string literal with the enum constant in 5021652b9. I used `CChainExportTransaction.txType.EXPORT_TX` (rather than `PChainTransactionType.EXPORT_TX`) because `tx.txType` here is the Glacier C-Chain atomic enum; comparing it against the vm-module `PChainTransactionType` is a cross-enum comparison TypeScript rejects as non-overlapping. This resolves the drift concern while keeping tsc happy.

packages/core-mobile/app/services/activity/ActivityService.tsx
```diff
@@ -13,8 +13,12 @@ import { getCachedTokenList } from 'hooks/networks/useTokenList'
 import { queryClient } from 'contexts/ReactQueryProvider'
 import { ReactQueryKeys } from 'consts/reactQueryKeys'
 import { AdjustedNormalizedBalancesForAccount } from 'services/balance/types'
+import { isAvalancheCChainId } from 'services/network/utils/isAvalancheNetwork'
+import GlacierService from 'services/glacier/GlacierService'
+import { Transaction } from 'store/transaction'
 import { ActivityResponse, GetActivitiesForAccountParams } from './types'
```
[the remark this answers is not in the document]
ruijia1in: Good catch — fixed in 43a5c1317. Switched to a type-only import pulled directly from the types module: `import type { Transaction } from 'store/transaction/types'`. `Transaction` is only used in type positions here, so `import type` guarantees it is fully erased at runtime, and importing from `store/transaction/types` (rather than the `store/transaction` barrel that re-exports `./api`) removes the ActivityService → store/transaction → api → ActivityService circular-import path entirely. tsc + the activity unit tests pass.

packages/core-mobile/app/services/activity/utils/convertCChainAtomicTransaction.ts
```diff
@@ -0,0 +1,92 @@
...
+  Transaction as InternalTransaction
+} from '@avalabs/vm-module-types'
+import {
+  CChainExportTransaction,
+  CChainImportTransaction,
+  PChainTransactionType
+} from '@avalabs/glacier-sdk'
+import { TokenUnit } from '@avalabs/core-utils-sdk'
+import { Transaction } from 'store/transaction'
+
```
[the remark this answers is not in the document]
ruijia1in: Good catch — fixed in 43a5c1317. Same treatment: `import type { Transaction } from 'store/transaction/types'`. It is used only as the return type here, so `import type` erases it at runtime and importing from the types module (not the `store/transaction` barrel that re-exports `./api`) avoids pulling in the runtime dependency / circular-import risk. tsc + `convertCChainAtomicTransaction.test.ts` pass.

packages/core-mobile/app/services/activity/utils/convertCChainAtomicTransaction.ts
```diff
@@ -0,0 +1,92 @@
...
+  const userAddress = exported
+    ? (legs[0] as CChainExportTransaction['evmInputs'][number] | undefined)
+        ?.fromAddress
+    : (legs[0] as CChainImportTransaction['evmOutputs'][number] | undefined)
+        ?.toAddress
+
+  const totalRaw = legs
+    .filter(l => l.asset?.assetId === primaryAsset?.assetId)
+    .reduce((sum, l) => sum + BigInt(l.asset.amount), 0n)
+
```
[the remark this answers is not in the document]
ruijia1in: Good catch — fixed in 755b09628. The filter now guards the leg asset explicitly (`l.asset && l.asset.assetId === primaryAsset?.assetId`) so a leg with a missing `asset` can no longer slip through when `primaryAsset` is also undefined (the `undefined === undefined` case), which would have thrown at `BigInt(l.asset.amount)` in the reduce. When the primary asset is absent, `totalRaw` now stays `0n` and the amount falls back to `0`. Added a regression test ("does not throw when a leg is missing its asset and falls back to 0"). tsc + the 5 unit tests pass.

packages/core-mobile/app/services/activity/utils/convertCChainAtomicTransaction.ts
```diff
@@ -0,0 +1,92 @@
...
+    .toDisplay()
+    .replaceAll(',', '')
+
+export const convertCChainAtomicTransaction = (
+  tx: CChainImportTransaction | CChainExportTransaction,
+  network: { chainId: number; explorerUrl?: string }
+): Transaction => {
+  const exported = isExport(tx)
+  const legs = exported ? tx.evmInputs : tx.evmOutputs
+  const primaryAsset = legs[0]?.asset
```
bogdandobritoiu, line 65: Both the asset and the user address are taken from `legs[0]` only. Two edge cases fall out of that:

1. If `legs` is ever empty, this still produces a phantom row — `amount: '0'`, symbol defaulting to `'AVAX'`, empty `from`/`to` — rendering as "0 AVAX imported". Returning `null` here (and filtering in `fetchCChainAtomicTransactions`) seems safer than showing a zero row.
2. A multi-asset atomic tx (historical ANT movements) gets summed and labelled by whatever asset happens to be first. Rare in practice, but worth either preferring the AVAX leg explicitly or a comment stating the AVAX-only v1 assumption.

packages/core-mobile/app/services/activity/utils/convertCChainAtomicTransaction.ts
```diff
@@ -0,0 +1,92 @@
...
+  const userAddress = exported
+    ? (legs[0] as CChainExportTransaction['evmInputs'][number] | undefined)
+        ?.fromAddress
+    : (legs[0] as CChainImportTransaction['evmOutputs'][number] | undefined)
+        ?.toAddress
+
+  const totalRaw = legs
```
bogdandobritoiu: Product question, not necessarily a code change: for exports, summing `evmInputs` gives the total consumed on C-Chain — transferred amount *plus* the atomic export fee — so "X AVAX exported" reads slightly higher than what arrives on the destination chain. The fixtures carry `amountUnlocked`/`amountCreated` but the converter never reads them. Which number does core-web (and our own X/P row rendering) display? Fine as a follow-up if this matches the existing convention.

packages/core-mobile/app/services/activity/ActivityService.cchainAtomic.test.ts
```diff
@@ -0,0 +1,106 @@
...
+beforeEach(() => {
+  jest.clearAllMocks()
+  ;(ModuleManager.loadModuleByNetwork as jest.Mock).mockResolvedValue({
+    getTransactionHistory: jest
+      .fn()
+      .mockResolvedValue({ transactions: [evmTx], nextPageToken: 'evm-next' })
+  })
+})
+
+describe('ActivityService C-Chain atomic merge', () => {
```
bogdandobritoiu, line 40: The doc comment on `fetchCChainAtomicTransactions` promises a Glacier failure never breaks the EVM history — that's the contract the fire-before-await concurrency depends on, but no test exercises it. One case with `listCChainAtomicTransactions` rejecting → expect the EVM transactions returned and no throw would pin it.

packages/core-mobile/app/services/activity/utils/convertCChainAtomicTransaction.ts
```diff
@@ -0,0 +1,92 @@
...
+  CChainImportTransaction,
+  PChainTransactionType
+} from '@avalabs/glacier-sdk'
+import { TokenUnit } from '@avalabs/core-utils-sdk'
+import type { Transaction } from 'store/transaction/types'
+
+const isExport = (
+  tx: CChainImportTransaction | CChainExportTransaction
+): tx is CChainExportTransaction =>
+  tx.txType === CChainExportTransaction.txType.EXPORT_TX
```
bogdandobritoiu, line 14: Optional: anything that isn't `EXPORT_TX` is treated as an import and read via `evmOutputs`, so an unexpected txType from the endpoint would silently become a mislabelled ImportTx row. The C-Chain primary-network endpoint should only ever serve atomic types, but filtering `transactions` to Import/Export before converting would make the `as ListCChainAtomicTransactionsResponse` cast in GlacierService honest.

bogdandobritoiu (commented):
Verified the integration seams before reviewing — the design is sound: tagging with `PChainTransactionType.IMPORT_TX/EXPORT_TX` is exactly what the existing infra expects (`ActivityTransactionType` union, `TransactionTypeIcon`'s `ImportTx`/`ExportTx` keys, and the token-detail Imported/Exported filter tabs all match), X/P rows render through separate `Xp*` components so they're unaffected, atomic rows can't leak into other C-Chain tokens' detail screens (`useTokenActivity` filters by symbol), and the seconds→ms normalization is correct and tested.

Four things inline — two small asks (empty-legs guard, failure-path test), one product question (export fee in the displayed amount), one optional guard. Non-blocking nits:

- `convertCChainAtomicTransaction` can return `Transaction` directly — the store type's `txType` is `ActivityTransactionType`, which already includes `PChainTransactionType`, so the `Omit<...> & { txType: ... }` intersection and the final `as Transaction` cast are both unnecessary.
- `[renderAmount(a1), ' ', s1, ' ', 'exported']` — sibling cases write the verb as one string (`' sent'`, `' approved'`); match the style.

packages/core-mobile/app/services/activity/utils/convertCChainAtomicTransaction.ts
ruijia1in: Both handled in 8e79b1f3a. Empty `legs` now returns `null` (and `fetchCChainAtomicTransactions` filters the nulls out), so we never emit a phantom "0 AVAX" row — covered by a new `returns null for an export with no evm legs (no phantom row)` test. For the multi-asset case I added a comment documenting the AVAX-only v1 assumption (primary asset = first leg); a genuinely multi-asset atomic tx would still be summed/labelled by the first asset, which is out of scope for v1.

packages/core-mobile/app/services/activity/utils/convertCChainAtomicTransaction.ts
ruijia1in: Checked the convention: the Avalanche module's X/P history mapper renders the emitted/created amount (net of fee), not the consumed inputs — so summing `evmInputs` read high by the atomic export fee and did **not** match. Fixed in 8e79b1f3a: the displayed amount now comes from `tx.amountCreated` (assets created by the outputs = net transferred) for both directions, which lines up with the X/P rows. Added a test (`shows the created amount, excluding the atomic export fee on the inputs`) pinning consumed = 1.501 → displayed = 1.5.

packages/core-mobile/app/services/activity/ActivityService.cchainAtomic.test.ts
ruijia1in: Added in 8e79b1f3a: `never breaks the EVM history when the atomic fetch rejects` mocks `listCChainAtomicTransactions` rejecting and asserts the EVM history (`['0xevm']` + `nextPageToken: 'evm-next'`) comes back intact with no throw — pinning the fire-before-await contract the doc comment promises.

packages/core-mobile/app/services/activity/utils/convertCChainAtomicTransaction.ts
ruijia1in: Done in 8e79b1f3a. `listCChainAtomicTransactions` now filters the response to `ExportTx`/`ImportTx` **before** the `as ListCChainAtomicTransactionsResponse` cast, so an unexpected txType from the endpoint is dropped rather than converted into a mislabelled ImportTx row. The cast is now honest.

ruijia1in:
Also addressed the two non-blocking nits from your review body in 8e79b1f3a:

- **Unnecessary cast** — `convertCChainAtomicTransaction` now builds and returns `Transaction` directly (return type `Transaction | null`); dropped the `Omit<InternalTransaction, 'txType'> & { txType }` intersection and the trailing `as Transaction` cast, since the store `Transaction`'s `txType` (`ActivityTransactionType`) already admits `PChainTransactionType`.
- **Verb style** — the export/import titles now read `[renderAmount(a1), ' ', s1, ' exported']` / `' imported'`, matching the sibling `' sent'` / `' approved'` cases in the same file.

bogdandobritoiu (dismissed):

onghwan (dismissed):
dev-tested

packages/core-mobile/app/services/activity/ActivityService.tsx
```diff
@@ -37,6 +41,16 @@ export class ActivityService {
...
 
+    // Kick off the C-Chain atomic fetch concurrently with the EVM history
+    // fetch below — it doesn't depend on the EVM result, and both are
+    // separate Glacier round-trips, so starting them in parallel avoids
+    // serializing the C-Chain activity critical path.
+    const atomicPromise = this.fetchCChainAtomicTransactions({
+      network: networkWithTokens,
+      address,
+      nextPageToken
+    })
```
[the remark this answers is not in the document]
ruijia1in: Fixed in 228c39abd. `getActivities` now threads its `pageSize` through `fetchCChainAtomicTransactions` to `GlacierService.listCChainAtomicTransactions`, so the atomic request matches the caller's page size instead of always pulling Glacier's 100-item default (the real activity callers pass `pageSize: 100`, so their behavior is unchanged). Added a regression test (`threads the caller pageSize through to the atomic Glacier fetch`) asserting the value is passed through. tsc + the activity unit tests pass.

packages/core-mobile/app/services/activity/ActivityService.tsx
```diff
@@ -57,12 +71,71 @@ export class ActivityService {
...
+    nextPageToken?: string
+  }): Promise<Transaction[]> {
+    if (nextPageToken !== undefined) return []
+    if (!isAvalancheCChainId(network.chainId)) return []
+
+    try {
+      const atomic = await GlacierService.listCChainAtomicTransactions({
+        address,
+        isTestnet: Boolean(network.isTestnet)
+      })
```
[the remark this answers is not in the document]
ruijia1in: Fixed in 228c39abd. `getActivities` now threads its `pageSize` through `fetchCChainAtomicTransactions` to `GlacierService.listCChainAtomicTransactions`, so the atomic request matches the caller's page size instead of always pulling Glacier's 100-item default (the real activity callers pass `pageSize: 100`, so their behavior is unchanged). Added a regression test (`threads the caller pageSize through to the atomic Glacier fetch`) asserting the value is passed through. tsc + the activity unit tests pass.

onghwan (approved):

bogdandobritoiu (approved):

PR #3990, opened 2026-07-16
CP-14585: route to onboarding when wallet keychain data is gone

ruijia1in:
## Description

**Ticket: [CP-14585](https://ava-labs.atlassian.net/browse/CP-14585)**

Fixes a seedless-wallet lockout where an expired session leaves the user stranded on the Unlock (PIN) screen instead of being routed back to onboarding.

**Root cause.** Wallet deletion is non-atomic. When a seedless session expires and re-authentication fails, `onLogOut → clearData` sets `walletState = NONEXISTENT` in memory but the redux-persist write is async, while the keychain (encryption key + wallet secret) is wiped immediately. If the app is killed/crashes in that window ("kicked out of app"), the next launch rehydrates a stale `INACTIVE` state with no keychain credential. `RootNavigator` then routes to the PIN screen — where the correct PIN reads as "wrong" (encryption key gone) or login silently fails (wallet secret gone). There was no code path back to onboarding.

**Approach: reactive PIN-flow recovery.** Rather than probing the keychain at launch (which would risk deleting a real wallet on a transient read error, a stale-legacy false positive, or a keychain-options mismatch), recovery happens only when the user deliberately attempts to unlock:

- `BiometricsSDK.loadEncryptionKeyWithPin()` now returns `'success' | 'wrong-pin' | 'no-credentials'`. `no-credentials` is derived from the same `getGenericPassword` read a normal unlock already uses; a *transient* keychain failure **throws** instead of reporting `no-credentials`, so a flaky read is never mistaken for a missing wallet.
- `checkEnteredPin` routes `no-credentials` to `deleteWallet()` → onboarding. `wrong-pin` still rate-limits exactly as before. A not-yet-migrated legacy wallet migrates first, so it is never mistaken for a missing wallet.
- `PinScreen` retains the safety net for the other variant (encryption key present but the per-wallet secret gone): a verified PIN with a missing secret deletes the wallet and routes to onboarding instead of silently swallowing the error.
- `init()` is unchanged from `main` — no launch-time keychain teardown.

## Screenshots/Videos

_N/A — no new UI; this is a recovery-path fix. Behavior is covered by unit tests; see Testing._

## Testing

**Dev Testing**
iOS: 9422
Android: 9423

Simulating the corrupt state (wallet "exists" per redux, but the keychain credential is gone):

1. Create/import a wallet and unlock it so the app is signed in.
2. Force-close the app.
3. Externally remove the keychain credential while leaving persisted redux intact — delete the `encryption-key-service` entry (variant A), or the `sec-storage-service-*` wallet-secret entry (variant B). Or reproduce via the ticket's seedless session-expiry → failed re-auth → force-close sequence.
4. Relaunch → the app shows the PIN screen (wallet still "exists" per redux). Enter the **correct** PIN.
   - **Variant A (encryption key gone):** instead of an endless "wrong PIN", the stale wallet is deleted and you land on **onboarding**.
   - **Variant B (wallet secret gone):** the PIN verifies, then the missing secret is detected → wallet deleted → **onboarding**.

Regression checks:

5. Normal locked wallet: background past the lock timeout, reopen → PIN screen; correct PIN unlocks, wrong PIN is rejected/rate-limited as before.
6. Not-yet-migrated legacy wallet: entering the correct PIN migrates and unlocks (is **not** treated as a missing wallet).
7. Fresh install (no wallet) → onboarding as before.

## Checklist

[CP-14585]: https://ava-labs.atlassian.net/browse/CP-14585?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/core-mobile/app/store/app/listeners.test.ts
```diff
@@ -0,0 +1,135 @@
...
+
+const flush = async (): Promise<void> => {
+  for (let i = 0; i < 20; i++) {
+    await Promise.resolve()
+    await new Promise(resolve => setTimeout(resolve, 0))
+  }
+}
+
+describe('app listeners — wallet state reconciliation on launch', () => {
+  it('routes to onboarding (NONEXISTENT) when a wallet is persisted but its keychain data is gone', async () => {
```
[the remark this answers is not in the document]
ruijia1in: Not applicable here — this repo's `jest.config.js` sets `clearMocks: true` (line 6), which runs `jest.clearAllMocks()` before every test, so mock call history is already reset per-test. The suite passes green, including this last `not.toHaveBeenCalled()` case (it would fail if prior calls leaked). Adding a local `beforeEach(jest.clearAllMocks)` would be redundant with the global config, so leaving as-is.

packages/core-mobile/app/utils/BiometricsSDK.ts
```diff
@@ -177,13 +177,30 @@ class BiometricsSDK {
...
+  ): Promise<'success' | 'wrong-pin' | 'no-credentials'> {
     const credentials = await Keychain.getGenericPassword(passcodeGetOptions)
-    if (!credentials) return false
+    if (!credentials) return 'no-credentials'
     const decrypted = await decrypt(credentials.password, pin)
-    if (!decrypted) return false
+    if (!decrypted) return 'wrong-pin'
     this.#encryptionKey = decrypted.data
-    return true
+    return 'success'
```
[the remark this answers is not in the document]
ruijia1in: Addressed in `loadEncryptionKeyWithPin`. You're right that the contract was not honored: in this codebase `decrypt` never returns a falsy value — a wrong PIN surfaces as a *thrown* bad-decrypt error (`BAD_DECRYPT` on Android, `Decrypt failed` on iOS), so the old `if (!decrypted) return 'wrong-pin'` line was effectively dead and a wrong PIN propagated as an exception instead.

It now wraps `decrypt` in try/catch: known bad-decrypt errors map to `'wrong-pin'`, and everything else (corrupt data, `NoSaltError`/version issues, transient keychain failures) is rethrown for the caller to handle distinctly. Added unit tests covering BAD_DECRYPT → wrong-pin, "Decrypt failed" → wrong-pin, and rethrow-on-other-errors.

packages/core-mobile/app/new/common/utils/ensureWalletSecret.ts
```diff
@@ -0,0 +1,29 @@
...
+  const result = await BiometricsSDK.loadWalletSecret(walletId)
+  if (!result.success) {
+    Logger.error(
+      'Wallet secret missing after PIN verification; deleting wallet',
+      result.error
+    )
+    onMissing()
+    return false
+  }
+  return true
```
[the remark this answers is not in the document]
ruijia1in: Addressed in `ensureWalletSecret`. It now only triggers recovery (`onMissing` → wallet delete) for terminal failures — `No credentials found`, `Failed to decrypt`, and bad-decrypt (`BAD_DECRYPT` / `Decrypt failed`). Any other error (transient keychain/IO) is rethrown instead of deleting; `PinScreen`'s existing try/catch around the call logs it and leaves the user on the PIN screen to retry, so a flaky read can no longer destroy a real wallet. This matches the same transient-vs-terminal principle already used for the encryption-key path. Added tests for the terminal-decrypt-error case (recovers) and the transient-error case (rethrows, no recovery).

packages/core-mobile/app/new/common/utils/ensureWalletSecret.test.ts
```diff
@@ -0,0 +1,66 @@
...
+    const onMissing = jest.fn()
+
+    const result = await ensureWalletSecret(walletId, onMissing)
+
+    expect(result).toBe(false)
+    expect(onMissing).toHaveBeenCalledTimes(1)
+  })
+
+  it('rethrows and does NOT recover on a transient keychain failure', async () => {
+    mockLoadWalletSecret.mockResolvedValue({
```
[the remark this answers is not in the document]
ruijia1in: Added tests in `ensureWalletSecret.test.ts` for the corrupt-secret-data terminal case. `isTerminalSecretError` now also matches `NoSaltError`/`InvalidVersionError` by `error.name` (the messages are dynamic), so a deterministic decrypt failure triggers recovery (`onMissing`) instead of being rethrown as transient and stranding the user post-PIN. Parameterized test asserts recovery runs for both names (c67b5951d).

packages/core-mobile/app/utils/BiometricsSDK.ts
```diff
@@ -177,13 +188,37 @@ class BiometricsSDK {
...
+    if (!credentials) return 'no-credentials'
+    try {
+      const decrypted = await decrypt(credentials.password, pin)
+      this.#encryptionKey = decrypted.data
+      return 'success'
+    } catch (e) {
+      if (isWrongPinError(e)) return 'wrong-pin'
+      throw e
+    }
   }
```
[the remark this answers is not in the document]
ruijia1in: Fixed at the ignored call site: `KeychainMigrator.runPinMigration` now asserts `loadEncryptionKeyWithPin(pin) === 'success'` and throws `Failed to load encryption key after migration: <result>` otherwise, instead of silently continuing to the later opaque "Encryption key not found". Verified the other two callers already check the result (`BiometricsSDK.ts:465`, `usePinOrBiometryLogin.ts:100`). Added a test for the non-success path (c67b5951d).

packages/core-mobile/app/new/common/components/PinScreen.tsx
```diff
@@ -91,17 +94,26 @@ export const PinScreen = ({
...
+          const canProceed = await ensureWalletSecret(walletId, deleteWallet)
+          if (!canProceed) {
+            // Wallet secret is gone; deleteWallet routes the user to onboarding.
+            return
           }
         }
         await unlock()
       } catch (error) {
         Logger.error('Failed to login:', error)
       }
```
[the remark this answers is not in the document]
ruijia1in: Fixed. The catch in `handleLoginSuccess` no longer just logs — it now stops the loading animation, restores the PIN input (`isProcessing.value = false`), clears the sticky verification + entered PIN via a new `resetLoginState()` exposed from `usePinOrBiometryLogin`, and refocuses the input, so a re-entered PIN re-triggers login instead of the user being stuck on a spinner. `handleStopLoading`/`resetLoginState` are memoized so the login effect stays stable on the success path (c67b5951d).

packages/core-mobile/app/new/common/hooks/usePinOrBiometryLogin.ts
```diff
@@ -97,8 +98,23 @@ export function usePinOrBiometryLogin({
         }
 
         // Load encryption key
-        const isValidPin = await BiometricsSDK.loadEncryptionKeyWithPin(pin)
-        if (!isValidPin) {
+        const pinResult = await BiometricsSDK.loadEncryptionKeyWithPin(pin)
+
+        if (pinResult === 'no-credentials') {
```
bogdandobritoiu, line 179: I traced the cold-start flow and I think this branch may be unreachable in the ticket's main scenario.

On a cold launch in the corrupt state, `RootNavigator` renders `loginWithPinOrBiometry` (`shouldRenderOnlyPinScreen` starts `true`), which mounts `PinScreen` with `isInitialLogin={true}`. So `checkEnteredPin` runs `migrator.migrateIfNeeded('PIN', pin)` **before** `loadEncryptionKeyWithPin`:

1. `getMigrationStatus('PIN')`: `hasEncryptionKeyWithPin()` → false, `hasEncryptionKeyWithBiometry()` → false → returns `RunPinMigration` (or `CompletePartialMigration` if the bio key survived the partial wipe — same outcome below).
2. `migrateIfNeeded` → `throwIfBadPin(pin)` → `isPinCorrect(pin, isLegacy=true)` → `loadLegacyWalletWithPin()` reads `LEGACY_SERVICE_KEY_BIO`, which never existed on a modern (non-legacy) install → `{ success: false }` → throws `BadPinError('Bad PIN')`.
3. The catch below treats `BadPinError` as a wrong PIN → `increaseAttempt()` + `onWrongPin()` — i.e. the endless "wrong PIN" loop from the ticket, unchanged. `loadEncryptionKeyWithPin` never runs, so `no-credentials` never fires.

Since `loadEncryptionKeyWithPin` is only reached when `getMigrationStatus` returns `NoMigrationNeeded` (which requires the encryption key to exist), on the initial-login path this branch looks reachable only via a race between `hasGenericPassword` and the subsequent `getGenericPassword`. It *is* reachable from the in-session flows (`PinScreenOverlay`, `VerifyPin`) where migration is skipped — but those aren't the cold-start case.

Could you re-verify variant A on a **cold start** with a wallet that never had legacy keychain entries? If the test device had legacy entries (migration would silently recreate the key and unlock), or the test went through the in-session overlay, that would explain the passing dev test.

If it repros, one possible fix: have `getMigrationStatus` distinguish "no new keys **and** nothing legacy to migrate from" as its own terminal status, and route that to the same delete-and-onboard recovery instead of letting `throwIfBadPin` translate it into a rate-limited wrong PIN.

packages/core-mobile/app/new/common/hooks/usePinOrBiometryLogin.ts
```diff
@@ -97,8 +98,23 @@ export function usePinOrBiometryLogin({
...
+          // The encryption key is gone (e.g. an interrupted wallet deletion),
+          // so no PIN can ever unlock this wallet. Recover by deleting the
+          // stale wallet and routing to onboarding instead of reporting an
+          // endless "wrong PIN". (CP-14585)
+          Logger.error(
+            'Encryption key missing on PIN entry; deleting wallet',
+            new Error('no-credentials')
+          )
+          onStopLoading()
+          deleteWallet()
```
bogdandobritoiu: This hook is also used by `VerifyPin` and `VerifyWithPinOrBiometry` (reveal recovery phrase, change PIN, etc.), where `isInitialLogin` is false — so this `deleteWallet()` fires in in-session verification contexts too, with no user-facing message: the user types their PIN in settings and abruptly lands on onboarding.

Consider gating this recovery to `isInitialLogin`, or at minimum showing an `alertBadData`-style alert (explain + acknowledge) before deleting, like the corrupt-data path already does.

packages/core-mobile/app/new/common/components/PinScreen.tsx
```diff
@@ -166,6 +146,53 @@ export const PinScreen = ({
     setIsEnteringPin(true)
   }, [disableKeypad])
 
+  const handleLoginSuccess = useCallback(() => {
```
bogdandobritoiu, line 149: nit: `handleLoginSuccess` now has a wider dep list (e.g. `focusPinInput` changes with `disableKeypad`), and the `useEffect(() => { if (verified) handleLoginSuccess() })` re-fires on any identity change while `verified` is still true — which would schedule `unlock()` a second time. Likelihood is low (the zustand fns are stable refs), but an early-return guard on `isProcessing.value` at the top of this callback would close it cheaply.

packages/core-mobile/app/new/common/utils/ensureWalletSecret.ts
```diff
@@ -0,0 +1,67 @@
...
+ * wallet) only when the credential is genuinely absent, cannot be decrypted, or
+ * the stored data is structurally corrupt. A transient keychain/IO error must
+ * NOT be treated as a missing secret, or a flaky read would destroy a real
+ * wallet. (CP-14585)
+ */
+const isTerminalSecretError = (error?: Error): boolean => {
+  const message = error?.message ?? ''
+  return (
+    message.includes('No credentials found') ||
+    message.includes('Failed to decrypt') ||
```
bogdandobritoiu: nit: this `'Failed to decrypt'` match is dead code — `decrypt()` (EncryptionHelper) never returns falsy (it returns data or throws), so `loadWalletSecret`'s `if (!decrypted)` branch that produces this message can't execute. Real decrypt failures surface as `BAD_DECRYPT` / `Decrypt failed`, which are already matched below. Safe to drop (here and ideally the dead branch in `loadWalletSecret`).

bogdandobritoiu (commented):
Nice mechanism overall — the three-state `loadEncryptionKeyWithPin` result, the terminal-vs-transient split in `ensureWalletSecret`, and the stuck-spinner recovery are all solid, and all three callers were updated consistently. One trace I'd like to resolve before this merges though (first inline comment): on a cold start the migrator appears to intercept the wiped-keychain state before the new `no-credentials` branch can run.

Test suggestion: a hook-level test for `usePinOrBiometryLogin` starting from the *full* corrupt state (`isInitialLogin: true`, with `KeychainMigrator.getMigrationStatus` resolving as it would with a wiped keychain) would pin down the end-to-end routing — that's the test that would catch the migration-preemption issue if it repros.

packages/core-mobile/app/new/common/hooks/usePinOrBiometryLogin.ts
B0Y3R-AVA: Confirmed — I reproduced your trace exactly: on a cold start with a wiped keychain, `getMigrationStatus` returned `RunPinMigration`, `throwIfBadPin` → `loadLegacyWalletWithPin` found no `LEGACY_SERVICE_KEY_BIO` entry and translated it into `BadPinError` → the rate-limited wrong-PIN loop. The `no-credentials` branch was unreachable on the cold-start path; only the in-session flows (or a read race) could hit it.

Fixed as you suggested (abd78a18f): `getMigrationStatus` now returns a terminal `MigrationStatus.NoKeychainData` when there are no new keys **and** nothing legacy to migrate from, and `migrateIfNeeded` surfaces it without attempting a migration. Both the PIN and BIO initial-login paths route it to delete-and-onboard recovery (behind an acknowledge alert — see the other thread).

The "nothing legacy" check is a new strict `BiometricsSDK.hasAnyWalletData()` probe that **throws** on transient keychain failures instead of returning false, so a flaky read can never present as a wiped wallet. Verified against react-native-keychain's native sources: iOS `hasCredentialsWithSecClass` resolves true even on `errSecInteractionNotAllowed` (present-but-locked, so a bio-protected entry never reads as absent) and rejects any unexpected `OSStatus`; Android reads the DataStore prefs entry rather than touching the keystore.

Also added the hook-level test you suggested: `usePinOrBiometryLogin.test.ts` renders the hook with `isInitialLogin: true` in the full corrupt state and pins the end-to-end routing — recovery alert + `deleteWallet` on acknowledge, no `onWrongPin`, no rate-limiter increment, `loadEncryptionKeyWithPin` never reached; plus the wrong-pin/success/BIO variants.

One related hardening that fell out of an adversarial pass over this branch: the migrations now write the wallet secret *before* the new encryption keys (and `runPinMigration`'s redundant post-store reload-assert is gone), so an app death mid-migration leaves the migration re-runnable from legacy data instead of a key-without-secret state that would read as a wiped wallet after unlock. `ensureWalletSecret` additionally refuses destructive recovery while legacy entries still exist.

packages/core-mobile/app/new/common/hooks/usePinOrBiometryLogin.ts
B0Y3R-AVA: Addressed in abd78a18f, in both directions:

1. Every recovery deletion is now behind an `alertBadData`-style acknowledge alert ("Wallet data not found" → Okay → delete), so an in-session `VerifyPin` can never silently drop the user into onboarding.
2. The in-session case you pointed at turned out to be genuinely destructive, not just jarring: a partially completed biometric migration (`runBiometricMigration` intentionally leaves the new bio key with no PIN key until the next initial PIN login — which for a biometrics user can be arbitrarily far away) reads as `no-credentials` from `VerifyPin`, and deleting there would wipe the legacy entries `completePartialMigration` still needs. The `no-credentials` branch is now gated on the strict `hasAnyWalletData()` probe: recovery is only offered when the keychain is provably empty; if any other credential exists (bio key, legacy entries) it surfaces as a failed PIN check — the pre-CP-14585 behavior for that state — rather than an offer to delete.

Both cases are covered in the new `usePinOrBiometryLogin.test.ts` (empty keychain → alert-gated recovery; partial-bio state → no alert, no delete, wrong-PIN handling).

I kept recovery available in-session (rather than gating on `isInitialLogin`) because when the keychain is provably empty the wallet can never be unlocked by any flow, and the alert now makes that explicit wherever the user happens to discover it.

packages/core-mobile/app/new/common/components/PinScreen.tsx
B0Y3R-AVA: Added — `handleLoginSuccess` now early-returns when `isProcessing.value` is already true, so a callback-identity change while `verified` is still true can't schedule a second `unlock()` (abd78a18f).

packages/core-mobile/app/new/common/utils/ensureWalletSecret.ts
B0Y3R-AVA: Dropped, both here and the dead `if (!decrypted)` branch in `loadWalletSecret` (with a note that `decrypt` returns data or throws, never falsy). Left the equivalent dead branch in `loadLegacyWalletWithPin` untouched to keep this diff scoped — same pattern, can go in a separate cleanup (abd78a18f).

gergelylovas (approved):

PR #3992, opened 2026-07-17
CP-14832: persist advanced filter toggles as set and exclude fast stakes from the Delegation filter

onghwan:
## Description

**Ticket: [CP-14832](https://ava-labs.atlassian.net/browse/CP-14832)**

Two stake filter fixes:

### 1. Advanced filter toggles persist as set (CP-14832)

The Advanced filters sheet used to treat each row's toggle as "the user has deviated from the web default": enabling a filter at its default value collapsed back to "off" on reopen, and the active-filter badge ignored it. QA read this as filters silently not applying — they enabled all four filters, yet only one showed as applied.

The sheet's toggles now persist exactly as set:

- The filter store is split into two layers: `filters` (the user's explicit toggles/values, mirrored verbatim by the sheet) and `defaults` (the web-parity baseline seeded on flow entry — uptime ≥ 75%, fee ≤ the network minimum, min time remaining).
- `resolveEffectiveDelegateFilters` computes what actually narrows the node list: a user-enabled dimension replaces the baseline (so loosening, e.g. uptime below 75%, genuinely widens the list); untouched dimensions fall back to it.
- The badge counts the filters the user toggled on (`countEnabledFilters`), matching what the sheet shows, instead of diffing against defaults.
- A fresh flow still opens with every toggle off and the baseline filtering underneath — no visual change to the default picker.

### 2. Stake home: the Delegation chip no longer shows fast stakes

The Delegation filter matched on txType alone, and fast stakes ARE delegation transactions on-chain — so cards wearing the "Fast stake" badge showed up under Delegation, making the chip look broken. The filter is now mutually exclusive with Fast stakes (`isDelegationTx && !isFastStakeTx`), matching the card badges exactly. Fee-less fast stakes carry no escrow output, so — like their badge — they stay under Delegation.

No new dependencies. State is in-memory zustand only (reseeded per flow), so no migration.

## Screenshots/Videos

N/A — behavior changes are in filter logic; see the recording on the ticket for the original issue.

## Testing
iOS: 9325
Android: 9326

Advanced filters:
1. Start a Delegate flow and open Advanced filters — every toggle is off; the node list is pre-filtered as before.
2. Enable all four filters without changing values, tap Apply — the badge shows 4.
3. Reopen the sheet — all four toggles are still on with their values.
4. Enable Uptime range and drag below 75% — the node list widens (user filter replaces the baseline).
5. Toggle a filter off and apply — that dimension falls back to the web-parity baseline (the list does not blow up to every node).

Stake home chips:
1. With both fast stakes and plain delegations on the account, select the Delegation chip — only cards with the "Delegating" badge appear.
2. Select Fast stakes — only "Fast stake" badge cards appear.

- The scenario from the ticket: enabling filters (even at their default values) must persist — the sheet reopens exactly as applied and the badge counts every enabled filter.
- The Delegation and Fast stakes chips on the stake home are mutually exclusive, matching the card badges.

## Checklist

[CP-14832]: https://ava-labs.atlassian.net/browse/CP-14832?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core-mobile/app/new/features/stake/v2/screens/AdvancedFiltersScreen.tsx
+++ b/packages/core-mobile/app/new/features/stake/v2/screens/AdvancedFiltersScreen.tsx
@@ -105,55 +105,41 @@ const FilterCard: FC<FilterCardProps> = ({
   )
 }
 
-// Whether an applied `{enabled, value}` filter deviates from its seeded
-// default. Drives each row's toggle: a row reads as "on" only when the user has
-// overridden the web default.
-const isOverridden = (
-  applied: { enabled: boolean; value: number },
-  base: { enabled: boolean; value: number }
-): boolean =>
-  applied.enabled !== base.enabled ||
-  (applied.enabled && applied.value !== base.value)
-
 /**
- * Builds an editable draft from the currently applied filters. Here a row's
- * `enabled` flag means "the user is overriding the web default" (it drives the
- * toggle + whether the value control shows), NOT "the filter is active" — the
- * web-parity defaults (uptime ≥ 75%, fee ≤ 2%, min time remaining) always apply
- * underneath. So a row reads as toggled on only when its applied value differs
- * from the seeded default; an untouched picker opens with every toggle off even
- * though it's pre-filtered. Numeric values are kept inside the live bounds, and
- * a row sitting at its default seeds its control at the default value so
- * flipping the toggle on starts there (then drag up to tighten or down to
- * loosen, e.g. uptime to 0% to see every node).
+ * Builds an editable draft from the user's saved filters. Toggles mirror
+ * those filters verbatim — a fresh flow opens with every toggle off (the
+ * web-parity baseline still filters underneath; see
+ * `resolveEffectiveDelegateFilters`), and whatever the user applies is
+ * exactly what they see on reopen. A user-enabled dimension replaces the
+ * baseline, so e.g. enabling uptime and dragging it below 75% genuinely
+ * loosens the list. Numeric values are kept inside the live bounds; a
+ * disabled row seeds its control at the baseline value so flipping the
+ * toggle on starts there.
  */
 const seedDraft = (
   applied: DelegateFilters,
-  defaults: DelegateFilters,
   bounds: DelegateFilterBounds
 ): DelegateFilters => ({
   uptime: {
-    enabled:
-      applied.uptime.enabled !== defaults.uptime.enabled ||
-      (applied.uptime.enabled && applied.uptime.min !== defaults.uptime.min),
+    enabled: applied.uptime.enabled,
     min: clamp(applied.uptime.min, bounds.uptime.min, bounds.uptime.max)
   },
   maxFee: {
-    enabled: isOverridden(applied.maxFee, defaults.maxFee),
+    enabled: applied.maxFee.enabled,
     // Free numeric filter: allow any fee down to 0% (a value below the network
     // floor simply yields no matches) and retain exactly what was entered on
     // reopen — only the upper bound (100%) is enforced.
     value: clamp(applied.maxFee.value, 0, MAX_FEE_INPUT)
   },
   minAvailable: {
-    enabled: isOverridden(applied.minAvailable, defaults.minAvailable),
+    enabled: applied.minAvailable.enabled,
     value:
       applied.minAvailable.value > 0
         ? Math.max(applied.minAvailable.value, bounds.minAvailable.min)
         : bounds.minAvailable.min
   },
   minTimeRemaining: {
-    enabled: isOverridden(applied.minTimeRemaining, defaults.minTimeRemaining),
+    enabled: applied.minTimeRemaining.enabled,
     value:
       applied.minTimeRemaining.value > 0
         ? clamp(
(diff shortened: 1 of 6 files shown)


packages/core-mobile/app/new/features/stake/v2/screens/AdvancedFiltersScreen.tsx
```diff
@@ -105,55 +105,41 @@ const FilterCard: FC<FilterCardProps> = ({
...
- * loosen, e.g. uptime to 0% to see every node).
+ * Builds an editable draft from the user's saved filters. Toggles mirror
+ * those filters verbatim — a fresh flow opens with every toggle off (the
+ * web-parity baseline still filters underneath; see
+ * `resolveEffectiveDelegateFilters`), and whatever the user applies is
+ * exactly what they see on reopen. A user-enabled dimension replaces the
+ * baseline, so e.g. enabling uptime and dragging it below 75% genuinely
+ * loosens the list. Numeric values are kept inside the live bounds; a
+ * disabled row seeds its control at the baseline value so flipping the
+ * toggle on starts there.
```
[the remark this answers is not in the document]
onghwan: Good catch! The behavior (retaining the last saved value on re-enable) is intentional — it matches the maxFee inline note ("retain exactly what was entered on reopen") and is friendlier than snapping back to the baseline. Updated the docstring to describe it accurately: a disabled row keeps its last saved value (baseline on a fresh flow, range minimum when it has none). Fixed in 4ea3aad25.

packages/core-mobile/app/new/features/stake/v2/store.test.ts
```diff
@@ -0,0 +1,80 @@
...
+    expect(effective.minTimeRemaining).toEqual(BASELINE.minTimeRemaining)
+  })
+})
+
+describe('countEnabledFilters', () => {
+  it('is 0 right after seeding, so the picker opens without a badge', () => {
+    useDelegateFilters.getState().seedDefaults(BASELINE)
+    expect(countEnabledFilters(useDelegateFilters.getState().filters)).toBe(0)
+  })
+
```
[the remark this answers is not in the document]
onghwan: Good catch! Replaced the per-describe reset with a single file-level afterEach that resets the shared store, so no describe block can leak seeded state into the next. Fixed in 4ea3aad25.

packages/core-mobile/app/new/features/stake/v2/screens/AdvancedFiltersScreen.tsx
```diff
@@ -168,34 +155,20 @@ const seedDraft = (
...
-        ? { enabled: true, value: draft.minAvailable.value }
-        : defaults.minAvailable,
-      minTimeRemaining: draft.minTimeRemaining.enabled
-        ? { enabled: true, value: draft.minTimeRemaining.value }
-        : defaults.minTimeRemaining
-    })
+    // Saved verbatim, toggles included, so the sheet reopens exactly as
+    // applied. A row toggled off falls back to the web-parity baseline for
+    // that dimension (it does NOT list every node).
+    setFilters(draft)
```
B0Y3R-AVA, line 169: Non-blocking, flagging for awareness: saving the whole draft verbatim means rows the user never touched also get persisted with their bound-clamped display values — so repeated open/apply of the sheet can silently drift a *disabled* row's stored `value` to whatever the live bounds clamp it to. Harmless for filtering (disabled rows are ignored by `resolveEffectiveDelegateFilters`) and consistent with the "resume where you left off" intent, but it's a change from the old revert-to-defaults-on-apply behavior — just confirming it's intended.

packages/core-mobile/app/new/features/stake/v2/screens/AdvancedFiltersScreen.tsx
```diff
@@ -105,55 +105,42 @@ const FilterCard: FC<FilterCardProps> = ({
...
   },
   maxFee: {
-    enabled: isOverridden(applied.maxFee, defaults.maxFee),
+    enabled: applied.maxFee.enabled,
     // Free numeric filter: allow any fee down to 0% (a value below the network
     // floor simply yields no matches) and retain exactly what was entered on
     // reopen — only the upper bound (100%) is enforced.
     value: clamp(applied.maxFee.value, 0, MAX_FEE_INPUT)
   },
   minAvailable: {
```
B0Y3R-AVA, line 135: Pre-existing nit (not introduced here, just adjacent): `minAvailable` is only floored — no ceiling clamp here, and its `FilterNumberInput` has no `max` prop, unlike the other numeric rows. Fine to leave as-is or fold into a follow-up.

B0Y3R-AVA (commented):
Ran an adversarial review over this (two independent reviewers + adjudication) — the core changes check out: toggle persistence round-trips correctly through the store split, `resolveEffectiveDelegateFilters` falls back to the baseline (not "no filter") for disabled dimensions, the badge matches the sheet, no dangling `countModifiedFilters` references, and the Delegation/Fast-stakes exclusion classifies identically to the card badges (same `selectIsDeveloperMode` input on both paths). No blocking issues.

One pre-existing edge worth a follow-up ticket (not this PR's regression — it reproduces on `main` too): the sheet clamps displayed values to the *current* bounds while the node list filters on the *raw stored* values. If bounds move mid-flow (dev-mode flip Fuji ↔ mainnet with e.g. min time remaining enabled at 1 day), the reopened sheet shows the clamped 14 days but the list keeps filtering at 1 day, and Cancel preserves the hidden looser filter. Narrow (dev-mode only) and consistent with the flow's existing entry-time-values-stick policy, so fine to punt.

Two non-blocking inline comments below.

packages/core-mobile/app/new/features/stake/v2/screens/AdvancedFiltersScreen.tsx
onghwan: Good observation — confirmed intended. Two things bound the drift in practice: disabled dimensions are ignored by `resolveEffectiveDelegateFilters`, and the store is in-memory and reseeded by `applyDefaultDelegateFilters` on every flow entry, so any clamped value only lives within a single delegate session. Keeping the verbatim save preserves the "resume where you left off" behavior for values the user actually typed.

packages/core-mobile/app/new/features/stake/v2/screens/AdvancedFiltersScreen.tsx
onghwan: Agreed — the missing upper clamp on `minAvailable` predates this PR (the seeding logic moved here unchanged), and an oversized value just yields zero matches rather than misbehaving. Leaving it as-is here; happy to fold a `max` (e.g. from the live bounds) into a follow-up alongside any other filter-input polish.

B0Y3R-AVA (approved):

ruijia1in (approved):
dev-tested

PR #3993, opened 2026-07-20
CP-14749: show stake complete notifications in the notification center

onghwan:
## Description

**Ticket: [CP-14749](https://ava-labs.atlassian.net/browse/CP-14749)**

Shows stake-complete notifications in the notification center, sourced from the local push notifications this device schedules.

**How it works**

- A new zustand-persist store (`stakeCompleteNotificationRecords`) records `{txHash, endTimestamp, accountId, isDeveloperMode}` whenever the scheduling listener creates stake-complete push triggers (Notifee only keeps PENDING triggers, so fired pushes would otherwise be unrecoverable). A record whose end time has passed IS a fired push — the center derives its items purely from time, so it works even though iOS doesn't wake the app on local-notification delivery.
- The center lists fired records from the last 30 days (capped at 20), labeled with the owning account and a "Testnet" tag for cross-environment items. The push scheduler covers the active wallet's accounts on BOTH networks, and the center mirrors that.
- Tapping mirrors the push-tap semantics end to end: switch environment if needed → switch wallet/account (`setActiveAccount` handles cross-wallet activation) → open the stake detail. `handleProcessNotificationData` also gained the `isDeveloperMode` branch so tapping the OS push itself switches environments too.
- Swiping an item (or Clear All) deletes its record for good — no separate dismissed-set needed.

**Why local records instead of deriving from on-chain stakes:** the center stays 1:1 with the pushes the user actually received — it respects the notification settings, never backfills stakes that predate the feature, and needs no Glacier polling. The trade-off (device-local, lost on reinstall) matches the pushes themselves; stake history remains available under the stake home's Completed filter.

**Lifecycle guards**

- Records are only written when scheduling actually happens (earn + notification-settings guards), and not-yet-fired records are dropped when pending triggers get cancelled.
- Items are hidden while earn is feature-flag blocked (same guard as the push deeplink path).
- Records whose account no longer exists are dropped (accounts vanish with their wallet, so this also covers wallet removal).
- Wallet-scoped zustand stores (stake records + fusion transfers) are now cleared in the `onLogOut` listener — this also covers forced logouts (session-expired flow), which previously leaked fusion transfers to the next wallet on the device.

**Adjacent fix:** `NotificationsService.cleanupNotifications` fed Notifee's millisecond trigger timestamps into `fromUnixTime` (which expects seconds), mapping every trigger to the far future — the stale-trigger cleanup silently never cancelled anything. Now compares in ms.

## Screenshots/Videos

https://github.com/user-attachments/assets/082a5911-f22f-4f39-93e9-85e82c09d556

## Testing
iOS: 9339
Android: 9340

Happy path:
1. Stake with a short duration and let it complete — the local push fires and the same item appears in the notification center (All / Transactions tabs), labeled with the account.
2. Tap the item — the app switches to the stake's environment/account if needed and opens the stake detail.
3. Swipe the item away — it's gone for good (survives reopening the center).

Edge cases:
- Testnet stake viewed from mainnet: item shows a "Testnet" tag; tapping switches to testnet first (mirrors tapping the push).
- Delete the wallet / log out: the center is empty for the next wallet.

- The center should list exactly the stake-complete pushes this device received in the last 30 days — items appear at the moment the push fires, not before.
- Tap = same behavior as tapping the push itself (environment + account switch, then stake detail).
- Dismissal is permanent; Clear All includes stake items.
- Notification center items are device-local: reinstalling the app or switching devices starts the list fresh (stake history is still on the stake home's Completed filter).

## Checklist

[CP-14749]: https://ava-labs.atlassian.net/browse/CP-14749?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/core-mobile/app/store/notifications/listeners/handleScheduleStakingCompleteNotifications.ts
```diff
@@ -28,6 +30,27 @@ export const handleScheduleStakingCompleteNotifications = async (
...
+              isDeveloperMode: data.isDeveloperMode ?? false
+            }
+          ]
+        : []
+    )
+  )
+
   await NotificationsService.updateStakeCompleteNotification(
     stakeCompleteNotification
   )
```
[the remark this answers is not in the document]
onghwan: Good catch! Records without an accountId could never render anyway (the center labels rows by account and drops records whose account is unknown), so the upsert now requires `data.accountId` instead of falling back to `''`. The push-scheduling fallback in `updateStakeCompleteNotification` is pre-existing behavior and unchanged. Fixed in fa1985e3e.

packages/core-mobile/app/store/notifications/listeners/handleScheduleStakingCompleteNotifications.ts
```diff
@@ -28,6 +30,27 @@ export const handleScheduleStakingCompleteNotifications = async (
     return
   }
 
+  // Record what gets scheduled so the notification center can list the
+  // pushes after they fire (Notifee only keeps PENDING triggers). Upserting
+  // the full batch — not just newly created triggers — also backfills
+  // records for triggers scheduled before this store existed.
+  // `endTimestamp` arrives in unix seconds (same conversion as
```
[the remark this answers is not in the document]
onghwan: Good catch — this was the most important finding on the PR. The periodic scheduler is the main producer of these triggers, so bypassing the listener meant most fired pushes never got recorded. It now dispatches `scheduleStakingCompleteNotifications` instead of calling the service directly, so every scheduling request flows through the single recording path (the scheduler had already run the same earn/settings guards, so behavior is otherwise unchanged). Fixed in fa1985e3e.

packages/core-mobile/app/new/features/notifications/screens/NotificationsScreen.tsx
```diff
@@ -332,6 +418,31 @@ export const NotificationsScreen = (): JSX.Element => {
...
+      }
+      router.dismiss()
+      router.navigate({
+        // @ts-ignore route is defined under (modals)/stakeDetail
+        pathname: '/stakeDetail',
+        params: { txHash: item.txHash, source: 'deeplink' }
+      })
+    },
+    [activeAccount?.id, isDeveloperMode, dispatch]
+  )
```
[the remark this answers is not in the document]
onghwan: This screen uses the module-level `router` singleton (`import { router } from 'expo-router'`, line 7), not the `useRouter()` hook — module imports aren't reactive values, so `react-hooks/exhaustive-deps` doesn't require them in the array and there's no stale-closure risk. ESLint runs clean on this file.

packages/core-mobile/app/new/features/notifications/hooks/useStakeCompleteNotifications.test.ts
```diff
@@ -0,0 +1,169 @@
+import {
+  StakeCompleteNotificationRecord,
+  stakeCompleteNotificationRecordsStore
+} from '../store/stakeCompleteNotificationRecords'
+import {
+  deriveStakeCompleteNotifications,
+  STAKE_COMPLETE_NOTIFICATION_MAX_ITEMS,
+  STAKE_COMPLETE_NOTIFICATION_WINDOW_DAYS
+} from './useStakeCompleteNotifications'
```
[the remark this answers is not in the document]
onghwan: The suite runs green as-is (repeatedly verified — 10/10 pass) without mocking `utils/mmkv`: react-native-mmkv operates in-memory under Jest in this repo's setup. The suites that do mock `utils/mmkv` do so for their own reasons; it isn't required here, and the passing runs are the evidence.

packages/core-mobile/app/new/features/notifications/screens/NotificationsScreen.tsx
```diff
@@ -246,19 +326,25 @@ export const NotificationsScreen = (): JSX.Element => {
     clearAllTimerRef.current = setTimeout(() => {
       markAllAsRead()
       if (isFusionEnabled) clearCompletedTransfers()
+      if (stakeCompleteItems.length > 0) {
+        dismissStakeNotifications(stakeCompleteItems)
+      }
```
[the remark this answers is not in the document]
onghwan: Good catch! Clear All now calls a new `removeFired(now)` store action that drops every fired record — not just the capped rendered page — so nothing resurfaces after clearing. Not-yet-fired records stay for future pushes, and per-item swipe dismissal is unchanged. Covered with a unit test (30 fired records + 1 pending → only the pending one survives). Fixed in e5440a039.

packages/core-mobile/app/store/notifications/listeners/handleScheduleStakingCompleteNotifications.ts
```diff
@@ -28,6 +30,29 @@ export const handleScheduleStakingCompleteNotifications = async (
...
+  // `scheduleNotification`). An `accountId` is required: the center labels
+  // rows by account and drops records whose account is unknown, so a record
+  // without one could never render.
```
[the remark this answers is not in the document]
onghwan: Good catch! The real unit is SECONDS — the dispatch sites pass Glacier'\''s `PChainTransaction.endTimestamp` / `getUnixTime()` values, and `scheduleNotification` converts via `fromUnixTime` (which expects seconds). The "milliseconds" note on `scheduleNotification`'\''s param was a pre-existing doc error; corrected it to seconds and made this comment state the source unit explicitly instead of cross-referencing. Fixed in fab7d35c8.

packages/core-mobile/app/new/features/notifications/hooks/useStakeCompleteNotifications.ts
```diff
@@ -0,0 +1,108 @@
...
+        : deriveStakeCompleteNotifications({
+            records,
+            accounts,
+            now: Date.now()
+          }),
+    [isEarnBlocked, records, accounts]
+  )
+
+  return { items }
+}
```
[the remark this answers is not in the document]
onghwan: Good catch — and it was actually worse than a stale render: `now` wasn'\''t a memo dependency, so even the re-renders from the backend feed'\''s 1-minute poll couldn'\''t recompute the list; only a remount could. The hook now keeps a `now` state ticked by a 60s interval (alive only while the center screen is mounted) and includes it in the derivation deps, so items surface within a minute of the push firing. React import updated accordingly. Fixed in 840661634.

packages/core-mobile/app/store/notifications/listeners/handleScheduleStakingCompleteNotifications.ts
```diff
@@ -28,6 +30,27 @@ export const handleScheduleStakingCompleteNotifications = async (
     return
   }
 
+  // Record what gets scheduled so the notification center can list the
+  // pushes after they fire (Notifee only keeps PENDING triggers). Upserting
+  // the full batch — not just newly created triggers — also backfills
+  // records for triggers scheduled before this store existed.
+  // `endTimestamp` arrives in unix seconds (same conversion as
+  // `scheduleNotification`).
+  stakeCompleteNotificationRecordsStore.getState().upsert(
```
B0Y3R-AVA, line 42: **Critical: the periodic scheduler never writes records, so the center will miss most real pushes.**

This listener only runs off the `scheduleStakingCompleteNotifications` action, which is dispatched from exactly two places — `StakeConfirmScreen.tsx:529` and `addStake/confirm.tsx:270`, i.e. only when a brand-new stake is confirmed in-app. The authoritative scheduler, `scheduleNotificationsForActiveStakesPeriodically.ts:123` (every app unlock + every 3 min, all accounts), calls `NotificationsService.updateStakeCompleteNotification()` **directly** and never goes through this listener, so it creates triggers without records. Concretely:

- **Existing stakes at ship time**: user gets the push (trigger path untouched) but no center item — the comment above about backfilling "triggers scheduled before this store existed" never actually fires for them, because the backfill only runs through the confirm-screen action.
- **Restakes**: `useRestake` has no notification wiring at all; it relies on the periodic rescan → push fires, no record, no center item.
- **`removePending` drops are permanent**: toggle stake notifications off→on (or an earn-flag blip) and `handleStakeCompleteNotificationCleanup` drops the pending records; the 3-min loop then re-schedules the *pushes* but nothing re-records them. The store's comment "the scheduler's next pass re-records them" doesn't hold — that pass is the periodic one.
- **Every Android 16+ device hits this once unconditionally**: the one-time Notifee-bug cleanup branch in `handleNotificationCleanup.ts` now also calls `removePending`, wiping all pending records while the pushes get silently rescheduled minutes later.

No test exercises either scheduling listener, so the gap is invisible to the suite. Suggested direction: write records where triggers are actually created (e.g. upsert inside `updateStakeCompleteNotification`, or have the periodic path dispatch this same action). Note `getTransformedStakesForAllAccounts` fetches only the current mode (`isTestnet: isDeveloperMode`), so the PR description's "covers BOTH networks" claim would need a second look there too.

Smaller related note: the upsert here runs before `updateStakeCompleteNotification`, whose scheduling happens in fire-and-forget `setTimeout`s — a Notifee failure leaves a record for a push that never fires. Narrow (the permission/settings guards run first), but worth knowing if scheduling errors ever show up.

packages/core-mobile/app/new/features/notifications/screens/NotificationsScreen.tsx
```diff
@@ -332,6 +418,31 @@ export const NotificationsScreen = (): JSX.Element => {
...
   )
 
+  // Mirrors the push-notification tap semantics
+  // (`handleProcessNotificationData`): stake-complete pushes carry the
+  // scheduling environment and an accountId, and tapping them switches
+  // both BEFORE navigating. The center's rows do the same — the item list
+  // spans both environments (the underlying service feeds the cross-mode
+  // push scheduler), so without the mode switch a testnet stake tapped
+  // from mainnet lands on a stake detail that can never resolve.
+  const handleStakeCompletePress = useCallback(
```
B0Y3R-AVA, line 436: **Cross-environment tap can land on a loading/not-found detail screen.** `setActiveAccount` itself is synchronous (verified — cross-wallet switch included), but `toggleDeveloperMode()` kicks off the async `reloadAccounts` listener (`store/account/listeners.ts`) that re-derives addresses before `setAccounts` lands, and `useStake(txHash)` is keyed off the active account/network. Tapping a testnet item from mainnet can mount `StakeDetailScreen` before the reload finishes → transient spinner/not-found until queries catch up.

This mirrors the existing push-tap flow (`handleProcessNotificationData` has the same dispatch-then-navigate shape), so it's an inherited pattern rather than a regression — flagging because this PR adds a second call site for it. Fine to ship as-is if the transient state is acceptable, but worth a quick manual check of the testnet-item-from-mainnet tap.

packages/core-mobile/app/new/features/notifications/screens/NotificationsScreen.tsx
```diff
@@ -246,19 +326,25 @@ export const NotificationsScreen = (): JSX.Element => {
     clearAllTimerRef.current = setTimeout(() => {
       markAllAsRead()
       if (isFusionEnabled) clearCompletedTransfers()
+      if (stakeCompleteItems.length > 0) {
+        dismissStakeNotifications(stakeCompleteItems)
```
B0Y3R-AVA: Two Clear All edge cases (minor):

1. This checks the raw `stakeCompleteItems` rather than the tab-filtered `clearableItems`, so Clear All from the Price Updates tab — where no stake row is visible — still wipes every stake record. Consistent with `clearCompletedTransfers()` above, so maybe intentional, but calling it out.
2. `stakeCompleteItems` is capped at 20, so for a user with >20 fired records in the 30-day window, Clear All dismisses the visible 20 and the next-oldest batch surfaces immediately after. Clearing by derivation input (all fired record txHashes) instead of the capped output would make it stick.

packages/core-mobile/app/new/features/notifications/components/StakeCompleteItem.tsx
```diff
@@ -0,0 +1,67 @@
...
+  const {
+    theme: { colors }
+  } = useTheme()
+
+  const baseSubtitle = 'Your staking period has ended'
+  // Cross-environment items are tappable from either mode (tapping switches
+  // the app), so testnet ones carry an explicit marker.
+  const parts = [
+    accountLabel,
+    item.isDeveloperMode ? 'Testnet' : null,
```
B0Y3R-AVA: nit: the environment label is asymmetric — testnet items get a "Testnet" tag, but a mainnet item viewed *from* testnet mode is indistinguishable from a same-environment row, with no cue that tapping it flips the app back to mainnet. Cosmetic, but since the list intentionally spans both environments, a mirrored "Mainnet" tag (or tagging any cross-environment item relative to the current mode) would make the switch less surprising.

B0Y3R-AVA (commented):
Ran an adversarial review over this (two independent passes, findings verified against source). The design is solid and most of the PR's claims held up under scrutiny — both timestamp-unit fixes are correct (Notifee's `TimestampTrigger.timestamp` is documented as ms; `endTimestamp` producers are unix seconds), `setActiveAccount` does handle cross-wallet switching synchronously, the window/cap math and `removePending` predicate are right and well-tested, and moving the store clears into the `onLogOut` listener genuinely fixes the session-expired leak.

One finding is significant though: **records are only ever written at stake-confirm time, but the authoritative scheduling path is the periodic scanner, which bypasses the new store entirely** (inline comment on `handleScheduleStakingCompleteNotifications.ts`). I'd address that one before merge — details inline, plus a few smaller notes.

packages/core-mobile/app/store/notifications/listeners/handleScheduleStakingCompleteNotifications.ts
onghwan: Good catch — this was real and is fixed in b70d6a9cc: the periodic scheduler now dispatches `scheduleStakingCompleteNotifications` instead of calling the service directly, so every scheduling request (periodic rescan included) flows through this recording listener. That also resolves the derived scenarios: existing stakes at ship time get recorded on the first periodic pass, restakes are covered both by the confirm-screen dispatch and the rescan, and `removePending` drops (settings toggle, earn blip, the Android 16 one-time cleanup) are re-recorded by the next pass — the store comment's "scheduler's next pass re-records them" now actually holds. The test gap is closed too: `handleScheduleStakingCompleteNotifications.test.ts` (added in 3e2c2503f) covers the record write (seconds → ms), the accountId requirement, and both guard paths.

packages/core-mobile/app/new/features/notifications/screens/NotificationsScreen.tsx
onghwan: Agreed with the analysis — the transient loading window exists and is inherited from the push-tap flow (`handleProcessNotificationData` has the same dispatch-then-navigate shape). We manually verified the exact scenario (tapping a testnet item from mainnet) on device: the mode switches, the detail shows a brief spinner while accounts reload, then resolves correctly. Keeping as-is for parity with the push path; if we later want to harden it, both call sites should get the same fix.

packages/core-mobile/app/new/features/notifications/screens/NotificationsScreen.tsx
onghwan: 1. Intentional consistency — Clear All is "empty the inbox", not "empty this tab": `markAllAsRead()` clears every backend kind and `clearCompletedTransfers()` clears swaps regardless of the active tab, so stake records follow the same semantics. 2. Fixed in 1604aed06 — Clear All now calls `removeFired(now)` on the store (every fired record, not the capped rendered page), with a unit test covering the >cap case (30 fired + 1 pending → only the pending record survives).

packages/core-mobile/app/new/features/notifications/components/StakeCompleteItem.tsx
onghwan: Good point — implemented the relative version in 3e2c2503f: any item from the OTHER environment gets tagged with its environment name ('Testnet' from mainnet, 'Mainnet' from testnet), and same-environment items carry no tag. That makes the tag exactly a "tapping this switches your environment" cue, symmetric in both directions.

B0Y3R-AVA (approved):
All review findings verified as fixed against the current head (3e2c250):

- **Periodic scheduler bypass (the critical one)**: now dispatches `scheduleStakingCompleteNotifications` so every scheduling request flows through the recording listener. Confirmed `getTransformedStakesForAllAccounts` carries `accountId`/`isDeveloperMode` per stake and genuinely covers both networks, so existing stakes, restakes, and post-`removePending` recovery (incl. the Android 16 one-time cleanup) all get recorded on the next periodic pass. The new listener tests pin the recording behavior, the seconds→ms conversion, the accountId requirement, and both guards.
- **Clear All**: `removeFired(now)` drops every fired record, not just the capped page — with a unit test.
- **Stale derivation**: the ticking `now` fix is a nice catch beyond the original finding — items now surface within a minute while the center stays open, interval properly scoped to mount.
- **Environment tag**: symmetric relative tagging is exactly the right semantics ("tapping this switches your environment").
- **Tap-flow transient loading**: agreed with keeping parity with the push path given the on-device verification.

CI green. Nice work — approving.

ruijia1in (approved):

PR #3994, opened 2026-07-20
CP-14604: reject WalletConnect signing requests from accounts not granted to the session

bogdandobritoiu:
## Description

**Ticket: [CP-14604](https://ava-labs.atlassian.net/browse/CP-14604)**

Restores WalletConnect request-time session-account authorization that was lost in the vm-module migration (CP-8450 left a TODO, CP-8365/CP-8916 removed the legacy `isAddressApproved` check). Since then, a connected dApp could request signing with **any** account in the active wallet, including accounts the user never granted to that dApp's session.

This is the follow-up Bogdan split out of CP-14468. PR #3906 closed the cross-wallet case (different wallet, same index) with the `account.walletId !== walletId` guard. Two accounts in the *same* wallet share a `walletId`, so that guard waves them through, and this PR closes that remaining hole.

Worth flagging for reviewers: CP-14604's ticket description is Eunji's jup.ag regression checklist copied from CP-14468. Steps 1 to 5 are a non-regression check, not a reproduction of the bug. The happy path was never broken on main.

**Changes:**

- `isAccountApproved.ts`: adds `isAddressApproved(address, caip2ChainId, namespaces)`, an address-based sibling of the existing (previously unused) helper. Checks the address against the session's granted accounts for the request's CAIP-2 namespace, any chain within the namespace. Case-insensitive for EVM, exact match for non-EVM addresses, since Solana base58 and Avalanche bech32 are case-sensitive. `isAccountApproved` now delegates to it.
- `ApprovalController.requestApproval`: when `signingData` carries an `account` (the module-resolved signer, covering EVM tx, all EVM message signing, all 3 Solana methods, both Bitcoin methods) and the request is not in-app, the WC session for the request's topic is looked up and the request is rejected with `providerErrors.unauthorized('Requested address is not authorized')`, the same error as pre-migration, before the approval screen opens.
- `isAvalancheSignMessageAuthorized.ts` plus a call site in `handleRequestViaVMModule`: `avalanche_signMessage` is the one signing method whose signer is chosen by a dApp-supplied account *index* rather than an address, so its `signingData` carries `accountIndex` and never reaches the check above. This resolves the signer the same way the approval screen does and checks it against the session grant at dispatch time.
- `ApprovalController.onApprove`: re-checks `avalanche_signMessage` at approval time, because that method picks its signer from live state and could otherwise be outrun by a programmatic account switch between dispatch and approval.
- `ApprovalScreen/utils.ts` and `ApprovalScreen.tsx`: new `getDisplayAccountAddress()`. The `avalanche_signMessage` approval sheet previously showed no account at all, only Network and the message, because the avalanche module emits only `network` where the SVM module emits an `addressItem`. On a ticket whose entire subject is "the signed address must match the displayed address", the user could not actually perform that check. This falls back to the account the screen already resolved via `getAccountSelector`, so the address shown is the same object handed to `onApprove` and displayed equals signer by construction rather than by two sources agreeing. Scoped to `avalanche_signMessage` so `solana_signMessage` does not gain a duplicate row.

**Design notes:**

- Enforced at the ApprovalController (a single chokepoint on the exact address that will sign) instead of `validateRequest`, which avoids per-method param parsing that could drift from the VM modules.
- The `!isInAppRequest(request)` guard is load-bearing. `WalletConnectService.getSession` throws while the WC client is initializing, so in-app Send/Swap during app startup must not touch it. In-app and injected-browser requests ride `CORE_MOBILE_TOPIC` (a per-launch uuid, not spoofable by a dApp) and keep their own account gating.
- Fails closed. A request with no live session (deleted or expired mid-flight, or WC uninitialized) has no grants to check against, so it is rejected. Its response is undeliverable anyway.
- `AVALANCHE_SEND/SIGN_TRANSACTION` are still not checked, and that is deliberate rather than an oversight: their `signingData` carries no `account` and they sign with the server-injected active account (`getContext` in `handleRequestViaVMModule`), which is the account the user can see.
- Legit flows cannot be falsely rejected. The SVM module schemas require the signer account in request params and never fall back to the active account, and dApps only ever learn granted accounts (`accountsChanged` grants before it announces).

**Correction to an earlier version of this description:** it claimed the `avax` namespace is only granted to Core-domain dApps. That is not true. A test dApp self-identifying as `https://jup.ag` was offered and granted the `avax` namespace with both P-Chain and X-Chain, reproduced twice during dev testing. `isCoreDomain()` trusts a dApp-supplied metadata string. That makes the `avalanche_signMessage` half of this PR load-bearing rather than defense in depth.

**Dependencies:** rebased onto main. The branch had been sitting on stale `@avalabs/*-module` 3.9.0 pins while main moved to 4.0.1 in #4014, and 4.0.1's avalanche module requires the `getAuthHeaders` wiring main added in #4012. No conflicts, and no changes to `package.json` or `yarn.lock` from this branch.

## Screenshots/Videos

Android, Pixel 8 Pro, internal build, on the merged 4.0.1 tree. Setup in every shot: one wallet with 7 accounts, a WalletConnect session granting **only Account 1**, and Core's active account left on **Account 2**.

The authorization change itself has deliberately nothing to show, because rejected requests never open a sheet. A recording of those paths is a still portfolio, so the evidence for them is the dApp-side error codes and timings in the Testing section. The one visible change is the display fix:

<table>
<tr>
<td width="50%"><b>Before</b><br>Network and message only. No account shown, so the user cannot check the signed address against the displayed one.</td>
<td width="50%"><b>After</b><br><code>Account: P-avax1jsp…f2mrt</code>, matching the granted <code>P-avax1jsphj9g0unpdjs354gxsdmkylducq36v3f2mrt</code>, while Core sat on Account 2.</td>
</tr>
<tr>
<td>01-BEFORE-avalanche_signMessage-no-account-row</td>
<td>02-AFTER-avalanche_signMessage-shows-account-P-avax1jsp</td>
</tr>
</table>

Same request (`avalanche_signMessage`, `accountIndex: 0`) on the same tree, before and after the fix.

<details>
<summary><b>Supporting captures</b> (7): display-fix regression, approval paths, the control, a reject path, and case R</summary>

<br>

**Display-fix regression: `solana_signMessage` still shows exactly one Account row**, so the change did not add a duplicate.

03-solana_signMessage-still-single-account-row

**Case A. `solana_signMessage` from granted Account 1.** Sheet shows `8rxgp3YA…M14U`. The returned signature verifies against Account 1 and fails against Account 2.

04-caseA-solana_signMessage-shows-granted-account-1

**Case D. `solana_signTransaction`, real Jupiter swap, granted Account 1.** `From 8rxgp3YA…M14U`. The "may fail" warning is expected: Account 1 holds almost no SOL, so Blockaid's simulation errors.

05-caseD-solana_signTransaction-jupiter-from-account-1

**Case I, the control. `accountIndex: 1` where Account 2 *is* granted.** Correctly reaches the approval screen rather than being gated. Rejecting in the UI returned `4001`, not `4100`, which shows the gate did not short-circuit it.

06-caseI-control-index1-granted-reaches-approval

**Case B. What a reject actually looks like.** The screen immediately after a rejected request: portfolio sitting still, no sheet ever opened. Pair with the dApp-side `4100` at 929ms.

07-caseB-reject-no-sheet-portfolio-unchanged

**Case R, the strongest regression check.** In-app Send with `From A8PsKy2J…ZTYT`, which is Account 2, an account the live WC session does **not** grant. Correctly not gated, then broadcast and finalized on mainnet as `gbvYRX6N…Tuz9TU`.

08-caseR-inapp-send-FROM-non-granted-account-2

**The session grant used throughout:** 1 of 7 accounts, Account 1 only.

09-wc-session-granting-only-account-1

</details>

iOS not covered in this round.

## Testing

Bitrise: iOS 9439, Android 9440.

Unit: 200 tests across 9 suites (`ApprovalController` and its validators, `quickSwapsBypass`, `isAccountApproved`, `isAvalancheSignMessageAuthorized`, `ApprovalScreen/utils`). Full `tsc -p .` clean.

Verified on device (Android, Pixel 8 Pro, internal build) against a real WalletConnect v2 dApp harness that lets the requesting account be chosen per request. Setup throughout: mainnet, one wallet with 7 accounts, a WC session granting **only Account 1**, and Core's active account left on **Account 2**, which is exactly this ticket's scope.

| # | Scenario | Expected | Result |
|---|---|---|---|
| A | `solana_signMessage`, granted Acct 1 | Sheet shows Acct 1, signs | PASS. Sheet showed `8rxgp3YA…M14U`; signature verifies against Acct 1 and fails against Acct 2 |
| B | `solana_signMessage`, non-granted Acct 2 | Reject, no sheet | PASS. 929ms, `4100` |
| C | `solana_signTransaction`, real Jupiter swap, Acct 2 | Reject | PASS. 868ms, `4100` |
| D | `solana_signTransaction`, real Jupiter swap, Acct 1 | Sheet shows Acct 1, signs | PASS. Sheet showed `From 8rxgp3YA…M14U`; signature verifies against Acct 1 and fails against Acct 2 |
| E | `avalanche_signMessage`, `accountIndex: 1` (Acct 2) | Reject | PASS. 376ms, `4100` |
| F | `avalanche_signMessage`, `accountIndex: 0` (Acct 1) | Sheet shows Acct 1, signs | PASS |
| H | `avalanche_signMessage`, `accountIndex: 0.5` | Reject | PASS. 318ms, `4100` |
| I | Control: `accountIndex: 1` where Acct 2 **is** granted | Reach approval | PASS. Reached the sheet; rejecting in the UI returned `4001`, not `4100` |
| R | In-app Send 0.001 SOL **from** non-granted Acct 2 | Not gated | PASS. Broadcast and finalized on mainnet |
| G | Session deleted mid-flight, then Approve | Fail closed | Inconclusive, see below |

Notes on the ones that carry the most weight:

- **A and D are cryptographic, not cosmetic.** The returned ed25519 signatures verify against Account 1's public key and fail against Account 2's, so the account shown really is the account that signed.
- **H is the sharpest test.** `0.5` passes the avalanche module's `z.number().nonnegative()` schema, which is not `.int()`, so `selectAccountByIndex` matches nothing and the signer falls back to `accounts[0]`, meaning Account 1. It was set up adversarially, with a session granting Account 2 and Core's active account also on Account 2, so an implementation that treated a fractional index as "no index" and checked the active account would have passed the gate and then signed with a different account. It rejected in 318ms.
- **R is the strongest regression check.** It signs with Account 2, an account the live session does not grant, and is correctly not gated, which proves the `!isInAppRequest(request)` guard is doing real work. Logs confirm the path: topic `1bfab1bf-…` (a uuid, so `CORE_MOBILE_TOPIC`) and `getSolanaSigner called { accountIndex: 1 }`. On-chain receipt `gbvYRX6NZMBwVTwNiZUCMMurGMTjeeD9HhMwLt7DSkWzpNEwzbFP1zUfFh8wZpHsgQHJ3bQnAjkKB1FF7Tuz9TU`: 1,000,000 lamports from `A8PsKy2J…ZTYT` to `8rxgp3YA…M14U`, `err: null`, finalized.
- **Timing is itself evidence.** Rejects land in 318ms to 929ms while approvals take 30s or more, since they wait on a human tap. That gap shows rejects short-circuit before any sheet is constructed.
- **G is inconclusive by nature.** With the session deleted the response is undeliverable, so a fail-closed rejection cannot be distinguished from having nothing to deliver. Covered by unit test `ApprovalController.test.ts:1414` instead.

Display fix regression checks, run with the fix applied: the gate still rejects `accountIndex: 1` (491ms, `4100`), and `solana_signMessage` still shows exactly one Account row.

Two limitations worth stating plainly:

- A mid-flight active-account switch is unreachable from Android's UI. The approval form sheet covers the account switcher and a scrim tap does nothing, so the approval-time re-check can only be exercised against a programmatic switch. Not a defect, it just bounds what a device test can reach.
- iOS was not tested in this round, and neither were the EVM or Bitcoin reject paths on device. Both are covered by unit tests and share the same `signingData.account` chokepoint as the Solana methods that were verified.

Acceptance: a dApp can only trigger signing with accounts that are part of its WalletConnect session grant. Requests for any other account are rejected with an explicit error and no approval prompt, including accounts belonging to the same wallet as the granted one.

1. Onboard a wallet with at least 2 accounts, connect to a WC dApp granting **only** Account 1.
2. Switch Core's active account to Account 2.
3. Sign from the dApp. The approval sheet must show Account 1, and signing must succeed as before.
4. Confirm the address on the sheet matches the address the dApp reports back as the signer.

Regression surface to sweep: in-app Send, Swap and Stake, injected-browser dApp signing, Core web over WC including BTC and AVAX flows, and Ledger approvals. All are unaffected paths.

## Checklist

- [ ] I have included screenshots / videos of android and ios (Android only; iOS not covered this round)

[CP-14604]: https://ava-labs.atlassian.net/browse/CP-14604?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core-mobile/app/store/rpc/utils/isAccountApproved/isAccountApproved.test.ts
+++ b/packages/core-mobile/app/store/rpc/utils/isAccountApproved/isAccountApproved.test.ts
@@ -1,4 +1,94 @@
-import { isAccountApproved } from './isAccountApproved'
+import { isAccountApproved, isAddressApproved } from './isAccountApproved'
+
+describe('isAddressApproved', () => {
+  const namespaces = {
+    eip155: {
+      accounts: [
+        'eip155:43113:0xC7E5ffBd7843EdB88cCB2ebaECAa07EC55c65318',
+        'eip155:43114:0xcA0E993876152ccA6053eeDFC753092c8cE712D0'
+      ],
+      methods: ['eth_sendTransaction'],
+      events: ['chainChanged', 'accountsChanged']
+    },
+    solana: {
+      accounts: [
+        'solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp:9gQmZ7fTTgv5hVScrr9QqT6SpBs7i4cKLDdj4tuae3sW'
+      ],
+      methods: ['solana_signTransaction'],
+      events: []
+    }
+  }
+
+  it('returns true when the address is granted in the namespace', () => {
+    expect(
+      isAddressApproved(
+        '0xcA0E993876152ccA6053eeDFC753092c8cE712D0',
+        'eip155:43114',
+        namespaces
+      )
+    ).toBe(true)
+  })
+
+  it('matches addresses granted under another chain of the same namespace', () => {
+    expect(
+      isAddressApproved(
+        '0xC7E5ffBd7843EdB88cCB2ebaECAa07EC55c65318',
+        'eip155:43114',
+        namespaces
+      )
+    ).toBe(true)
+  })
+
+  it('matches case-insensitively', () => {
+    expect(
+      isAddressApproved(
+        '0xca0e993876152cca6053eedfc753092c8ce712d0',
+        'eip155:43114',
+        namespaces
+      )
+    ).toBe(true)
+  })
+
+  it('returns false when the address is not granted', () => {
+    expect(
+      isAddressApproved(
+        '0x341b0073b66bfc19FCB54308861f604F5Eb8f51b',
+        'eip155:43114',
+        namespaces
+      )
+    ).toBe(false)
+  })
+
+  it('returns true for a granted Solana account', () => {
+    expect(
+      isAddressApproved(
+        '9gQmZ7fTTgv5hVScrr9QqT6SpBs7i4cKLDdj4tuae3sW',
+        'solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp',
+        namespaces
+      )
+    ).toBe(true)
+  })
+
+  it('returns false for a Solana account that was never granted', () => {
+    expect(
+      isAddressApproved(
+        '4Nd1mYQZ6X8jY6nWn6iVrDpcLzJq9z2NKV1S1nGiqNz1',
+        'solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp',
+        namespaces
+      )
+    ).toBe(false)
+  })
+
+  it('returns false when the session has no namespace for the chain', () => {
+    expect(
+      isAddressApproved(
+        'bc1qxy2kgdygjrsqtzq2n0yrf2493p83kkfjhx0wlh',
+        'bip122:000000000019d6689c085ae165831e93',
+        namespaces
+      )
+    ).toBe(false)
+  })
+})
 
 describe('isAccountApproved', () => {
   it('should return true if address is approved', () => {
--- a/packages/core-mobile/app/store/rpc/utils/isAccountApproved/isAccountApproved.test.ts
+++ b/packages/core-mobile/app/store/rpc/utils/isAccountApproved/isAccountApproved.test.ts
@@ -39,7 +39,7 @@ describe('isAddressApproved', () => {
     ).toBe(true)
   })
 
-  it('matches case-insensitively', () => {
+  it('matches EVM addresses case-insensitively (EIP-55 checksum casing)', () => {
     expect(
       isAddressApproved(
         '0xca0e993876152cca6053eedfc753092c8ce712d0',
@@ -73,6 +73,16 @@ describe('isAddressApproved', () => {
     expect(
       isAddressApproved(
         '4Nd1mYQZ6X8jY6nWn6iVrDpcLzJq9z2NKV1S1nGiqNz1',
+        'solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp',
+        namespaces
+      )
+    ).toBe(false)
+  })
+
+  it('returns false for a case variant of a granted Solana account (base58 is case-sensitive)', () => {
+    expect(
+      isAddressApproved(
+        '9gqmz7fttgv5hvscrr9qqt6spbs7i4cklddj4tuae3sw',
         'solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp',
         namespaces
       )
--- a/packages/core-mobile/app/store/rpc/utils/isAccountApproved/isAccountApproved.ts
+++ b/packages/core-mobile/app/store/rpc/utils/isAccountApproved/isAccountApproved.ts
@@ -4,12 +4,13 @@ import {
   getAddressForChainId
 } from 'store/rpc/handlers/wc_sessionRequest/utils'
 
-export const isAccountApproved = (
-  account: CoreAccountAddresses,
+// A granted account under ANY chain of the request's namespace authorizes the
+// address — sessions grant per-namespace account access, not per-chain.
+export const isAddressApproved = (
+  address: string,
   caip2ChainId: string,
   namespaces: SessionTypes.Namespaces
 ): boolean => {
-  const address = getAddressForChainId(caip2ChainId, account)
   const namespace = caip2ChainId.split(':')[0]
 
   if (!namespace || !namespaces[namespace]) {
@@ -17,9 +18,20 @@ export const isAccountApproved = (
   }
 
   return Boolean(
-    address &&
-      namespaces[namespace]?.accounts.some(
-        acc => acc.split(':')[2]?.toLowerCase() === address.toLowerCase()
-      )
+    namespaces[namespace]?.accounts.some(
+      acc => acc.split(':')[2]?.toLowerCase() === address.toLowerCase()
+    )
+  )
+}
+
+export const isAccountApproved = (
+  account: CoreAccountAddresses,
+  caip2ChainId: string,
+  namespaces: SessionTypes.Namespaces
+): boolean => {
+  const address = getAddressForChainId(caip2ChainId, account)
+
+  return Boolean(
+    address && isAddressApproved(address, caip2ChainId, namespaces)
   )
 }
--- a/packages/core-mobile/app/store/rpc/utils/isAccountApproved/isAccountApproved.ts
+++ b/packages/core-mobile/app/store/rpc/utils/isAccountApproved/isAccountApproved.ts
@@ -17,11 +17,15 @@ export const isAddressApproved = (
     return false
   }
 
-  return Boolean(
-    namespaces[namespace]?.accounts.some(
-      acc => acc.split(':')[2]?.toLowerCase() === address.toLowerCase()
-    )
-  )
+  // Only EVM hex addresses are case-insensitive (EIP-55 checksums vary the
+  // casing); other namespaces (e.g. base58 Solana) must match exactly.
+  const matches =
+    namespace === 'eip155'
+      ? (acc: string): boolean =>
+          acc.split(':')[2]?.toLowerCase() === address.toLowerCase()
+      : (acc: string): boolean => acc.split(':')[2] === address
+
+  return Boolean(namespaces[namespace]?.accounts.some(matches))
 }
 
 export const isAccountApproved = (
--- a/packages/core-mobile/app/vmModule/ApprovalController/ApprovalController.ts
+++ b/packages/core-mobile/app/vmModule/ApprovalController/ApprovalController.ts
@@ -551,10 +551,19 @@ class ApprovalController implements VmModuleApprovalController {
     // Avalanche signingData variants carry no `account` and are knowingly not
     // checked: the avax namespace is only ever granted to Core-domain dApps,
     // and those methods always sign with the active account the user sees.
+    // Fails closed: a request with no live session (deleted/expired mid-flight,
+    // or WC uninitialized) has no grants to check against, so it's rejected —
+    // its response is undeliverable anyway.
     if ('account' in signingData && !isInAppRequest(request)) {
-      const session = WalletConnectService.getSession(request.sessionId)
+      const session = (() => {
+        try {
+          return WalletConnectService.getSession(request.sessionId)
+        } catch {
+          return undefined
+        }
+      })()
       if (
-        session &&
+        !session ||
         !isAddressApproved(
           signingData.account,
           request.chainId,
--- a/packages/core-mobile/app/vmModule/ApprovalController/ApprovalController.ts
+++ b/packages/core-mobile/app/vmModule/ApprovalController/ApprovalController.ts
@@ -33,6 +33,9 @@ import WalletService from 'services/wallet/WalletService'
 import { Curve } from 'utils/publicKeys'
 import { ledgerParamsStore } from 'features/ledger/store'
 import { OnApproveParams } from 'services/walletconnectv2/walletConnectCache/types'
+import WalletConnectService from 'services/walletconnectv2/WalletConnectService'
+import { providerErrors } from '@metamask/rpc-errors'
+import { isAddressApproved } from 'store/rpc/utils/isAccountApproved/isAccountApproved'
 import { WalletType } from 'services/wallet/types'
 import { promptForAppReviewAfterSuccessfulTransaction } from 'features/appReview/utils/promptForAppReviewAfterSuccessfulTransaction'
 import { CONFETTI_DURATION_MS } from 'common/consts'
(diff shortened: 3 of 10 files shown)


B0Y3R-AVA (commented):
Adversarial review (Claude + Codex, cross-checked against `svm-module`/`bitcoin-module`/`evm-module` @ 3.9.0). The core fix looks sound and, importantly, **does** cover every method it claims: `signingData.account` is populated at the module level for all EVM/Solana/Bitcoin signing methods (verified in module source — `sign-and-send-transaction.ts` requires `account: z.string()`, `bitcoin-send-transaction.ts` sets `account: from`, `eth-sign.ts` sets `account: address`, etc.), so the `'account' in signingData` guard genuinely reaches them all. Batch EVM is `CORE_MOBILE_TOPIC`-only, Avalanche/HVM correctly excluded (they force the signer from the active account, not dApp params). Nice, tightly-scoped chokepoint.

One thing I'd change before merge (F1 inline), plus two low-severity follow-ups:

**F1 — fail-open on a missing session (inline).** The `session &&` short-circuit skips the check entirely when `getSession()` returns `undefined`. Suggest failing closed.

**F2 — injected in-app browser is fully excluded (informational).** `!isInAppRequest(request)` excludes `CORE_MOBILE_TOPIC`, which the injected in-app browser also rides. Its safety rests entirely on the injected provider being active-account-only (CP-14382/CP-14385). Out of scope here, but worth a one-line confirmation that a site can't supply a non-active `from`/`account` on that path.

**F3 — case-insensitive address match (inline, pre-existing).** Not exploitable, matches pre-migration behavior; noting for a follow-up.

Tests are genuine (only `getSession` is mocked; `isAddressApproved` runs for real). Minor nit: the three "accept" cases don't `await requestApproval`, relying on navigation happening synchronously in the Promise executor — they'd silently pass if a pre-nav `await` is ever introduced.

packages/core-mobile/app/vmModule/ApprovalController/ApprovalController.ts
```diff
@@ -538,6 +541,34 @@ class ApprovalController implements VmModuleApprovalController {
...
+    // address that would sign. In-app and injected-browser requests
+    // (CORE_MOBILE_TOPIC) are excluded: they keep their own account gating,
+    // and getSession would throw while the WC client is still initializing.
+    // Avalanche signingData variants carry no `account` and are knowingly not
+    // checked: the avax namespace is only ever granted to Core-domain dApps,
+    // and those methods always sign with the active account the user sees.
+    if ('account' in signingData && !isInAppRequest(request)) {
+      const session = WalletConnectService.getSession(request.sessionId)
+      if (
+        session &&
```
B0Y3R-AVA: **F1 (suggest fail-closed):** when `getSession()` returns `undefined`, the whole gate is skipped and the approval opens unchecked. `requestApproval` runs well after the WC `session_request` event and after async module work (gas/simulation/XP-address fetch), so a `session_delete`/expiry landing in that window makes this fresh live lookup return `undefined` — turning "only granted accounts may sign" into no check. Exploitation is narrow (precise race; the torn-down topic can't receive the response), but it's a fail-open in the exact gate this PR adds.

Since the only non-in-app path here is real WC (where a live session must exist), failing closed is safe:
```ts
if ('account' in signingData && !isInAppRequest(request)) {
  const session = WalletConnectService.getSession(request.sessionId)
  if (!session || !isAddressApproved(signingData.account, request.chainId, session.namespaces)) {
    return { error: providerErrors.unauthorized('Requested address is not authorized') }
  }
}
```

**F4 (defense-in-depth, non-blocking):** `getSession` is unguarded here; its `client` getter throws via `assertNotUndefined` when WC isn't initialized. Practically unreachable for a real `session_request` (implies an initialized client), but a `try/catch` would avoid the CP-14756-style unhandled-rejection class if a future caller reaches this path pre-init.

packages/core-mobile/app/store/rpc/utils/isAccountApproved/isAccountApproved.ts
```diff
@@ -4,22 +4,34 @@ import {
...
     return false
   }
 
   return Boolean(
-    address &&
-      namespaces[namespace]?.accounts.some(
-        acc => acc.split(':')[2]?.toLowerCase() === address.toLowerCase()
-      )
+    namespaces[namespace]?.accounts.some(
+      acc => acc.split(':')[2]?.toLowerCase() === address.toLowerCase()
```
B0Y3R-AVA: **F3 (low severity, pre-existing — follow-up):** `.toLowerCase()` on both sides is correct for EVM hex but semantically wrong for case-sensitive base58 (Solana / legacy Bitcoin). Not exploitable — `address` here is a real wallet-derived signer, so a case-fold collision between two distinct owned keys is cryptographically negligible — and this matches the pre-migration behavior. Worth a follow-up to make matching namespace-aware (exact match for non-`eip155`, case-insensitive only for hex).

bogdandobritoiu: Agreed it is not exploitable and matches pre-migration behavior, so keeping the case-insensitive match in this PR to stay a pure restoration. Namespace-aware matching (exact for non-eip155, case-insensitive for hex) noted as a follow-up.

packages/core-mobile/app/vmModule/ApprovalController/ApprovalController.ts
bogdandobritoiu: Fixed in 0efb698ce — the gate now fails closed: a missing session rejects with `unauthorized`, and `getSession` is wrapped so the uninitialized-client throw takes the same rejection path (F4). The former "skips the check when no session" test now asserts rejection, and there is a new case for the throwing client — both use a *granted* address so they prove the missing session alone causes the rejection.

On the test nit: `await requestApproval` on the accept cases would deadlock (the promise only settles via onApprove/onReject), so the accept assertions now flush pending tasks before asserting navigation instead — same effect, they fail loudly if navigation ever moves behind an await.

bogdandobritoiu:
Re F2 (injected in-app browser exclusion) — traced the injected path to confirm a site cannot supply a non-active/ungranted signer:

- The injected router runs its own per-request gate before anything reaches `ApprovalController`: `requestedSignerAddresses()` extracts the dApp-named signer per method (`eth_sendTransaction` → `params[0].from`, `personal_sign`/typed-data → the address param) and `dispatchSigningRequest` rejects with `unauthorized` (no prompt) unless **every** requested signer is in `getGrantedAddresses(origin)` — `app/hooks/browser/injectedProvider/router.ts:74-94, 318-342`. Granted-but-non-active is intentionally allowed (multi-account); ungranted never signs. Covered by `router.test.ts:1351-1497` including the exact `from`-substitution attack and the Permit/off-chain class for `personal_sign`/`eth_signTypedData_v4`.
- `eth_sendTransactionBatch` deliberately gates on the **active** account instead of `from` (the batch handler signs with the active index and ignores per-tx `from`, so gating on `from` would be bypassable) — `router.ts:65-69`, tests at `router.test.ts:1407, 1430`.
- Solana has no injected path at all: `approvalMethods.ts` only classifies `eth_`/`personal_`/`avalanche_` as signing methods, so `solana_*` falls through to the read-only dispatcher and dies with `methodNotFound`.
- `avalanche_*` methods are first-party-origin-only and always sign with the active account (`router.ts:919-929`).

So the CORE_MOBILE_TOPIC exclusion in this PR rests on a gate that exists and is tested, not just on convention.

F1/F4 are fixed in 0efb698ce (fail closed on missing session + guarded `getSession`); F3 stays as a follow-up per the inline thread.

packages/core-mobile/app/store/rpc/utils/isAccountApproved/isAccountApproved.ts
```diff
@@ -4,22 +4,34 @@ import {
...
+    )
+  )
```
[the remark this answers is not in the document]
bogdandobritoiu: Known limitation, deliberately kept out of this PR's scope — the case-insensitive comparison matches the pre-migration `isAddressApproved` behavior this PR restores. It's not exploitable in practice: both sides of the comparison originate from our own account addresses (the session namespaces are populated from the wallet's accounts at connect time, and `signingData.account` is resolved by the VM module from the same store), so casing is consistent; a case-variant base58 string would also be a different/invalid pubkey that fails signing downstream. Agreed it should still be namespace-aware (case-insensitive only for `eip155`, exact match otherwise) — tracking that as a follow-up rather than expanding this PR.

packages/core-mobile/app/store/rpc/utils/isAccountApproved/isAccountApproved.test.ts
```diff
@@ -1,4 +1,94 @@
...
+
+  it('returns true for a granted Solana account', () => {
+    expect(
+      isAddressApproved(
+        '9gQmZ7fTTgv5hVScrr9QqT6SpBs7i4cKLDdj4tuae3sW',
+        'solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp',
+        namespaces
+      )
+    ).toBe(true)
+  })
```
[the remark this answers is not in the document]
bogdandobritoiu: A case-sensitivity regression test would fail against the current (deliberate) behavior — the comparison intentionally mirrors the pre-migration implementation for this PR. The test will land together with the namespace-aware comparison in the follow-up (see the thread on `isAccountApproved.ts`).

packages/core-mobile/app/store/rpc/utils/isAccountApproved/isAccountApproved.ts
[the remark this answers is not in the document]
bogdandobritoiu: Update: fixed in this PR after all (056b381cb) — the comparison is now namespace-aware: case-insensitive only for `eip155` (EIP-55 checksum casing), exact match for everything else including Solana base58.

packages/core-mobile/app/store/rpc/utils/isAccountApproved/isAccountApproved.test.ts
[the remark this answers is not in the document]
bogdandobritoiu: Update: regression test added in 056b381cb — a case variant of the granted Solana account now returns false.

packages/core-mobile/app/store/rpc/utils/isAccountApproved/isAccountApproved.ts
bogdandobritoiu: Update: fixed in this PR after all (056b381cb) after Copilot re-flagged the same thing — `isAddressApproved` is now namespace-aware (case-insensitive only for `eip155`, exact match otherwise) with a Solana case-variant regression test.

bohdan-struk-avalabs (approved):

PR #3995, opened 2026-07-20
CP-14671: present perps enable-trading checklist as a formsheet modal route

bogdandobritoiu:
## Description

**Ticket: [CP-14671](https://ava-labs.atlassian.net/browse/CP-14671)**

The perps "Set up trading account" checklist (agent approval, builder-fee approval, unified account) shipped in #3984 as an inline half-height RN `Modal` rendered inside each screen. The design ([Figma](https://www.figma.com/design/aj9mmgDMaaxZXkuIRKLhIn/Core-Mobile-2026?node-id=22848-36144)) specifies the standard stacked K2 formsheet instead. This PR converts it to a proper modal route:

- **New `/perpetualsEnableTrading` route** registered with `secondaryModalScreensOptions` (stacked-sheet detents), since it always presents on top of the place-order/close/manage modals.
- **New `PerpetualsEnableTradingScreen`** built on `ScrollScreen isModal` with title/subtitle per design; ports the three-step checklist logic unchanged. Dismissal is the sheet's native swipe/grabber (the design drops the old "Not now" button) plus a once-guarded `router.back()` when setup completes or is already satisfied.
- **`usePerpsEnableTradingGate`** is now a plain `.ts` hook: `requireTradingEnabled()` navigates to the route instead of toggling a locally rendered modal; the `enableTradingModal` render-prop is removed from PlaceOrder/Close/Manage screens and `PerpsEnableTradingModal` is deleted.
- **k2-alpine `Checklist` polish** to match design: plain check icon on completed rows, `ActivityIndicator` replaces the action button while a step is in flight, `subtitle2` description.
- **Proactive entry point on the details screen** (per design; core-web similarly gates its trade form until setup is done): the footer always shows Short / Long once funded, but for a first-time user the slide opens the enable-trading sheet instead of place-order, carrying the selected side. When the last setup step completes, the sheet `router.replace`s into place-order with the originally selected Short/Long (so back returns to the details screen, not the sheet). While setup status is still resolving the slide goes straight to place-order so a set-up user is never detoured on cold load (`usePerpsEnableTradingGate` now exposes `isTradingStatusLoading`; `PerpsProvider` now surfaces `isLoadingAgent`); the submit-time gate remains as the backstop for deep links and cold-load edge cases.

No new dependencies; no breaking changes.

## Screenshots/Videos

https://github.com/user-attachments/assets/b84e58e9-7bf3-4dbc-ac75-55c5fc83e30e

## Testing

**Dev Testing**

iOS - 9382
Android - 9383

Happy path:

1. Use a wallet with deposited HL balance that has not completed perps setup and open a market's details screen — the footer shows the normal Short / Long slider.
2. Slide Short or Long — instead of place-order, the "Set up trading account" formsheet presents stacked on top, matching the Figma.
3. Complete the three steps in order (each button shows a spinner while in flight, then a green check) — after the last step the sheet closes and place-order opens with the side you originally slid (verify both Short and Long are carried through).
4. Back from place-order returns to the details screen (not the setup sheet); subsequent slides go straight to place-order.
5. Backstop path: reach the place-order screen without setup (e.g. deep link) and slide to confirm — the sheet presents at submit time (no side carried), dismisses in place after setup, and the order is not sent until re-submitted.

Edge cases:

- Swipe the sheet down mid-setup — the order is not submitted; re-submitting reopens the sheet with completed steps still checked.
- Reject a signature request — a snackbar shows and the step stays incomplete.
- Open the gate on a wallet that already finished setup — the sheet dismisses itself immediately.

**QA Testing**

- Verify all perps order actions (place order, market/limit close, manage TP/SL) still gate on trading setup and present the new formsheet, on both iOS and Android.
- Verify checklist steps enable strictly in order (each button disabled until the previous step is done) and completed steps show a check.
- Expected: behavior is identical to the previous inline sheet; only the presentation changed to a full-height stacked formsheet.

## Checklist

[CP-14671]: https://ava-labs.atlassian.net/browse/CP-14671?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core-mobile/app/new/features/trade/perpetuals/hooks/usePerpsEnableTradingGate.ts
+++ b/packages/core-mobile/app/new/features/trade/perpetuals/hooks/usePerpsEnableTradingGate.ts
@@ -1,5 +1,5 @@
-import React, { useCallback, useMemo, useState } from 'react'
-import { PerpsEnableTradingModal } from '../components/PerpsEnableTradingModal'
+import { useRouter } from 'expo-router'
+import { useCallback, useMemo } from 'react'
 import { usePerps } from '../contexts/PerpsProvider'
 import { usePerpsBuilderFee } from './usePerpsBuilderFee'
 import { usePerpsUnifiedAccount } from './usePerpsUnifiedAccount'
@@ -13,29 +13,28 @@ export type PerpsEnableTradingGate = {
   readonly isTradingEnabled: boolean
   /**
    * Call at the top of a submit handler. Returns `true` when one-time trading
-   * setup is complete; otherwise opens the enable-trading sheet and returns
-   * `false` so the caller can bail out (the user re-submits once set up).
+   * setup is complete; otherwise presents the enable-trading form sheet and
+   * returns `false` so the caller can bail out (the user re-submits once set
+   * up).
    */
   readonly requireTradingEnabled: () => boolean
-  /** Render this in the screen tree so the enable-trading sheet can appear. */
-  readonly enableTradingModal: React.ReactNode
 }
 
 /**
  * Shared enable-trading gate for perps actions (place order, close, manage).
  * Mirrors core-web's flow: any action that signs an L1 order first ensures the
  * account is set up (agent + builder fee + unified account, matching the
- * three-step checklist) and, if not, surfaces {@link PerpsEnableTradingModal}
- * instead of submitting.
+ * three-step checklist) and, if not, presents the enable-trading form sheet
+ * (`/perpetualsEnableTrading`) instead of submitting.
  */
 export const usePerpsEnableTradingGate = (): PerpsEnableTradingGate => {
+  const router = useRouter()
   const { hasAgent } = usePerps()
   const {
     isApproved: isBuilderFeeApproved,
     feeTenthsBps: builderFeeTenthsBps
   } = usePerpsBuilderFee()
   const { isUnifiedAccount } = usePerpsUnifiedAccount()
-  const [visible, setVisible] = useState(false)
 
   const isTradingEnabled = useMemo(
     () =>
@@ -47,15 +46,11 @@ export const usePerpsEnableTradingGate = (): PerpsEnableTradingGate => {
 
   const requireTradingEnabled = useCallback((): boolean => {
     if (!isTradingEnabled) {
-      setVisible(true)
+      router.navigate('/perpetualsEnableTrading')
       return false
     }
     return true
-  }, [isTradingEnabled])
+  }, [isTradingEnabled, router])
 
-  const enableTradingModal = (
-    <PerpsEnableTradingModal open={visible} onClose={() => setVisible(false)} />
-  )
-
-  return { isTradingEnabled, requireTradingEnabled, enableTradingModal }
+  return { isTradingEnabled, requireTradingEnabled }
 }
--- a/packages/core-mobile/app/new/features/trade/perpetuals/hooks/usePerpsEnableTradingGate.ts
+++ b/packages/core-mobile/app/new/features/trade/perpetuals/hooks/usePerpsEnableTradingGate.ts
@@ -1,4 +1,4 @@
-import { useRouter } from 'expo-router'
+import { useNavigation, useRouter } from 'expo-router'
 import { useCallback, useMemo } from 'react'
 import { usePerps } from '../contexts/PerpsProvider'
 import { usePerpsBuilderFee } from './usePerpsBuilderFee'
@@ -36,6 +36,7 @@ export type PerpsEnableTradingGate = {
  */
 export const usePerpsEnableTradingGate = (): PerpsEnableTradingGate => {
   const router = useRouter()
+  const navigation = useNavigation()
   const { hasAgent, isLoadingAgent } = usePerps()
   const {
     isApproved: isBuilderFeeApproved,
@@ -58,11 +59,16 @@ export const usePerpsEnableTradingGate = (): PerpsEnableTradingGate => {
 
   const requireTradingEnabled = useCallback((): boolean => {
     if (!isTradingEnabled) {
-      router.navigate('/perpetualsEnableTrading')
+      // Callers may await before gating (e.g. place-order's geo re-check). If
+      // the user dismissed the calling screen during that await, don't present
+      // the setup sheet over wherever they are now — just abort the submit.
+      if (navigation.isFocused()) {
+        router.navigate('/perpetualsEnableTrading')
+      }
       return false
     }
     return true
-  }, [isTradingEnabled, router])
+  }, [isTradingEnabled, navigation, router])
 
   return { isTradingEnabled, isTradingStatusLoading, requireTradingEnabled }
 }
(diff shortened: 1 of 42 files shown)


packages/core-mobile/app/new/features/trade/perpetuals/screens/PerpetualsEnableTradingScreen.tsx
```diff
@@ -47,6 +33,17 @@ export const PerpsEnableTradingModal = ({
...
 
+  // The last step's handler dismisses eagerly while the effect below also
+  // reacts to the queries flipping done — guard so we only pop this sheet once.
+  const dismissedRef = useRef(false)
+  const dismiss = useCallback(() => {
+    if (dismissedRef.current) {
+      return
+    }
+    dismissedRef.current = true
+    router.canGoBack() && router.back()
```
B0Y3R-AVA: `dismissedRef` only gets set by our own `dismiss()` calls. Nothing flips it when the user swipe-dismisses the sheet natively.

So: tap "Enable unified account", swipe the sheet down while the tx is in flight. When the promise resolves, `dismiss()` still runs and `router.back()` fires against a stack where this sheet is already gone, popping the place-order/close/manage screen underneath it. The old code just did `setVisible(false)`, harmless after unmount, so this is new with the route conversion.

Android makes it worse: a surplus `back()` can close the app (there's a comment about exactly that in `(signedIn)/_layout.tsx`), and `navigationGuard` patches `push`/`navigate`/`replace` but not `back()`.

I'd set `dismissedRef` from a `beforeRemove` listener, or check `navigation.isFocused()` before calling `back()`. Dismissing by route identity instead of raw `back()` would also do it.

packages/core-mobile/app/new/features/trade/perpetuals/screens/PerpetualsPlaceOrderScreen.tsx
```diff
@@ -101,8 +101,7 @@ export const PerpetualsPlaceOrderScreen = (): JSX.Element => {
 
   const { isGeoBlocked, recheckGeoBlock } = usePerpsAvailability()
   const { submitOrder } = usePerpsOrderSubmit()
-  const { requireTradingEnabled, enableTradingModal } =
-    usePerpsEnableTradingGate()
+  const { requireTradingEnabled } = usePerpsEnableTradingGate()
```
B0Y3R-AVA: Same shape of bug on the caller side (the code is unchanged in this PR but the gate's behavior changed under it). In `handleConfirm`, the geo re-check awaits *before* the gate, and the gate now navigates instead of setting local state. If the user swipe-dismisses the place-order sheet while the re-check is in flight (it can look hung), the continuation still runs and `router.navigate('/perpetualsEnableTrading')` opens the setup sheet over whatever screen they're on now, after they already bailed. Previously `setVisible(true)` on the unmounted screen was a no-op.

A mounted/focused check after the await covers it.

packages/k2-alpine/src/components/Checklist/Checklist.tsx
```diff
@@ -91,26 +91,28 @@ const ChecklistRow = ({
...
         completedIcon ?? (
-          <Icons.Action.CheckCircle
+          <Icons.Navigation.Check
             width={28}
             height={28}
             color={theme.colors.$textSuccess}
           />
         )
+      ) : loading ? (
+        <ActivityIndicator size="small" />
```
B0Y3R-AVA, line 107: Doc nits while you're in here: the `loading` docstring on `ChecklistItem` still says the spinner shows "in place of the action label" (it now replaces the whole button), and the `done` docstring still names `Icons.Action.CheckCircle` as the default.

B0Y3R-AVA (commented):
Ran an adversarial pass on this (Claude + Codex cross-check). The conversion itself is clean: the rename left no dead refs, all four gate call sites still treat `requireTradingEnabled` as a sync boolean, and nothing else consumes the k2 `Checklist` so that change is contained.

Found two real bugs though, both the same shape: the modal used to be local state, now it's global navigation, so async work that finishes after the user has already left can still poke the nav stack. Details inline.

One more thing worth eyeballing when you do the device recordings: this is the first secondary sheet we present on top of another secondary sheet (`perpetualsClose`). Every other `secondaryModalScreensOptions` screen opens from a primary modal. Probably fine, but nobody has seen the stacked detents/gestures on a real device yet.

packages/core-mobile/app/new/features/trade/perpetuals/screens/PerpetualsEnableTradingScreen.tsx
bogdandobritoiu: Fixed in 89547fdbe. A `beforeRemove` listener now marks the sheet dismissed on any removal (same pattern as `PerpetualsOnboardingScreen`), so a continuation resolving after a native swipe-dismiss no-ops instead of firing `back()` — or, since the slide-to-order change, `replace()`ing the underlying screen into place-order, which made this worse than the original report.

packages/core-mobile/app/new/features/trade/perpetuals/screens/PerpetualsPlaceOrderScreen.tsx
bogdandobritoiu: Fixed in 89547fdbe, in the gate hook rather than this call site: `requireTradingEnabled()` now checks `navigation.isFocused()` before presenting the sheet (it still returns `false`, so the submit aborts either way). Hook-level covers all four call sites — place-order is the only one with an await before the gate today, but close/manage won't regress if one grows an async precheck.

packages/k2-alpine/src/components/Checklist/Checklist.tsx
bogdandobritoiu: Fixed in 89547fdbe — `done` now names `Icons.Navigation.Check` and `loading` says the spinner replaces the button.

packages/core-mobile/app/new/features/trade/perpetuals/hooks/usePerpsEnableTradingGate.ts
```diff
@@ -0,0 +1,74 @@
...
+      // Callers may await before gating (e.g. place-order's geo re-check). If
+      // the user dismissed the calling screen during that await, don't present
+      // the setup sheet over wherever they are now — just abort the submit.
+      if (navigation.isFocused()) {
+        router.navigate('/perpetualsEnableTrading')
+      }
+      return false
+    }
+    return true
+  }, [isTradingEnabled, navigation, router])
```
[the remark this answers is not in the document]
bogdandobritoiu: Considered and not taking this one. Short-circuiting on `isTradingStatusLoading` turns a cold-load submit into a dead slide — `requireTradingEnabled()` returns false, the button resets, and the user gets zero feedback; if a setup query hangs (e.g. offline), they can never reach the sheet or see why. The current behavior self-corrects in the race this describes: for an already-configured account the sheet's already-satisfied effect auto-dismisses it the moment the queries resolve, and for an unconfigured account the sheet was the right destination anyway. The proactive path (details-screen slide) does consult `isTradingStatusLoading` and skips the detour — the doc comment is about that consumer. Added unit coverage for the gate's semantics in e2c36c9bd.

PR #3999, opened 2026-07-21
CP-14849: i18n runtime foundation (i18next v26 + custom lazy-load RN backend + language slice)

ruijia1in:
## Description

**Ticket: [CP-14849](https://ava-labs.atlassian.net/browse/CP-14849)** — also seeds font-stack placeholders for [CP-14859](https://ava-labs.atlassian.net/browse/CP-14859).

Stands up the **i18next runtime foundation + persisted language preference** for Core Mobile. Self-contained — it also delivers the Redux `settings/language` slice + schema migration originally scoped as CP-14850 (that ticket is superseded).

- **Stack:** `i18next@26.3.6` + `react-i18next@17.0.10`; single flat `translation` namespace, natural-English-string keys, `keySeparator/nsSeparator: false`, `en-US` full-locale folders — matches Core Web's convention so the shared Crowdin TM applies. (Empirically verified on v26/v17: tsc clean, 29 unit tests pass.)
- **Loading:** a custom React Native backend loads bundled per-locale JSON via a static `require`-thunk map; Metro `inlineRequires` keeps non-active locales out of memory. This replaces `i18next-http-backend`, which can't work in RN (no origin to fetch from; remote fetch would be OTA, which is disallowed).
- **No cold-start flash:** i18next is seeded **synchronously** from a non-encrypted MMKV cache at the top of bootstrap, so the first render is already in the persisted language and `i18n.language` is correct before the first analytics event (Redux is encrypted + async-rehydrated, so unreadable that early).
- **State:** Redux `settings/language` slice + selector + write-through listener; schema migration `31` (store VERSION 30→31); app-level, shared across accounts, default `en-US`, **no device-locale detection**.
- **Wiring:** `I18nextProvider` app-wide (outside the async Redux provider); `initI18n()` runs before `AppRegistry.registerComponent`.
- **Catalogs:** all 11 locales bundled. `en-US` is a small scaffold; the CJK + Devanagari locales carry placeholder strings for an OS **font-stack** rendering check (CP-14859). Real strings arrive later via the extraction sweep (CP-14852) + Crowdin sync (CP-14853).

**Also includes — CJK/Devanagari lineHeight fix (k2-alpine `Text`):** several k2-alpine text variants set `lineHeight` at/near `fontSize`, which clips taller non-Latin glyphs (CJK ideographs, Devanagari matras) on iOS when they render via the OS fallback font. The `Text` primitive now injects a per-script relaxed `lineHeight` — **CJK ×1.2, Devanagari ×1.35** — but only when it exceeds the variant's own `lineHeight`; Latin text and any caller-set `lineHeight` are untouched. Device-validated: both scripts clear clipping, Latin unchanged. (Landed here on the foundation branch so it's available to everything stacked on top.)

**Dependencies introduced:** `i18next@26.3.6`, `react-i18next@17.0.10`. Note: this puts mobile *ahead* of Core Web (24.2.2 / 15.4.0); convention + shared-TM parity is version-independent, so TM reuse is unaffected.

**Breaking:** none. The i18n runtime has no `t()` call sites yet; the `Text` lineHeight change only affects rendering of CJK/Devanagari content and is a no-op for Latin text.

## Screenshots/Videos

N/A — this PR contains **no user-facing UI**. The runtime foundation has no `t()` call sites yet (those land with the extraction sweep, CP-14852), so there is no visible surface to screenshot. The CJK/Devanagari catalogs are data-only; a visible font-rendering surface for Design/Product is deferred (CP-14859).

## Testing

**Dev Testing**

- Launch the build on iOS and Android — the app should behave **exactly as before**. The i18n runtime initializes at bootstrap but nothing is wired to it yet, so there should be **no visible change and no regression** anywhere.
- Cold start: the app opens normally with no added flash or delay (the i18n seed is a single synchronous MMKV read).
- **Latin regression check (lineHeight fix):** across core flows, confirm all-Latin text is **visually unchanged** — same line spacing, no shifts. Any existing CJK/Devanagari text (e.g. a contact/token name) should render without top/bottom clipping.
- Build triggered on Bitrise — see Build Numbers below.

**QA Testing**

- Regression pass only: confirm no functional or visual regressions across core flows (portfolio, send/receive, account settings, onboarding). The foundation is intended to be invisible in this PR.
- Language-selector UI and per-locale rendering/overflow QA are out of scope here and tracked under later epic tickets.

## Build Numbers

- iOS: 9386
- Android: 9387
- Pipeline: https://app.bitrise.io/app/7d7ca5af7066e290/pipelines/595a0dc8-c14a-44ee-9cf6-4f6357d956b9

## Checklist

- [ ] I have included screenshots / videos of android and ios — N/A, no UI in this PR

[CP-14849]: https://ava-labs.atlassian.net/browse/CP-14849?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CP-14859]: https://ava-labs.atlassian.net/browse/CP-14859?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/core-mobile/index.js
```diff
@@ -85,6 +86,7 @@ FCMService.listenForMessagesBackground()
 // PRESS events for notifications displayed by notifee (data-only Android push).
 NotificationsService.registerBackgroundNotificationHandler()
 
+initI18n()
 AppRegistry.registerComponent(expo.name, () => AppEntryPoint)
```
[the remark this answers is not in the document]
ruijia1in: Fixed in 7e656cb4d — the boot call is now `initI18n().catch(err => Logger.error('[i18n] init failed', err))`, so a rejected init logs instead of becoming an unhandled rejection.

packages/core-mobile/app/store/settings/language/slice.ts
```diff
@@ -0,0 +1,32 @@
...
+
+const reducerName = 'language'
+
+export const languageSlice = createSlice({
+  name: reducerName,
+  initialState,
+  reducers: {
+    setSelectedLanguage: (state, action: PayloadAction<string>) => {
+      state.selected = action.payload
+    }
```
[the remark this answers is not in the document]
ruijia1in: Fixed in 7e656cb4d — the reducer now clamps at the boundary: `state.selected = SUPPORTED_LANGUAGE_CODES.includes(action.payload) ? action.payload : DEFAULT_LANGUAGE`, so state can never hold an unsupported code. Added a unit test for the clamp.

packages/core-mobile/app/store/settings/language/listeners.ts
```diff
@@ -0,0 +1,30 @@
...
+  const code = action.payload
+  // write-through to the synchronous bootstrap cache (Redux stays the
+  // official persisted preference; this is what the cold-start seed reads)
+  commonStorage.set(LANGUAGE_MMKV_KEY, code)
+  // drive the runtime re-render; catch so a failed locale load (should be
+  // impossible for bundled JSON) is logged rather than an unhandled rejection
+  i18n
+    .changeLanguage(code)
+    .catch(err => Logger.error(`[i18n] changeLanguage(${code}) failed`, err))
+}
```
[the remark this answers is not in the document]
ruijia1in: Fixed in 7e656cb4d — the listener now derives the effective code from state via `selectSelectedLanguage(listenerApi.getState())` instead of the raw payload, so only a validated/clamped code is written to MMKV and passed to `changeLanguage`. Added a clamp test.

packages/core-mobile/app/i18n/index.ts
```diff
@@ -0,0 +1,47 @@
...
+    .use(RequireBackend)
+    .use(initReactI18next)
+    .init({
+      lng,
+      fallbackLng: DEFAULT_LANGUAGE,
+      supportedLngs: SUPPORTED_LANGUAGE_CODES,
+      load: 'currentOnly',
+      keySeparator: false,
+      nsSeparator: false,
+      partialBundledLanguages: true,
```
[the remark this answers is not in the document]
ruijia1in: Good catch — applied, with a correction: `initImmediate` was **removed in i18next v26** (renamed to `initAsync`), so the equivalent is `initAsync: false`, now set in `init()` (8c1306727e9defde5a991678e4dd7553d9e1dfdc). This forces synchronous init; combined with the inline `resources` for the active locale and our synchronous `RequireBackend.read` (a bundled `require`), i18next is fully ready — `i18n.language` set, active-locale `t()` working — before `init()` returns, i.e. before `registerComponent`/first render. No async gap.

packages/core-mobile/index.js
```diff
@@ -85,6 +86,7 @@ FCMService.listenForMessagesBackground()
 // PRESS events for notifications displayed by notifee (data-only Android push).
 NotificationsService.registerBackgroundNotificationHandler()
 
+initI18n().catch(err => Logger.error('[i18n] init failed', err))
```
bogdandobritoiu: **[High] App continues on an uninitialized i18n, and this log may never reach Sentry.** `registerComponent` runs unconditionally on the next line. In the normal path `initAsync: false` means init settles synchronously before this `.catch` schedules, so it's fine — but if init *rejects*, the tree mounts against an uninitialized singleton and (with `useSuspense: false`) renders raw keys everywhere. Compounding it: `Logger.error` only forwards to Sentry after `store/posthog/listeners.ts` arms `shouldLogErrorToSentry` later in bootstrap, so a cold-start i18n failure can be logged *nowhere* in release. Suggest capturing with an explicit `errorId` / `SentryService.captureException`, and validating the seed catalog (see the `resources` note) so "init rejected" and "init succeeded but seeded empty" are both observable.

packages/core-mobile/app/i18n/index.ts
```diff
@@ -0,0 +1,53 @@
...
+      interpolation: { escapeValue: false },
+      returnNull: false,
+      // No Suspense boundary wraps the tree; keep useTranslation synchronous
+      // so a consumer mounting before a namespace resolves never throws.
+      react: { useSuspense: false },
+      resources: { [lng]: { translation: LOCALES[lng]?.() ?? {} } }
```
bogdandobritoiu: **[High] `?? {}` masks a broken active-locale catalog as an all-keys UI.** Two failure modes aren't handled: (1) if the `require` thunk *throws* (malformed bundled JSON, a bad OTA/CodePush update), a throw isn't `undefined`, so `?? {}` doesn't catch it — it escapes as an opaque bootstrap crash; (2) if the thunk returns a truthy-but-empty/partial object, `?? {}` silently seeds an empty catalog and, with `returnNull: false` and no missingKey handler, the whole app renders raw i18n keys for the user's active language — no log, no Sentry, no user signal. This is the highest-risk line in the PR and the one non-English users are least able to report. Suggest loading the active catalog explicitly, wrapping the thunk call in try/catch, and logging via an `errorId` when it's missing or empty:
```ts
const activeCatalog = LOCALES[lng]?.()
if (!activeCatalog || Object.keys(activeCatalog).length === 0) {
  Logger.error(`[i18n] active locale catalog empty/missing for ${lng}`, undefined, { errorId: ErrorId.I18N_CATALOG_MISSING, locale: lng })
}
```

bogdandobritoiu (commented):
Review from a multi-pass pass over the i18n foundation (correctness, error handling, types, tests, comments). Nice work overall — disciplined Redux layering, behavior-level tests, and unusually good "why" comments. Left inline notes below.

The through-line for the high-severity items: the design deliberately *survives* i18n failures by falling back to English / an empty catalog, which is the right non-fatal posture — but every fallback is currently silent, so a genuine break (corrupt/empty catalog, failed init, failed switch) shows up only as a UI full of raw dotted keys, for exactly the non-English users this feature serves, with no log/Sentry signal. The fix in each case is to keep the fallback but make it observable (log via an `errorId`) and, in the interactive switch case, give the user feedback.

**High (worth addressing before merge):** empty-catalog masking, Redux↔i18n cold-start divergence, swallowed `changeLanguage`, app continuing on uninitialized i18n.
**Medium:** `initAsync` version comment is wrong; a `LanguageCode` union would remove the scattered clamps and catch code↔locale drift at compile time; a few error-path test gaps.

Low-severity (non-blocking, not inlined): the clamp-to-`en-US` sites are all silent — a `Logger.warn` on an *unexpected* (non-empty, unsupported) drop would make a bug distinguishable from the expected "removed language" path; consider softening the "no async tick" wording and the hardcoded "11" in locales.ts.

packages/core-mobile/app/i18n/index.ts
```diff
@@ -0,0 +1,53 @@
...
+  return i18n
+    .use(RequireBackend)
+    .use(initReactI18next)
+    .init({
+      lng,
+      // force synchronous init so i18n.language + the active-locale resources
+      // are ready by the time init() returns (before registerComponent / first
+      // render) — the inline `resources` and the synchronous RequireBackend
+      // both resolve without a tick, so no async gap / no cold-start flash.
+      // (v26 renamed the old `initImmediate` option to `initAsync`.)
```
bogdandobritoiu: **[Medium] Version attribution is wrong.** `initImmediate` was renamed to `initAsync` in i18next **v24.0.0**; v26 only *removed* the deprecated `initImmediate` alias. `initAsync: false` itself is correct here and does exactly what the surrounding comment says. Suggest dropping the version parenthetical (the option name is self-documenting on the pinned v26) or correcting it to v24 so a future reader searching the changelog for a "v26 rename" isn't misled.

packages/core-mobile/app/store/settings/language/listeners.ts
```diff
@@ -0,0 +1,32 @@
...
+  // raw action payload, so an unsupported code can never be persisted or loaded
+  const code = selectSelectedLanguage(listenerApi.getState())
+  // write-through to the synchronous bootstrap cache (Redux stays the
+  // official persisted preference; this is what the cold-start seed reads)
+  commonStorage.set(LANGUAGE_MMKV_KEY, code)
+  // drive the runtime re-render; catch so a failed locale load (should be
+  // impossible for bundled JSON) is logged rather than an unhandled rejection
+  i18n
+    .changeLanguage(code)
+    .catch(err => Logger.error(`[i18n] changeLanguage(${code}) failed`, err))
```
bogdandobritoiu: **[High] Swallowing this rejection strands the UI in a half-switched state.** By the time `changeLanguage` runs, `commonStorage.set` (line 17) and the reducer have *already* committed the new code. So if it rejects, MMKV + Redux + the cold-start seed all say e.g. `es-ES`, the settings checkmark moves, but the live instance stays English — inconsistent across three stores with no user feedback. The `.catch` is a correct guard against an unhandled rejection, but logging alone isn't enough given the side effects are already persisted. Suggest either running `changeLanguage` *first* and writing MMKV / trusting reducer state only on success, or surfacing a toast on failure — and attaching an `errorId` so the well-formed error from `backend.ts` actually lands in Sentry.

Also: this rejection branch is untested — `listeners.test.ts` mocks `changeLanguage` as always-resolve, so a refactor dropping the `.catch` would pass every test. Add a reject case asserting `Logger.error` is called (needs a `utils/Logger` mock).

packages/core-mobile/app/store/settings/language/listeners.ts
```diff
@@ -0,0 +1,32 @@
...
+}
+
+export const addLanguageListeners = (
+  startListening: AppStartListening
+): void => {
+  startListening({
```
bogdandobritoiu, line 40: **[High] Language never reconciles from Redux → i18n after rehydration.** This listener subscribes only to `setSelectedLanguage`. Its sibling `settings/appearance/listeners.ts` also matches `onRehydrationComplete` so the runtime side-effect is re-applied from persisted Redux on every cold start — language has no equivalent. Since `initI18n` seeds purely from the non-encrypted MMKV key while encrypted Redux holds the "official" preference (per the comment above), if MMKV is cleared independently of Redux (storage reset, reinstall-restore paths), the next boot runs i18n in `en-US` while Redux still holds `es-ES`, and nothing re-syncs — the user is stuck in English with no self-heal except re-toggling. Suggest adding an `onRehydrationComplete` matcher that calls `changeLanguage(selectSelectedLanguage(state))`, mirroring appearance; this also makes Redux the authoritative source it's documented to be.

packages/core-mobile/app/store/settings/language/types.ts
```diff
@@ -0,0 +1,34 @@
+export type Language = {
+  code: string
+  name: string // English name
+  nativeName: string
+}
+
+export const SUPPORTED_LANGUAGES: Language[] = [
```
bogdandobritoiu: **[Medium] A `LanguageCode` union would delete the scattered runtime clamps and catch drift at compile time.** Codes are bare `string` everywhere, so four `.includes()` guards (reducer write, selector read, bootstrap seed, backend) exist to compensate. Two invariants are currently unguarded and only fail at runtime: `DEFAULT_LANGUAGE` being a supported code, and — highest value — `SUPPORTED_LANGUAGES` ↔ `LOCALES` staying in sync (add a locale to one list, forget the other → clean compile, runtime untranslated screen via the `?? {}` fallback). Deriving the union from a single source makes both compile errors, and since a string-literal union is a subtype of `string` every i18next call keeps working:
```ts
export const SUPPORTED_LANGUAGES = [ /* ... */ ] as const satisfies readonly Language[]
export type LanguageCode = (typeof SUPPORTED_LANGUAGES)[number]['code']
export const DEFAULT_LANGUAGE = 'en-US' as const satisfies LanguageCode
export type LanguageState = { selected: LanguageCode }
```
Then type `LOCALES` as `Record<LanguageCode, () => Catalog>` (the single highest-value change) and `setSelectedLanguage`'s payload as `LanguageCode`. Keep the read/bootstrap/backend clamps — those sit at genuine untrusted boundaries (redux-persist `any`, MMKV `string | undefined`, i18next calling `read` with an arbitrary string) — but add an `isLanguageCode` type guard so they narrow instead of re-widening to `string`. The Crowdin/dynamic-codes counter-argument doesn't apply here: locales are closed by construction (literal `require`s), so a new one can't appear without editing `locales.ts` anyway.

packages/core-mobile/app/i18n/backend.test.ts
```diff
@@ -0,0 +1,18 @@
...
+    RequireBackend.read?.('es-ES', 'translation', cb)
+    expect(cb).toHaveBeenCalledWith(
+      null,
+      expect.objectContaining({ Settings: 'Ajustes' })
+    )
+  })
+
+  it('errors (not throws) on an unsupported locale', () => {
+    const cb = jest.fn()
+    RequireBackend.read?.('xx-XX', 'translation', cb)
```
bogdandobritoiu, line 31: **[Test gap]** Covered: supported-locale success and the `!load` unsupported branch. Not covered: the inner `try/catch` where `load()` itself throws (corrupt/failing require-thunk). It's a distinct branch the file explicitly handles — register a throwing thunk (or mock the module) and assert `callback(expect.any(Error), false)`.

packages/core-mobile/app/store/schemaMigration/schemaMigrations.test.ts
```diff
@@ -33,3 +34,21 @@ describe('migration 30 — filterSmallUtxos default-fill', () => {
     expect(after.settings.advanced.filterSmallUtxos).toBe(false)
   })
 })
+
+describe('migration 31 — language default-fill', () => {
```
bogdandobritoiu, line 38: **[Test gap, minor]** The fill-vs-preserve cases cover the core contract well. One worth adding: `migrations[31]({})` with a fully-absent `state.settings` — the source relies on `state.settings?.language?.selected` optional chaining for very old persisted states, so pin that it yields `{ selected: 'en-US' }`.

packages/core-mobile/app/store/settings/language/listeners.test.ts
```diff
@@ -0,0 +1,30 @@
+// Inline mocks; read them back through the imports (no outer-var refs).
+jest.mock('utils/mmkv/storages', () => ({ commonStorage: { set: jest.fn() } }))
+jest.mock('i18next', () => ({
+  __esModule: true,
```
bogdandobritoiu, line 16: **[Test gap]** `changeLanguage` is mocked as always-resolve, so the `.catch` failure path in `listeners.ts` (the whole reason that `.catch` exists) has zero coverage — dropping it would pass every test. Add a case where `changeLanguage` returns a rejected promise and assert `Logger.error` fires (and no unhandled rejection escapes); needs a `jest.mock('utils/Logger', ...)`.

packages/core-mobile/index.js
ruijia1in: Fixed in 97d77a3. Confirmed `Logger.error` only forwards to Sentry after `shouldLogErrorToSentry` is armed later in bootstrap, so a cold-start i18n failure was reaching Sentry nowhere in release. The `.catch` now also calls `SentryService.captureException` (SentryService.init() has already run by this point in index.js) with an `errorId` tag. The "init succeeded but seeded empty/missing" case is handled at the seed site — see the catalog guard in `app/i18n/index.ts`.

packages/core-mobile/app/i18n/index.ts
ruijia1in: Fixed in 97d77a3. The active catalog is now loaded explicitly via a `loadActiveCatalog(lng)` helper instead of the inline `?? {}`: the thunk call is wrapped in try/catch (a throw no longer escapes as an opaque bootstrap crash), and a missing/empty catalog is reported via `SentryService.captureException` with an `errorId` tag (Logger doesn't reach Sentry this early) plus `Logger.error` for dev-console visibility.

packages/core-mobile/app/i18n/index.ts
ruijia1in: Fixed in 97d77a3. The comment now reads: `initImmediate` was renamed to `initAsync` in i18next v24; v26 only removed the deprecated alias.

packages/core-mobile/app/store/settings/language/listeners.ts
ruijia1in: Partly addressed in 97d77a3: added an `errorId` tag to the `.catch` so the failure surfaces in Sentry (this runs at runtime, where error-forwarding is armed), and added the missing reject-path test in `listeners.test.ts` (asserts `Logger.error` fires, no unhandled rejection).

I did **not** reorder `changeLanguage` before the MMKV/reducer writes: the optimistic write is intentional, and `changeLanguage` on a bundled catalog that RequireBackend resolves synchronously can't realistically reject — reordering would trade the cold-start-seed guarantee for a near-impossible failure (and a toast for an event the user can't perceive). Happy to reorder if you still prefer it — leaving this thread open for your call.

packages/core-mobile/app/store/settings/language/listeners.ts
ruijia1in: Fixed in 97d77a3. `addLanguageListeners` now uses `isAnyOf(setSelectedLanguage, onRehydrationComplete)`, mirroring `settings/appearance`, so i18n reconciles from persisted Redux on every cold start (self-heals the MMKV-cleared-but-Redux-intact case). Added a wiring test asserting the matcher fires on both actions.

packages/core-mobile/app/store/settings/language/types.ts
ruijia1in: Fixed in 97d77a3. Added a `LanguageCode` union derived from `SUPPORTED_LANGUAGES` and typed `LOCALES` as `Record<LanguageCode, () => Catalog>` — the drift guard you flagged as highest-value: adding a locale to one list without the other is now a compile error. Added an `isLanguageCode` type guard used at the three untrusted boundaries (selector read, bootstrap seed, backend `read`).

Two deviations: (1) kept `setSelectedLanguage`'s payload as `string` at the reducer write-boundary so the runtime clamp + its test stay meaningful (consistent with your "keep clamps at genuine boundaries" point); (2) used plain `as const` rather than `as const satisfies readonly Language[]` because the repo's prettier (2.6.2) can't parse that chain — the union still derives correctly.

packages/core-mobile/app/i18n/backend.test.ts
ruijia1in: Added in 97d77a3. `backend.test.ts` now covers the inner try/catch: a supported locale whose thunk throws asserts `callback(expect.any(Error), false)`. Mocked `./locales` so a *supported* code can reach the throw branch (the unsupported-code guard short-circuits before the thunk); real-catalog content stays asserted in `locales.test.ts`.

packages/core-mobile/app/store/schemaMigration/schemaMigrations.test.ts
ruijia1in: Added in 97d77a3. New case: `migrations[31]({})` with fully-absent `settings` yields `{ selected: 'en-US' }`, pinning the `state.settings?.language?.selected` optional-chain fallback.

packages/core-mobile/app/store/settings/language/listeners.test.ts
ruijia1in: Added in 97d77a3. `listeners.test.ts` now has a case where `changeLanguage` returns a rejected promise and asserts `Logger.error` fires with the `errorId` tag (added a `utils/Logger` mock; flushes microtasks so the `.catch` runs).

packages/k2-alpine/src/components/Primitives.ts
```diff
@@ -27,8 +30,25 @@ export const Text = React.forwardRef<
...
+      (props.sx as { lineHeight?: number } | undefined)?.lineHeight ??
+      (props.style as { lineHeight?: number } | undefined)?.lineHeight
+  })
   return React.createElement(StyledText, {
     ...props,
+    // spread props.sx last so a caller override always wins
+    sx:
+      scriptLineHeight !== undefined
+        ? { lineHeight: scriptLineHeight, ...props.sx }
+        : props.sx,
```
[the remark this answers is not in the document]
ruijia1in: Fixed in 8c0beae. Switched from spreading into `sx` to injecting via the `style` prop (prepended): `style={[{ lineHeight }, props.style]}`. Dripsy composes styles as `[variant, ...style, sx]` (last wins), so the injection overrides the variant's lineHeight but is still overridden by any caller `sx` — object **or** function form — because I no longer object-spread `sx`. The function-form-drop pitfall is gone.

packages/k2-alpine/src/components/Primitives.ts
```diff
@@ -10,6 +10,9 @@ import {
   TouchableOpacity as RNTouchableOpacity
 } from 'react-native'
 import React from 'react'
+import { useDripsyTheme } from 'dripsy'
+import type { TextVariant } from '../theme/tokens/text'
+import { resolveTextLineHeight } from '../utils/tallScriptLineHeight'
```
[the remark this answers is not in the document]
ruijia1in: Addressed in 8c0beae without needing a `StyleSheet` import. The injection now goes through the `style` prop, prepended to `props.style`, and Dripsy/RN flatten the resulting nested/array style. Since caller `style` and `sx` both compose *after* the injection (last-wins), a caller-provided lineHeight always wins by precedence — so there's no need to flatten `style` and detect it.

packages/k2-alpine/src/components/Primitives.ts
```diff
@@ -27,8 +30,25 @@ export const Text = React.forwardRef<
...
+  const variant = (props.variant ?? 'body1') as TextVariant
+  // Relax lineHeight for CJK/Devanagari content using a per-script ratio —
+  // several text variants set lineHeight at/near fontSize, which clips tall
+  // (non-Latin) glyphs on iOS. A caller-set lineHeight (sx or style) always wins.
+  const scriptLineHeight = resolveTextLineHeight({
+    spec: theme.text[variant],
+    children: props.children,
+    callerLineHeight:
+      (props.sx as { lineHeight?: number } | undefined)?.lineHeight ??
+      (props.style as { lineHeight?: number } | undefined)?.lineHeight
```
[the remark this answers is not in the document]
ruijia1in: Addressed in 8c0beae. Instead of flattening `style` to read a caller lineHeight, the injection is prepended to the `style` prop so any caller-provided lineHeight — in a `style` array / registered id, or in `sx` — composes after it and wins by precedence. RN's own style flattening handles the array/registered-id case, so no manual detection is needed.

packages/k2-alpine/src/components/Primitives.ts
```diff
@@ -10,6 +10,9 @@ import {
   TouchableOpacity as RNTouchableOpacity
 } from 'react-native'
 import React from 'react'
+import { useDripsyTheme } from 'dripsy'
+import type { TextVariant } from '../theme/tokens/text'
+import { resolveTallScriptLineHeight } from '../utils/tallScriptLineHeight'
```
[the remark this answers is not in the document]
ruijia1in: Fixed in e9769de. Merged the two `dripsy` imports — `useDripsyTheme` now sits in the top `import { styled, useDripsyTheme, Text as DripsyText, ... } from 'dripsy'` block.

packages/core-mobile/app/i18n/backend.ts
```diff
@@ -0,0 +1,31 @@
...
+    const load = isLanguageCode(language) ? LOCALES[language] : undefined
+    if (!load) {
+      callback(new Error(`i18n: unsupported locale ${language}`), false)
+      return
+    }
+    try {
+      callback(null, load())
+    } catch (e) {
+      callback(e as Error, false)
+    }
```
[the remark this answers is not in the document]
ruijia1in: Fixed in e9769de. The catch now normalizes non-Error throwables: `callback(e instanceof Error ? e : new Error(String(e)), false)`, so i18next always gets an Error shape. Added a test that a thunk throwing a string still yields `callback(expect.any(Error), false)`.

packages/core-mobile/app/i18n/index.ts
```diff
@@ -0,0 +1,92 @@
...
+ */
+export const initI18n = (): Promise<unknown> => {
+  const stored = commonStorage.getString(LANGUAGE_MMKV_KEY)
+  const lng: LanguageCode =
+    stored && isLanguageCode(stored) ? stored : DEFAULT_LANGUAGE
+
+  return i18n
+    .use(RequireBackend)
+    .use(initReactI18next)
+    .init({
```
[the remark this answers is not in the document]
ruijia1in: Fixed in e9769de. Plugin registration is now gated: `if (!i18n.isInitialized) { i18n.use(RequireBackend).use(initReactI18next) }`, then `i18n.init(...)` is still called on each invocation. Repeated `initI18n()` calls (tests / dev fast-refresh) no longer re-register the backend/react plugins.

packages/core-mobile/app/store/schemaMigration/schemaMigrations.ts
```diff
@@ -35,6 +35,7 @@ import {
   FILTER_SMALL_UTXOS_DEFAULT,
   QUICK_SWAPS_DEFAULT
 } from '../settings/advanced/types'
+import { DEFAULT_LANGUAGE } from '../settings/language/types'
```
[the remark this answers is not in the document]
ruijia1in: Fixed in 686ac1df5. Migration 31 now clamps the persisted `settings.language.selected` via `isLanguageCode`: an unsupported/corrupt code normalizes to the default at the migration source, so rehydrated state can no longer hold an invalid locale (previously this relied on the selector clamp indefinitely). Valid codes are still preserved. Added a regression test (`clamps an unsupported/corrupt persisted code back to the default`). tsc + the schemaMigrations tests pass.

packages/core-mobile/app/store/schemaMigration/schemaMigrations.ts
```diff
@@ -573,5 +574,18 @@ export const migrations = {
...
+    return {
+      ...state,
+      settings: {
+        ...state.settings,
+        language: {
+          selected: state.settings?.language?.selected ?? DEFAULT_LANGUAGE
+        }
+      }
+    }
   }
```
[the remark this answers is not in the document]
ruijia1in: Fixed in 686ac1df5. Migration 31 now clamps the persisted `settings.language.selected` via `isLanguageCode`: an unsupported/corrupt code normalizes to the default at the migration source, so rehydrated state can no longer hold an invalid locale (previously this relied on the selector clamp indefinitely). Valid codes are still preserved. Added a regression test (`clamps an unsupported/corrupt persisted code back to the default`). tsc + the schemaMigrations tests pass.

packages/k2-alpine/src/utils/tallScriptLineHeight.ts
```diff
@@ -0,0 +1,81 @@
...
+ */
+export const resolveTallScriptLineHeight = (
+  spec: { fontSize: number; lineHeight: number },
+  children: ReactNode
+): number | undefined => {
+  const script = detectTallScript(children)
+  if (script === undefined) return undefined
+  const relaxed = Math.ceil(spec.fontSize * RATIO[script])
+  return relaxed > spec.lineHeight ? relaxed : undefined
+}
```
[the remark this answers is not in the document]
ruijia1in: Fixed in 686ac1df5. `resolveTallScriptLineHeight` now short-circuits to `undefined` before `detectTallScript` when `spec.lineHeight` already clears `ceil(fontSize * DEVANAGARI_LINE_HEIGHT_RATIO)` — the tallest ratio, so no detected script could exceed it — skipping the flatten + regex scan on every Text render where no adjustment is possible. Behavior-preserving (ceil is monotonic; Devanagari > CJK). Added a fast-path regression test. tsc + the tallScriptLineHeight tests pass.

B0Y3R-AVA (approved):
dev tested

bogdandobritoiu (approved):

PR #4009, opened 2026-07-23
CP-14851: i18n extraction tooling (i18next-cli + ast-grep fallback)

ruijia1in:
## Description

**Ticket: [CP-14851](https://ava-labs.atlassian.net/browse/CP-14851)** (part of epic [CP-14646](https://ava-labs.atlassian.net/browse/CP-14646); discovery [CP-14647](https://ava-labs.atlassian.net/browse/CP-14647))

Establishes the automated **string-extraction tooling** for the i18n sweep, staying inside the i18next/Crowdin ecosystem (parity-safe). Builds on the runtime foundation from CP-14849 (now on `main` via #3999).

**What this adds (tooling-only — no product strings externalized here; that's the later sweep ticket):**

- **`i18next.config.ts`** — [i18next-cli](https://github.com/i18next/i18next-cli) 1.66.2 config aligned to the runtime (`app/i18n/index.ts`): single `translation` namespace, flat natural-English keys, `keySeparator`/`nsSeparator: false`. `extract` scans `app/**` and maintains the EN catalog; `types` generates TS definitions. `removeUnusedKeys: false` so the manually-seeded catalog isn't wiped while there are ~zero `t()` consumers yet.
- **npm scripts** — `i18n:extract`, `i18n:types`, `i18n:instrument`, `i18n:check` (CI staleness gate via `extract --ci`).
- **TS type-safety** — generated `app/i18n/@types/resources.d.ts` + a `declare module 'i18next'` augmentation, so `t()` keys are compile-checked (unknown keys are a type error).
- **ast-grep fallback rules** (`sgconfig.yml` + `scripts/i18n/rules/`) — report-only flags for the patterns i18next-cli `instrument` structurally misses: string-valued JSX props (`title`/`label`/`placeholder`/`message`/`description`), `showSnackbar`/`showAlert` args, and `accessibilityLabel`/`accessibilityHint`.
- **Runtime fix** — `returnEmptyString: false` in `app/i18n/index.ts` so not-yet-translated keys (seeded as `""` for Crowdin) fall back to English instead of rendering blank.

**Dependencies introduced:** `i18next-cli` + `@ast-grep/cli` (both **devDependencies**, not shipped in the app bundle). Their install scripts are disabled via Lavamoat `allowScripts: false` (they work without them).

**Validation gate (this ticket's core deliverable):** ran `instrument` on pilot areas `receive` + `notifications` (throwaway, reverted) → **59.3% hit-rate, 45.8% false-positive rate**, plus behavior-breaking string corruption and foundation-duplicating scaffolding. **Decision: fall back to ast-grep/ts-morph for the wrap; retain i18next-cli for `extract` + types** (identical JSON output → Crowdin parity holds). Full report in the ticket comment + design doc.

## Screenshots/Videos

N/A — tooling-only change with no UI/runtime surface (the only runtime touch is the `returnEmptyString` fallback config, which has no visible effect until the sweep adds translatable keys).

## Testing

**Dev Testing** — tooling-only, so no in-app QA. Verify from a clean checkout of this branch (monorepo root unless noted):

```bash
# 1. Install + setup must succeed (Lavamoat allow-scripts policy for the new devDeps)
yarn install && yarn setup
#    -> no "@lavamoat/allow-scripts has detected dependencies without configuration"

# 2. Lockfile consistent
yarn install --immutable            # -> no YN0028

# 3. Extraction is non-destructive (seed keys preserved)
yarn core i18n:extract
node -e "const j=require('./packages/core-mobile/app/i18n/locales/en-US/translation.json'); console.log(Object.keys(j).length, ['Settings','Send','Receive','Account'].every(k=>k in j))"
#    -> 4 true

# 4. Types generate + compile; key typing enforced (t('Settings') ok, t('nope') is a compile error)
yarn core i18n:types && yarn core tsc

# 5. CI staleness gate: exits 0 when current; add a t('New') call and re-run -> non-zero
yarn core i18n:check

# 6. ast-grep rules flag missed patterns, report-only (no file edits)
cd packages/core-mobile && ./node_modules/.bin/ast-grep scan -c sgconfig.yml app/new/features/notifications

# 7. i18n unit tests still pass (returnEmptyString change)
yarn core test app/i18n
```

Jira ticket moved to Testing.

## Build Numbers

- iOS: 9403
- Android: 9404
- Pipeline: https://app.bitrise.io/app/7d7ca5af7066e290/pipelines/c9c34e44-f7ad-4e57-bb55-e28a227e9a68

## Checklist

- [ ] I have included screenshots / videos of android and ios (N/A — no UI)
- [x] I have added/updated necessary unit tests (i18n foundation tests pass with the runtime change)
- [x] I have updated the documentation (design doc + report in Obsidian; Jira comment)

[CP-14851]: https://ava-labs.atlassian.net/browse/CP-14851?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CP-14646]: https://ava-labs.atlassian.net/browse/CP-14646?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CP-14647]: https://ava-labs.atlassian.net/browse/CP-14647?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/core-mobile/package.json
```diff
@@ -328,6 +333,7 @@
     "detox": "20.41.1",
     "eslint": "8.50.0",
     "eslint-plugin-avalabs-mobile": "workspace:*",
+    "i18next-cli": "^1.66.2",
```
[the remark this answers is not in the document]
ruijia1in: Investigated with `yarn why react`: the only source of `react@19.2.8` (via `^19.2.7`) is `i18next-cli` → `ink` (its terminal-UI renderer). `i18next-cli` is a **devDependency**, used only at build/CLI time — app code never imports it, so Metro never bundles it and this React never reaches the RN runtime. (`react@18.3.1` via `^18.2.0` already existed pre-PR, confirming a lockfile with multiple reacts is a pre-existing dev-only artifact, not a runtime condition.) Forcing a `resolutions` to collapse `^19.2.7`→19.2.3 would risk breaking ink for zero runtime benefit, so leaving as-is — no duplicate-React-in-bundle risk here.

packages/core-mobile/package.json
```diff
@@ -24,7 +24,11 @@
     "devicefarm:run": "./scripts/devicefarm/run-devicefarm.sh",
     "appium:smoke": "IS_SMOKE=true wdio run ./e2e-appium/wdio.conf.ts --mochaOpts.grep='\\[(s|S)moke\\]'",
     "appium:smokeIos": "IS_SMOKE=true PLATFORM=ios wdio run ./e2e-appium/wdio.conf.ts --mochaOpts.grep='\\[(s|S)moke\\]'",
-    "appium:smokeAndroid": "IS_SMOKE=true PLATFORM=android wdio run ./e2e-appium/wdio.conf.ts --mochaOpts.grep='\\[(s|S)moke\\]'"
+    "appium:smokeAndroid": "IS_SMOKE=true PLATFORM=android wdio run ./e2e-appium/wdio.conf.ts --mochaOpts.grep='\\[(s|S)moke\\]'",
+    "i18n:extract": "i18next-cli extract",
+    "i18n:types": "i18next-cli types",
+    "i18n:instrument": "i18next-cli instrument",
```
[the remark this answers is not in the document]
ruijia1in: Good catch — changed `i18n:instrument` to `i18next-cli instrument --dry-run` so a bare `yarn i18n:instrument` only previews. Since ast-grep is the chosen (report-only) fallback and instrument is just an optional accelerator, actually mutating source is now an explicit, deliberate step (invoke the binary without `--dry-run`). Fixed in 1ddf1d418.

packages/core-mobile/app/i18n/index.ts
```diff
@@ -86,6 +86,11 @@ export const initI18n = (): Promise<unknown> => {
     partialBundledLanguages: true,
     interpolation: { escapeValue: false },
     returnNull: false,
+    // Treat an empty-string value as missing so it falls back to `fallbackLng`
+    // (English) instead of rendering blank. The extraction pipeline seeds
+    // not-yet-translated keys as "" in non-EN catalogs (for Crowdin to fill),
+    // so without this an untranslated key would show empty in other languages.
+    returnEmptyString: false,
```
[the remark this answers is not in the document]
ruijia1in: Added a unit test in `app/i18n/index.test.ts` that seeds an empty-string value for an existing key in a non-EN catalog (es-ES) with the EN source present, and asserts `t()` returns the English value — proving the `returnEmptyString:false` fallback. Fixed in 1ddf1d418.

B0Y3R-AVA (commented):
Adversarial review (two independent passes, findings verified against i18next-cli v1.66.2 source, this branch, and the yarn.lock diff). No blockers for a tooling-only PR — overall this is in good shape and the config comments match upstream behavior (`defaultValue` arg order, `extract --ci`, `instrument --dry-run`, sorted/idempotent extraction all check out). Four substantive comments inline (`.ts` files invisible to the snackbar rule, missing `nsSeparator` in the type augmentation, `types.output` collision quirk, CI gate coverage/wiring), plus a couple of minor notes.

**Minor (no inline anchor):**
- The `i18n-jsx-string-prop` rule will flag non-user-facing `title`/`label`/`message`/`description` strings too (analytics payloads, dev-only props) — acceptable noise for report-only, just calibrating expectations for the sweep.
- Verified `@swc/core` genuinely arrives via `i18next-cli` / `i18next-resources-for-ts` in the lockfile (real new transitive dep, and it does ship a postinstall), so an allowScripts entry is needed — see inline note on the path naming.

packages/core-mobile/scripts/i18n/rules/snackbar-alert-args.yml
```diff
@@ -0,0 +1,14 @@
+id: i18n-snackbar-alert-arg
+language: tsx
```
B0Y3R-AVA, line 5: **Important:** `language: tsx` means ast-grep only scans `.tsx` files — but `showSnackbar`/`showAlert` calls routinely live in plain `.ts` hooks/utils (e.g. `app/new/features/stake/hooks/useStakeBalanceGuard.ts:54`, `app/contexts/DeeplinkContext/utils/handleDeeplink.ts:189`). Those are silently invisible to the sweep report. The two JSX rules are fine as `tsx`-only (JSX attrs can only exist in `.tsx`), but this rule needs a `language: typescript` twin (or a `languageGlobs` override in `sgconfig.yml`).

Two smaller notes while here: (1) the un-anchored `regex: (showSnackbar|showAlert)$` is what lets member calls like `Alert.showAlert(...)` match (node-text suffix) — worth a comment saying that's deliberate, since it also means a hypothetical `fooshowAlert()` would match; (2) `kind: string` doesn't match template literals (`template_string` in tree-sitter), so interpolated copy like `app/new/features/onboarding/hooks/useRegisterAndAuthenticateFido.ts:73` falls through here *and* through i18next-cli `instrument` — worth a line in the rule `message` or docs so the sweep doesn't treat the report as exhaustive.

packages/core-mobile/app/i18n/@types/i18next.d.ts
```diff
@@ -0,0 +1,14 @@
...
+
+// Bind the generated catalog to i18next so t() keys are type-checked against
+// the EN source. Single `translation` namespace (see app/i18n/index.ts).
+// `resources.d.ts` default-exports the `Resources` interface (a type), so it
+// is imported and used directly rather than via `typeof`.
+declare module 'i18next' {
+  interface CustomTypeOptions {
+    defaultNS: 'translation'
+    resources: Resources
+    keySeparator: false
```
B0Y3R-AVA, line 15: **Important:** this augmentation is missing `nsSeparator: false`. The runtime init sets both `keySeparator: false` and `nsSeparator: false` (`app/i18n/index.ts`), but i18next's `TypeOptions` defaults `nsSeparator` to `':'` — so any future natural-English key containing a colon (plenty of real copy does, e.g. `app/new/routes/signup.tsx:123`, `WalletConnectScanScreen.tsx:87`) will be type-parsed as `namespace:key` while the runtime resolves it as a flat literal key. One-line fix:

```ts
    keySeparator: false,
    nsSeparator: false
```

packages/core-mobile/i18next.config.ts
```diff
@@ -0,0 +1,47 @@
...
+    // Note: i18next-cli 1.66.x calls defaultValue as
+    // (key, namespace, language, value) — the language is the 3rd arg.
+    defaultValue: (key: string, _namespace: string, language: string) =>
+      language === 'en-US' ? key : ''
+  },
+  types: {
+    // Generate types from the EN source of truth (the flat `translation`
+    // namespace written by `extract`).
+    input: ['app/i18n/locales/en-US/translation.json'],
+    output: 'app/i18n/@types/resources.d.ts'
```
B0Y3R-AVA: **Important (fragile config):** in i18next-cli 1.66.2, `types.output` is the *augmentation* file (`i18next.d.ts`) and the Resources interface goes to `types.resourcesFile`, which defaults to `dirname(output)/resources.d.ts` (`src/types-generator.ts:126-127`). Pointing `output` at `resources.d.ts` makes `output === resourcesFile` — it only works because the generator writes the resources file first and then skips the main output when the path already exists (`types-generator.ts:207-219`). That's a write-order coincidence, not a contract; an upstream reorder would clobber `resources.d.ts` with an augmentation.

Idiomatic fix that keeps everything else working:

```ts
  types: {
    input: ['app/i18n/locales/en-US/translation.json'],
    output: 'app/i18n/@types/i18next.d.ts' // existing hand-written file is preserved (generator skips existing outputs)
    // resourcesFile defaults to app/i18n/@types/resources.d.ts
  }
```

packages/core-mobile/package.json
```diff
@@ -24,7 +24,11 @@
     "devicefarm:run": "./scripts/devicefarm/run-devicefarm.sh",
     "appium:smoke": "IS_SMOKE=true wdio run ./e2e-appium/wdio.conf.ts --mochaOpts.grep='\\[(s|S)moke\\]'",
     "appium:smokeIos": "IS_SMOKE=true PLATFORM=ios wdio run ./e2e-appium/wdio.conf.ts --mochaOpts.grep='\\[(s|S)moke\\]'",
-    "appium:smokeAndroid": "IS_SMOKE=true PLATFORM=android wdio run ./e2e-appium/wdio.conf.ts --mochaOpts.grep='\\[(s|S)moke\\]'"
+    "appium:smokeAndroid": "IS_SMOKE=true PLATFORM=android wdio run ./e2e-appium/wdio.conf.ts --mochaOpts.grep='\\[(s|S)moke\\]'",
+    "i18n:extract": "i18next-cli extract",
+    "i18n:types": "i18next-cli types",
+    "i18n:instrument": "i18next-cli instrument --dry-run",
+    "i18n:check": "i18next-cli extract --ci"
```
B0Y3R-AVA: **Important:** two gaps in the staleness gate as shipped:

1. Nothing runs it — no `bitrise.yml`/workflow change in this PR, so `i18n:check` is a script without a CI hook. If wiring lands in a later ticket, worth saying which one.
2. It doesn't cover types: `extract --ci` won't catch a stale `app/i18n/@types/resources.d.ts` (and with `removeUnusedKeys: false` it also won't fail on deleted/renamed keys, by design). `types --ci` exists in 1.66.2 — suggest:

```json
"i18n:check": "i18next-cli extract --ci && i18next-cli types --ci"
```

packages/core-mobile/package.json
```diff
@@ -399,7 +405,9 @@
       "@avalabs/crypto-nitro>react-native-nitro-modules": true,
       "@avalabs/bridge-unified>@lombard.finance/sdk>@lombard.finance/sdk-common>vite>esbuild": false,
       "@shopify/react-native-skia": false,
-      "@avalabs/k2-alpine>@shopify/react-native-skia": false
+      "@avalabs/k2-alpine>@shopify/react-native-skia": false,
+      "@ast-grep/cli": false,
+      "ts-node>@swc/core": false
```
B0Y3R-AVA, line 410: **Minor:** the new `@swc/core` in the lockfile actually arrives via `i18next-cli` and `i18next-resources-for-ts` (both declare it as a direct dep); `ts-node` only optional-peers it. If `ts-node>@swc/core` is the canonical path `yarn allow-scripts auto` emitted post-hoisting, fine (CI install passing suggests it is) — just confirming this wasn't hand-copied from another repo, since a wrong path here would resurface as an install warning later.

packages/core-mobile/app/i18n/index.test.ts
```diff
@@ -31,4 +31,17 @@ describe('initI18n', () => {
...
+    mockGetString.mockReturnValue('es-ES')
+    await initI18n()
+    // The extraction pipeline seeds not-yet-translated keys as "" in non-EN
+    // catalogs (for Crowdin to fill), with the EN source of truth present.
+    // Simulate that for an existing key: EN has the value, es-ES is empty.
+    i18n.addResource('en-US', 'translation', 'Send', 'Send')
+    i18n.addResource('es-ES', 'translation', 'Send', '')
+    // returnEmptyString:false → the empty es-ES value is treated as missing and
+    // falls back to en-US ('Send') instead of rendering blank.
+    expect(i18n.t('Send')).toBe('Send')
```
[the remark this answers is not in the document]
ruijia1in: Good catch — the key and en-US value were both `Send`, so the assertion couldn't distinguish a real fallback from t() echoing the key. Changed the en-US value to a sentinel (`__EN_FALLBACK__`) and assert on that, so it now explicitly proves the fallback returns the en-US *value*. Fixed in fa6ce2e1d.

packages/core-mobile/i18next.config.ts
```diff
@@ -0,0 +1,47 @@
...
+    // referenced in source) instead of pruning them. i18next-cli defaults
+    // removeUnusedKeys to true, which would wipe the seed catalog while
+    // there are ~zero t()/<Trans> consumers yet.
+    removeUnusedKeys: false,
+    // Natural-English keys: the key IS the English string. EN source gets
+    // key===value; every other locale stays empty for Crowdin to fill.
+    // Note: i18next-cli 1.66.x calls defaultValue as
+    // (key, namespace, language, value) — the language is the 3rd arg.
+    defaultValue: (key: string, _namespace: string, language: string) =>
+      language === 'en-US' ? key : ''
```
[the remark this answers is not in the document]
ruijia1in: Agreed — reworded to: a NEW key is written key===value in EN and `""` in every other locale (Crowdin placeholder), while existing non-EN translations are preserved (removeUnusedKeys:false). Fixed in fa6ce2e1d.

packages/core-mobile/scripts/i18n/rules/snackbar-alert-args.yml
ruijia1in: All three addressed (fa6ce2e1d). Confirmed the gap is real — 30 non-`.tsx` files call showSnackbar/showAlert (incl. the two you cited). Added a `language: typescript` twin `snackbar-alert-args-ts.yml` (verified it now flags `useStakeBalanceGuard.ts:54`). Also added rule comments: (1) the un-anchored `(showSnackbar|showAlert)$` is deliberate to catch member calls like `Alert.showAlert(...)`, with the `fooShowAlert` trade-off noted; (2) `kind: string` doesn't match `template_string`, so interpolated copy falls through here and through instrument — now called out in each rule's `message` so the report isn't treated as exhaustive. (The two JSX rules stay tsx-only, correct.)

packages/core-mobile/app/i18n/@types/i18next.d.ts
ruijia1in: Fixed (fa6ce2e1d) — added `nsSeparator: false` to the augmentation so a natural-English key containing `:` isn't type-parsed as `namespace:key` while the runtime resolves it flat. Mirrors app/i18n/index.ts. Thanks.

packages/core-mobile/i18next.config.ts
ruijia1in: Verified against the installed 1.66.2 (`types-generator` cjs: resourcesFile defaults to `dirname(output)/resources.d.ts`; always writes resourcesFile, then skips `output` only when it already exists) — you're right, our `output===resourcesFile` was a write-order coincidence. Applied your idiomatic fix (fa6ce2e1d): `output: app/i18n/@types/i18next.d.ts` + explicit `resourcesFile: app/i18n/@types/resources.d.ts`. Ran `yarn i18n:types`: the hand-written i18next.d.ts (with defaultNS + keySeparator/nsSeparator) is preserved via the skip-existing path, resources.d.ts is regenerated, and `i18n:check` is green.

packages/core-mobile/package.json
ruijia1in: (2) Fixed (fa6ce2e1d): `i18n:check` is now `i18next-cli extract --ci && i18next-cli types --ci` (confirmed `types --ci` exists in 1.66.2), so a stale resources.d.ts also fails. (1) Correct that nothing runs it yet — intentionally no bitrise.yml/workflow change in this tooling PR; the CI hook belongs with the Crowdin bidirectional-sync ticket (CP-14853), where push/pull GH Actions land and a staleness gate has real strings to guard. Noted there.

packages/core-mobile/package.json
ruijia1in: Confirmed not hand-copied: the `ts-node>@swc/core` path is exactly what Lavamoat's `allow-scripts` emitted in its "packages missing configuration" output (alongside `@ast-grep/cli`) — I pasted those verbatim. `yarn install && yarn setup` + CI install both pass with it, so the hoisted path resolves. Agree @swc/core also arrives via i18next-cli directly; Lavamoat just canonicalizes to one post-hoist path, and this is the one it chose.

PR #4027, opened 2026-08-06
CP-14918: dedupe watchlist query observers behind a shared provider

B0Y3R-AVA:
## Description

**Ticket: [CP-14918](https://ava-labs.atlassian.net/browse/CP-14918)**

Dedupes watchlist React Query observers. Before this change, every `useWatchlist()` call site (20+ across the app) instantiated its own `useTopTokensQuery`/`useTrendingTokensQuery` observer. A changed-fields histogram over 1,167 `observerResultsUpdated` events (idle + 2 navs) showed **93.7% fired with nothing changed** in the tracked fields - this was tracked-props over-subscription / duplicate `QueryObserver` mounts, not a refetch storm (structural sharing was working fine).

Fix: a new `WatchlistQueriesProvider` (mounted once in `(signedIn)/_layout.tsx`, wrapping the whole authenticated app) hoists both queries to a single shared observer, exposed through a narrow, memoized context (`watchlistQueriesContext.ts`) read by `useTopTokens`/`useGetTrendingTokens`. The context value is deliberately narrow (`{data, isLoading, isRefetching, refetch}`) with tracked-props defaults (not `notifyOnChangeProps: 'all'`) - a first-round version regressed to `'all'` plus the raw observer result as the context value, which reintroduced the exact fan-out one level up (a Context has no per-field selector, so any of the 20+ consumers re-rendering on the global 10s `staleTime` elapsing re-rendered all of them). That was caught and fixed before this state.

**Known tradeoff, not fixed here (Open decision, tracked as [CP-14922](https://ava-labs.atlassian.net/browse/CP-14922)):** this fix drops the old per-call-site `enabled: useIsFocused()` gate that paused both queries when no visible screen needed them. Building that back safely (ref-counted registration across 20+ call sites, race-safe under StrictMode double-effects) was judged too risky to land in this pass. Concretely: both `watchlistTopTokens` and `watchlistTrendingTokens` now poll continuously (2-minute `refetchInterval`, 10s global `staleTime`) while signed in and foregrounded, including on screens that import neither hook. This is bounded and low-risk (public market data, no balance/signing exposure) but is a real, deliberate behavior change. It matters more on iOS after PR 6 lands, since the six tabs that stay permanently live on iOS (once visited) all subscribe to these same hot keys.

Measured win: 93.7% of no-op `observerResultsUpdated` notifications eliminated for the watchlist keys.

No dependency on PR 1/2; can land independently. Does not stack.

## Screenshots/Videos

N/A - no visual change.

## Testing

**Dev Testing**

- Sign in, visit several screens that read watchlist data (Portfolio, Track, a token detail screen, etc.) and confirm data still loads and updates correctly (top tokens / trending tokens lists render, refresh correctly).
- Confirm `refetch()` from any consumer still triggers a real refetch visible to all consumers (shared observer, not per-consumer).
- Leave the app foregrounded and idle for several minutes; confirm both queries keep polling on their `refetchInterval` even when no watchlist-consuming screen is visible (this is the accepted tradeoff above, not a bug - just verify it doesn't regress further, e.g. into a refetch storm).
- `yarn workspace @avalabs/core-mobile test` - `WatchlistQueriesProvider.test.tsx`, plus `useTopTokens`/`useGetTrendingTokens` coverage. Confirm the suite **exits cleanly without `--forceExit`** (an earlier round had an un-cleared probe `setInterval` leak here; fixed with a refcounted start/stop pair - worth re-confirming on this diff).

## Checklist

[CP-14918]: https://ava-labs.atlassian.net/browse/CP-14918?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CP-14922]: https://ava-labs.atlassian.net/browse/CP-14922?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core-mobile/app/hooks/watchlist/useGetTrendingTokens.ts
+++ b/packages/core-mobile/app/hooks/watchlist/useGetTrendingTokens.ts
@@ -30,7 +30,7 @@ export const useTrendingTokensQuery = (): UseQueryResult<
       return tokens ?? []
     },
     refetchInterval: 120000 // 2 mins
-    // Deliberately NOT `notifyOnChangeProps: 'all'` -- see useTopTokens.ts. CP-14918.
+    // Deliberately NOT `notifyOnChangeProps: 'all'` -- see useTopTokens.ts.
   })
 }
 
@@ -46,7 +46,7 @@ export const useGetTrendingTokens = <TData = TrendingToken[]>(
   const query = useContext(TrendingTokensQueryContext)
   if (!query) {
     throw new Error(
-      'useGetTrendingTokens() must be used within a <WatchlistQueriesProvider> (CP-14918: shared observer fix)'
+      'useGetTrendingTokens() must be used within a <WatchlistQueriesProvider>'
     )
   }
 
--- a/packages/core-mobile/app/hooks/watchlist/useGetTrendingTokens.ts
+++ b/packages/core-mobile/app/hooks/watchlist/useGetTrendingTokens.ts
@@ -35,20 +35,10 @@ export const useTrendingTokensQuery = (): UseQueryResult<
 }
 
 /**
- * Shared, deduped read of the trending-tokens query, with the same `select`
- * ergonomics the old per-call-site `useQuery({ select })` had. `select` is
- * applied locally (memoized on `[select, query.data]`, matching react-query's
- * own per-observer select memoization) rather than inside the shared
- * `useQuery` call, since a `select` baked into the single shared observer
- * would only ever serve whichever caller happened to render first.
- *
- * Must be called under <WatchlistQueriesProvider> (mounted once in
- * `(signedIn)/_layout.tsx`). Throws otherwise, rather than silently creating
- * a second observer.
- *
- * Returns a narrow `{ data, isLoading, isRefetching, refetch }` shape, not
- * the full `UseQueryResult` -- see the comment on `WatchlistQueryValue` in
- * watchlistQueriesContext.ts.
+ * `select` is applied locally (memoized on `[select, query.data]`, matching
+ * react-query's own per-observer select memoization) rather than baked into
+ * the shared query -- a `select` inside the single shared observer would
+ * only ever serve whichever caller happened to render first.
  */
 export const useGetTrendingTokens = <TData = TrendingToken[]>(
   select?: (data: TrendingToken[]) => TData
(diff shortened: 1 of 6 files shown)


packages/core-mobile/app/hooks/watchlist/useGetTrendingTokens.ts
```diff
@@ -26,14 +29,56 @@ export const useGetTrendingTokens = <TData = TrendingToken[]>(
...
+ * would only ever serve whichever caller happened to render first.
+ *
+ * Must be called under <WatchlistQueriesProvider> (mounted once in
+ * `(signedIn)/_layout.tsx`). Throws otherwise, rather than silently creating
+ * a second observer.
+ *
+ * Returns a narrow `{ data, isLoading, isRefetching, refetch }` shape, not
+ * the full `UseQueryResult` -- see the comment on `WatchlistQueryValue` in
+ * watchlistQueriesContext.ts.
+ */
```
gergelylovas, line 42: Same as in the other PR regarding comments. This comment doesn't really provide value as it only describes how regular contexts work and what is visible in the code itself

gergelylovas (changes requested):

packages/core-mobile/app/hooks/watchlist/useGetTrendingTokens.ts
B0Y3R-AVA: Removed in c4ea698de — trimmed that block to just the select-applied-locally rationale (the one non-obvious why) and deleted the sibling what-narration JSDoc above useTopTokens as well.

packages/core-mobile/app/hooks/watchlist/watchlistQueriesContext.ts
```diff
@@ -0,0 +1,35 @@
+import { UseQueryResult } from '@tanstack/react-query'
```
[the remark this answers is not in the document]
B0Y3R-AVA: Fixed in c4ea698de — switched to `import type`.

gergelylovas (approved):

PR #4030, opened 2026-08-06
CP-14918: restructure PriceChart Skia rendering and VolumeRow highlight

B0Y3R-AVA:
## Description

**Ticket: [CP-14918](https://ava-labs.atlassian.net/browse/CP-14918)**

k2-alpine `PriceChart` performance restructure. The chart's mount commit was the single largest app-owned cost in the token-detail tap-to-paint window; this PR cuts its self-time from ~517-716ms to ~203ms median (dev rig, Pixel 8 Pro; absolute numbers are dev-build only, methodology in `docs/cp14918-portfolio-performance-session.md`). `PriceChart` is shared with Track (CP-14309), so both surfaces are affected.

Changes:

- `Series.tsx` - `Candles` collapses from 2N Skia elements (per-candle `<Line>` + `<RoundedRect>`) to 4 `SkPath`s (up/down bodies, up/down wicks). Wicks draw before bodies so paint order matches the original. `Candles` renders in both line and candlestick modes, so this helps the default path too.
- `PriceChart.tsx` - the label font becomes a module-level singleton (`useFont` has no internal cache; every mount was reallocating a native `SkTypeface`), with a retry on failed load (a failed promise is not cached forever). Worklets read `linePoints`/`candles` geometry from flattened `Float32Array` shared values instead of object-array closures, avoiding a shareable-conversion per dependency change.
- `VolumeRow.tsx` - bar highlight restructured from one `makeMutable` per candle to a single static idle `SkPath` plus two highlight `RoundedRect`s driven by one `SharedValue`. The original two-bar crossfade drag behavior is preserved: `volumeCrosshairWeights` (new pure worklet in `helpers.ts`, unit-tested) reproduces main's per-candle weight formula, and the rect opacity is compositing-compensated (`crossfadeRectOpacity`) so idle-path-plus-rect source-over composition lands on exactly the original per-bar opacity for every weight. The compensation identity is pinned by tests.
- `helpers.ts` - the four `Intl.DateTimeFormat().format()` call sites are replaced with hand-rolled month/weekday-table formatters (~4ms per Intl call, ~96 calls per mount). Two device-found parity divergences are handled and byte-pinned by tests: midnight renders `00:00` (Hermes/Android ICU h23, not Node ICU's `24:00`), and the AM/PM separator is U+202F (narrow no-break space, ICU 72+), not an ASCII space.
- `ChartHeader.tsx` / `ChartFooter.tsx` / `hooks.ts` - per-candle display strings are formatted lazily: the eager mount path formats only the last candle (O(1)); the full array computes on idle or first crosshair activation (`useIsFullFormatNeeded`). `ChartFooter`'s volume formatting is additionally gated on `showVolume`, since in line mode (the default) that text is never visible.

Known and accepted, called out for reviewers:

- The hand-rolled date formatters hardcode en-US; the app ships 11 UI languages. Four sibling screens already hardcode en-US, but a real product/i18n decision is tracked as [CP-14923](https://ava-labs.atlassian.net/browse/CP-14923).
- The gesture memo in `PriceChart.tsx` still bails from React Compiler (a `SharedValue`-in-deps pattern shared by three effects in this file); tracked in the compiler-bailout backlog [CP-14925](https://ava-labs.atlassian.net/browse/CP-14925).
- `lineYsSV`/`volumesSV` lazy-init closes the mount-window staleness but not the range-switch case (needs an in-progress drag during a range switch to trigger; bounds-safe and self-heals within one JS tick). Accepted as low severity.

No dependency on the core-mobile CP-14918 PRs; can land independently.

## Screenshots/Videos

To be added: chart render in line and candlestick modes, plus a crosshair-drag recording showing the volume-bar crossfade.

## Testing

**Dev Testing**

iOS: 9457
Android: 9458

- Open a token detail screen (line mode): chart renders, crosshair drag works, header values update while dragging.
- Switch to candlestick mode: candles and volume bars render correctly; drag the crosshair slowly between two bars and confirm the highlight crossfades (no snapping) and matches the pre-restructure behavior visually.
- Switch ranges (1D/1W/...) repeatedly, including mid-drag, and confirm no stale geometry or crash.
- Kill and relaunch with airplane mode toggled during first load to exercise the font-load retry path.
- `yarn workspace @avalabs/k2-alpine test` - full suite; `helpers.test.ts` carries the formatter byte-pinning and the crossfade/compositing identity tests.

**QA Testing**

- Verify chart date/time labels against a device set to a non-English locale: labels are expected to render in English (accepted, CP-14923) - confirm nothing renders blank or garbled.
- Compare crosshair drag behavior on this build vs production on the same token/range: bar highlight brightness and crossfade should be indistinguishable.

## Checklist

[CP-14918]: https://ava-labs.atlassian.net/browse/CP-14918?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CP-14923]: https://ava-labs.atlassian.net/browse/CP-14923?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/k2-alpine/src/components/Chart/PriceChart/VolumeRow.tsx
+++ b/packages/k2-alpine/src/components/Chart/PriceChart/VolumeRow.tsx
@@ -35,16 +35,10 @@ const BAR_WIDTH_RATIO = 0.6
 type BarGeom = { x: number; barHeight: number; radius: number }
 
 /**
- * Bar highlight: one static `SkPath` for idle bars + two dynamic
- * `RoundedRect`s for the two nearest bars, driven by a single `SharedValue`
- * holding both indices and their crossfade weights (was: one `makeMutable`
- * per candle). Crossfades between the two nearest candles exactly like
- * main's original per-candle implementation — see `volumeCrosshairWeights`
- * in `helpers.ts` for the interpolation this restores. Because the
- * highlight rects paint on top of the idle `Path` (which already painted
- * those same bars), their opacity is compositing-compensated via
- * `crossfadeRectOpacity` so `idlePath + rect` composes to main's per-bar
- * target opacity exactly, not a double-counted overshoot. CP-14918 item 2.
+ * One static Path for all idle bars + two RoundedRects for the bars
+ * bracketing the crosshair, replacing one makeMutable and one rect per
+ * candle. Rect opacity is compositing-compensated (`crossfadeRectOpacity`)
+ * because it paints over bars the idle Path already drew. CP-14918.
  */
 export const VolumeRow: FC<Props> = ({
   candles,
@@ -95,8 +89,10 @@ export const VolumeRow: FC<Props> = ({
     return { barGeom: geom, idlePath: path }
   }, [candles, innerWidth, height, barWidth, maxVolume])
 
-  // Mirrored into a SharedValue via an effect rather than closed over
-  // directly inside the worklets below.
+  // Worklets capture by value: closing over `barGeom` would copy all N geoms
+  // into the UI runtime once per derived value, per render. One SharedValue
+  // copies once per geometry change — written in an effect, not during
+  // render, because Reanimated doesn't support SharedValue writes mid-render.
   const barGeomSV = useSharedValue<BarGeom[]>(barGeom)
   useEffect(() => {
     barGeomSV.value = barGeom
@@ -117,9 +113,8 @@ export const VolumeRow: FC<Props> = ({
     [candles.length, innerWidth]
   )
 
-  // Low bar — the candle at or just below the crosshair. Opacity is
-  // compositing-compensated (see `crossfadeRectOpacity`) since this rect
-  // paints on top of `idlePath`, which already painted the same bar.
+  // Low/high bar = the two candles bracketing the crosshair. Opacity goes
+  // through `crossfadeRectOpacity` because these paint over `idlePath`.
   const lowOpacity = useDerivedValue(() => {
     const c = crossfadeSV.value
     return c.lowIndex === NO_HIGHLIGHT_INDEX
@@ -151,8 +146,6 @@ export const VolumeRow: FC<Props> = ({
     return g ? g.radius : 0
   })
 
-  // High bar — the candle at or just above the crosshair. Same compositing
-  // compensation as `lowOpacity` above.
   const highOpacity = useDerivedValue(() => {
     const c = crossfadeSV.value
     return c.highIndex === NO_HIGHLIGHT_INDEX
--- a/packages/k2-alpine/src/components/Chart/PriceChart/VolumeRow.tsx
+++ b/packages/k2-alpine/src/components/Chart/PriceChart/VolumeRow.tsx
@@ -1,30 +1,51 @@
-import { Canvas, RoundedRect } from '@shopify/react-native-skia'
-import React, { FC, useMemo } from 'react'
+import { Canvas, Path, RoundedRect, Skia } from '@shopify/react-native-skia'
+import React, { FC, useEffect, useMemo } from 'react'
 import { View } from 'react-native'
 import {
-  makeMutable,
   SharedValue,
-  useAnimatedReaction
+  useAnimatedReaction,
+  useDerivedValue,
+  useSharedValue
 } from 'react-native-reanimated'
 import { useTheme } from '../../../hooks'
 import { CHART_INSET } from './constants'
-import { indexToX } from './helpers'
+import {
+  crossfadeRectOpacity,
+  IDLE_VOLUME_CROSSFADE,
+  indexToX,
+  NO_HIGHLIGHT_INDEX,
+  VOLUME_IDLE_OPACITY,
+  volumeCrosshairWeights
+} from './helpers'
 import { OhlcCandle } from './types'
 
 type Props = {
   candles: OhlcCandle[]
   width: number
   height: number
-  /** When provided, each bar's opacity tracks the crosshair X — closest bars
-   * pop to full opacity, falling off linearly within one candle's distance. */
+  /** When provided, the two candles nearest the crosshair X crossfade —
+   * the closer one pops toward full opacity, the other falls back toward
+   * idle, linearly within one candle's distance. */
   crosshairX?: SharedValue<number>
   isActive?: SharedValue<boolean>
 }
 
 const BAR_WIDTH_RATIO = 0.6
-const IDLE_OPACITY = 0.1
-const ACTIVE_OPACITY = 1
 
+type BarGeom = { x: number; barHeight: number; radius: number }
+
+/**
+ * Bar highlight: one static `SkPath` for idle bars + two dynamic
+ * `RoundedRect`s for the two nearest bars, driven by a single `SharedValue`
+ * holding both indices and their crossfade weights (was: one `makeMutable`
+ * per candle). Crossfades between the two nearest candles exactly like
+ * main's original per-candle implementation — see `volumeCrosshairWeights`
+ * in `helpers.ts` for the interpolation this restores. Because the
+ * highlight rects paint on top of the idle `Path` (which already painted
+ * those same bars), their opacity is compositing-compensated via
+ * `crossfadeRectOpacity` so `idlePath + rect` composes to main's per-bar
+ * target opacity exactly, not a double-counted overshoot. CP-14918 item 2.
+ */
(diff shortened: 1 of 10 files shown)


packages/k2-alpine/src/components/Chart/PriceChart/helpers.ts
```diff
@@ -1,6 +1,7 @@
 import type { Skia } from '@shopify/react-native-skia'
 import { PriceChangeStatus } from '../../PriceChangeIndicator/types'
 import {
+  CHART_INSET,
```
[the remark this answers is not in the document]
B0Y3R-AVA: Taken as a tidy-up, but the error claim here is not accurate — flagging that so the reasoning is on the record.

`typeof Skia.Path.Make` inside a type alias is a *type query*, i.e. a type position, where `import type` bindings are legal. The "cannot be used as a value" error only fires in value positions (`extends Skia`, `new Skia()`). Two independent confirmations:

- These two lines are pre-existing on `main` (`helpers.ts:1,9`), untouched by this PR — they surfaced here only because they sit adjacent to a diff hunk. Root `yarn tsc` covers all packages in the `Test` job and passes on both `main` and this PR head.
- Compiled the pattern directly against the installed skia types: `ReturnType<typeof Skia.Path.Make>` and the directly-imported `SkPath` are mutually assignable, no diagnostics.

That said, the suggestion is worth doing on its own merits — sibling `Series.tsx:5` in this same directory already imports `SkPath` from the package, so the local alias was a pointless inconsistency. Done in 2e72b037f.

One deliberate deviation: I used `import type { SkPath }` rather than the inline `import { type SkPath }` form `Series.tsx` uses. `helpers.ts` is pure logic, Jest-tested without skia native mocks, and the standalone `import type` form is guaranteed fully elided — the inline-modifier form can emit a bare `import {} from ...` under `verbatimModuleSyntax`, which would pull skia in at runtime.

packages/k2-alpine/src/components/Chart/PriceChart/helpers.ts
```diff
@@ -64,19 +176,89 @@ export const rangeBounds = (
...
+/** en-US `{ weekday: 'short', month: 'short', day: 'numeric', year:
+ * 'numeric' }` — e.g. "Wed, Apr 29, 2026". */
+const formatWeekdayMonthDayYear = (d: Date): string => {
+  assertValidDate(d)
+  return `${WEEKDAY_ABBR[d.getDay()]}, ${
+    MONTH_ABBR[d.getMonth()]
+  } ${d.getDate()}, ${d.getFullYear()}`
+}
+
 /** e.g. "Last update: Wed, Apr 29, 2026 at 9:41 AM" */
```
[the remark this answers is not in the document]
B0Y3R-AVA: Correct, and the accidental-cleanup risk is the real point — the `AM_PM_SEPARATOR` constant carries a "do not clean up" warning but the example two hunks down quietly contradicted it. Fixed in 2e72b037f:

```ts
/** e.g. "Last update: Wed, Apr 29, 2026 at 9:41\u202fAM" — the AM/PM
 * separator is U+202F (narrow no-break space), not an ASCII space; see
 * `AM_PM_SEPARATOR`. */
```

Spelled as the visible `\u202f` escape rather than embedding a literal narrow no-break space, matching the convention already used in the `formatHour12Minute` doc above it. An invisible NNBSP byte sitting in a comment would have been a worse trap than the wrong example was.

_(Edited: the first version of this reply had its `\u202f` collapsed into an actual narrow no-break space in transit, which made the snippet misstate what landed.)_

packages/k2-alpine/src/components/Chart/PriceChart/helpers.ts
```diff
@@ -64,19 +174,91 @@ export const rangeBounds = (
...
-    weekday: 'short',
-    month: 'short',
-    day: 'numeric',
-    year: 'numeric'
-  })
-  const timePart = d.toLocaleTimeString(undefined, {
-    hour: 'numeric',
-    minute: '2-digit'
-  })
+  const datePart = formatWeekdayMonthDayYear(d)
```
gergelylovas, line 213: Whats the reason for this change? What does it improve? It's not apparent form the comments.

packages/k2-alpine/src/components/Chart/PriceChart/VolumeRow.tsx
```diff
@@ -34,84 +55,159 @@ export const VolumeRow: FC<Props> = ({
...
+        path.addRRect(
+          Skia.RRectXY(Skia.XYWHRect(x, y, barWidth, barHeight), radius, radius)
+        )
+      })
+    }
+    return { barGeom: geom, idlePath: path }
+  }, [candles, innerWidth, height, barWidth, maxVolume])
+
+  // Mirrored into a SharedValue via an effect rather than closed over
+  // directly inside the worklets below.
```
gergelylovas: Right, but why? This comment would be a lot more useful if it told why it's this way not describe the what

gergelylovas (changes requested):
In general claude was quite wordy in the comments explaining the what again instead of the why. I think comments need a bit of tweaking and be more concise and useful. 

Also did dev testing. The chart works fine for all timespans except 1D. See screenhot:

Image

packages/k2-alpine/src/components/Chart/PriceChart/helpers.ts
B0Y3R-AVA: The why: every `Intl`/`toLocale*` `.format()` call crosses Hermes -> JSI -> native ICU, which costs multiple ms per call on device, and a chart mount was making ~96 of those calls synchronously before first paint. Deferring the full-array formatting handles most of it, but `formatLastUpdate` and `formatActiveTime` still run on the idle render, so they had to become cheap too. Hand-rolling keeps the whole thing in Hermes; the trade-off is hardcoded en-US output.

The justification existed but sat ~80 lines above this hunk, so it was invisible here. Fixed in d9a8632df: the reason now lives once at `MONTH_ABBR` and each formatter points at it.

packages/k2-alpine/src/components/Chart/PriceChart/VolumeRow.tsx
B0Y3R-AVA: Reworded in d9a8632df to carry the why:

```ts
// Worklets capture by value: closing over `barGeom` would copy all N geoms
// into the UI runtime once per derived value, per render. One SharedValue
// copies once per geometry change — written in an effect, not during
// render, because Reanimated doesn't support SharedValue writes mid-render.
```

B0Y3R-AVA:
Ran down the 1D breakage. Root cause is data-side, not the Skia rewrite, and it reproduces on `main` too: the rendering math in this PR is a line-for-line copy of the per-candle geometry it replaced.

`useTokenChartCandles` sends 1D (alone) to the `market_chart/range` endpoint expecting 5-minutely points, but that endpoint returns hourly granularity on our plan, so a 24h window yields ~25 points against 48 buckets. One point per candle means every candle came out `open == close == high == low`: 1px flat bodies (the `Math.max(1, ...)` floor), zero-length wicks, and all green since `close >= open`. Your screenshot confirms it, since the 8pt body-width cap only engages below ~26 candles, and "Last update 8:00 PM" on a 21:01 clock is the hourly signature.

Fixed in f0e83949c:
- 1D now fetches `market_chart?days=1`, which does return 5-minutely data (~288 points, 6 per candle). This also makes the candle fetch consistent with `TrackTokenDetailScreen`'s existing `1D -> days: 1` mapping for the header delta.
- `bucketPointsIntoCandles` now emits fewer multi-point candles instead of N one-point flat ones when upstream granularity is coarser than the bucket count, so a coarse response degrades to real candles rather than dashes.
- The old test that asserted the degenerate 48-flat-candle output is rewritten, with new coverage for hourly-shaped responses and endpoint selection.

On the comments: d9a8632df is a full pass over the files in this PR. Restatements of the code are cut, internal doc-section references are stripped, the perf measurement is stated once (at `MONTH_ABBR`) and referenced elsewhere, and the two spots you flagged now answer the why in the code itself. Net: ~130 comment lines removed.

Would appreciate a re-test of 1D on device when you get a chance.

gergelylovas (approved):

PR #4033, opened 2026-08-11
CP-14934: Fix buy-flow lag: stop running the 57k-token pipeline to resolve individual Meld tokens

B0Y3R-AVA:
## Description

**Ticket: [CP-14934]**

Fixes serious tap-to-response lag throughout the Meld buy flow — every button press stalled while navigating Buy → token select → amount.

**Root cause** (found by profiling a dev build on a physical iPhone 14 with render/JS-stall instrumentation): the C-Chain contract-token list is ~57k tokens, and the buy screens pushed all of it through `useSearchableTokenList`'s map → merge → filter → sort pipeline to resolve one or two specific tokens:

- `SelectToken` ran the pipeline just to locate AVAX + USDC (~340ms per commit; the pipeline's zero-balance stubs could never even match the USDC lookup, since `internalId` only exists on balance-data tokens)
- `SelectAmount` ran it **three times over** — `useSelectAmount`, `useFiatSourceAmount`, and `useCreateCryptoQuote` each chain into `useMeldTokenWithBalance`, which spun up two pipeline instances each (~700ms of hook time per render, commits of 1.0–1.8s)
- The Meld token-list screen indexed all 57k tokens into latched zustand maps so ~50 Meld-supported currencies could look themselves up
- Balance polling re-triggers all of it: `useAccountBalances` refetches every 5s in prod (30s in dev) and emits one `setQueryData` per network per poll, so these trees re-rendered near-constantly and taps queued behind multi-hundred-ms commits

**Changes**

1. Memoize `useErc20ContractTokens` (unmemoized `.filter()` broke memo identity for every consumer downstream)
2. `SelectToken` resolves AVAX/USDC from the account's held tokens (~24) directly — provably equivalent matching, since the old lookup could only ever match balance-data tokens
3. `useMeldTokenWithBalance` searches held tokens first, then falls back to a cheap prefiltered scan of the contract list, stubbing only the match (Solana deliberately stays restricted to USDC_SOLANA — widening SPL support is a product decision, not a perf fix)
4. New `useMeldSupportedCryptoCurrencies` joins Meld's ~50 supported currencies against small held-token maps (contract-token map memoized on the stable query arrays, so balance ticks only rebuild the ~24-entry held maps). This replaces and deletes the `tokenIndex`/`supportedCryptoCurrencies` latch stores, `useResetMeldTokenList`, and `useSearchableERC20AndSolanaTokenList`. Side benefit: the token list's balances now live-update instead of freezing per visit.

**Measured on device** (iPhone 14, dev build, React `<Profiler>` + JS event-loop stall detector):

| Metric | Before | After |
| --- | --- | --- |
| Token-select screen commit | ~340ms | 5–8ms |
| Amount screen commit | 950–1,700ms | ~17ms |
| Idle JS stalls on amount screen | 1.4–6.5s, continuous | rare, small |

No new dependencies. No breaking changes.

**Review notes:** two adversarial review passes were run against the real type definitions. Round 2 caught a genuine bug that is fixed here: held SPL balance tokens carry only `networkChainId` (the balance mapper deliberately omits `chainId` on the SPL branch), so they need dedicated map keying — otherwise held Solana USDC vanishes from the offramp list and shows as zero-balance on the onramp list. Also noted: pre-existing (untouched) `isLastItem` separator misplacement in `TokenList` during search.

**Follow-ups (separate tickets):** batch `useAccountBalances`' per-network `setQueryData` bursts and revisit the 5s prod refetch interval (app-wide win); single-pass filtering + optional sort in `useSearchableTokenList` for the portfolio consumers.

## Screenshots/Videos

N/A — no visual changes. Perf before/after measurements in the table above.

## Testing

iOS: 9458
Android: 9459

Happy path:
1. Portfolio → Buy → "What token do you want to buy?" — screen should open without stalling; AVAX and USDC rows show (USDC requires a balance-data entry, as before)
2. Select AVAX → amount screen opens instantly, token row shows AVAX with logo
3. Type an amount → quotes fetch, provider appears, Buy proceeds to the Meld widget
4. Back on token select → "Select other token" → full list renders fast: AVAX first, USDC second, rest sorted by balance; held tokens show balances; search filters correctly; selecting a token lands on the amount screen with it selected

Edge cases:
- Offramp (withdraw) flow: token list shows only tokens with balance > 0
- Account with Solana USDC balance: it appears with its real balance in the onramp list (SPL keying fix)
- Blacklisted (hidden) tokens and disabled networks stay excluded from both flows

Regression-test the full buy and withdraw flows end to end (token selection, amount entry, payment method, provider selection, widget launch) across an account with balances on multiple networks. Expected: identical behavior to before, with no multi-second freezes between taps anywhere in the flow.

## Checklist

[CP-14934]: https://ava-labs.atlassian.net/browse/CP-14934?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/core-mobile/app/new/features/meld/hooks/useMeldTokenWithBalance.ts
```diff
@@ -14,20 +49,73 @@ export const useMeldTokenWithBalance = ({
...
+    // Cheap prefilter covering every isTokenTradable arm (address match for
+    // erc20/SPL, symbol match for BTC/SOL) so we only build stubs for a
+    // handful of candidates instead of all ~57k.
+    const meldAddress = meldToken.contractAddress?.toLowerCase()
+    const candidates = [...erc20ContractTokens, ...usdcSolanaTokens].filter(
+      tk =>
+        tk.address?.toLowerCase() === meldAddress ||
+        tk.symbol === 'BTC' ||
+        tk.symbol === 'SOL'
+    )
```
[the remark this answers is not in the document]
B0Y3R-AVA: Tightened the prefilter so an undefined address can't match:

```ts
(meldAddress !== undefined && tk.address.toLowerCase() === meldAddress) ||
```

On the crash risk itself: `NetworkContractToken` is a union of `ERC20Token`, `SPLToken`, `ERC1155Token`, `ERC721Token` and `NONERCToken`, and all five declare `address: string` as required, so a token with a missing address can't come out of `erc20ContractTokens` or `solanaTokens`. The non-optional `token.address.toLowerCase()` in `isSupportedToken` also predates this PR. So it wasn't reachable, but the `undefined === undefined` match was sloppy regardless and the guard costs nothing.

Behavior is unchanged for the real case. When a Meld currency has no `contractAddress`, all three address-based arms of `isTokenTradable` already return false, leaving only the symbol-based BTC/SOL arms, which the prefilter still covers.

packages/core-mobile/app/new/features/meld/hooks/useMeldTokenWithBalance.ts
```diff
@@ -14,20 +49,74 @@ export const useMeldTokenWithBalance = ({
...
+    // Cheap prefilter covering every isTokenTradable arm (address match for
+    // erc20/SPL, symbol match for BTC/SOL) so we only build stubs for a
+    // handful of candidates instead of all ~57k.
+    const meldAddress = meldToken.contractAddress?.toLowerCase()
+    const candidates = [...erc20ContractTokens, ...usdcSolanaTokens].filter(
+      tk =>
+        (meldAddress !== undefined &&
+          tk.address.toLowerCase() === meldAddress) ||
+        tk.symbol === 'BTC' ||
+        tk.symbol === 'SOL'
```
meeh0w: @copilot Is the symbol comparison safe here?

packages/core-mobile/app/new/features/meld/hooks/useMeldTokenWithBalance.ts
```diff
@@ -14,20 +49,74 @@ export const useMeldTokenWithBalance = ({
...
+
+    // Solana stubs are deliberately restricted to USDC_SOLANA — widening SPL
+    // support is a product decision, not a perf fix.
+    const usdcSolanaTokens = isSolanaSupportBlocked
+      ? []
+      : solanaTokens.filter(
+          tk =>
+            'chainId' in tk &&
+            tk.chainId === ChainId.SOLANA_MAINNET_ID &&
+            tk.address === tokenAddresses.USDC_SOLANA
```
gergelylovas: This feels very inefficient. We are fetching the full token list for solana from the network and then looping through the whole list to filter out everything thats not USDC.

It should be a `find` at the very least. However having a simple hook that makes a single request to `/v1/token/lookup` with all of the missing token addresses, you could get all token information in a single step without any filtering including the the one on the `erc20ContractTokens` list.

packages/core-mobile/app/new/features/meld/hooks/useMeldSupportedCryptoCurrencies.ts
```diff
@@ -0,0 +1,216 @@
...
+          'chainId' in token &&
+          token.chainId === ChainId.SOLANA_MAINNET_ID &&
+          token.address === tokenAddresses.USDC_SOLANA
+        ) {
+          map.set(
+            `${ChainId.SOLANA_MAINNET_ID}-${token.address.toLowerCase()}`,
+            token
+          )
+        }
+      }
```
gergelylovas: Same here about looping through 13668 tokens to find a single one.

packages/core-mobile/app/new/features/meld/hooks/useMeldSupportedCryptoCurrencies.ts
```diff
@@ -0,0 +1,216 @@
...
+import { useErc20ContractTokens } from 'common/hooks/useErc20ContractTokens'
+import { useSolanaTokens } from 'common/hooks/useSolanaTokens'
+import { useTokensWithBalanceForAccount } from 'features/portfolio/hooks/useTokensWithBalanceForAccount'
+import {
+  NATIVE_ERC20_TOKEN_CONTRACT_ADDRESS,
+  SOLANA_MELD_CHAIN_ID,
+  ServiceProviderCategories
+} from '../consts'
+import { CryptoCurrency, CryptoCurrencyWithBalance } from '../types'
+import { asZeroBalanceToken } from './useMeldTokenWithBalance'
```
gergelylovas: Please extract this to a util so we don't have to import across hooks

packages/core-mobile/app/new/features/meld/hooks/useMeldSupportedCryptoCurrencies.ts
```diff
@@ -0,0 +1,216 @@
...
+}
+
+const passesListFilters = (
+  token: LocalTokenWithBalance,
+  { includeZeroBalance, tokenVisibility, enabledChainIds }: ListFilterOptions
+): boolean =>
+  (includeZeroBalance || token.balance > 0n) &&
+  isTokenVisible(tokenVisibility, token) &&
+  token.type !== TokenType.ERC1155 &&
+  token.type !== TokenType.ERC721 &&
```
gergelylovas: Should we filter what it should be instead? This can break if we add a new token type like in case of hyperliqid

gergelylovas (changes requested):

packages/core-mobile/app/new/features/meld/hooks/useMeldTokenWithBalance.ts
B0Y3R-AVA: Both scans are gone. There's a new `useMeldContractTokenMap` hook that indexes the contract tokens once by `{chainId}-{address}`, and resolution here is now a single map hit keyed off the Meld currency instead of a pass over the catalog. The Solana side is a `find` for USDC rather than a filter over all of them.

On `/v1/token/lookup`: that's CP-14936, already in flight. #4035 moves the other consumers onto aggregator lookups, and the branch stacked after it replaces this Meld picker with the v2 tokens API and drops the full-catalog fetch entirely. I'd rather not restack this PR behind that one, since the numbers here are already measured against the current data path.

One behavior change worth flagging: the old candidate scan also matched contract tokens by symbol for BTC/SOL, so in principle it could have stubbed an unrelated ERC20 named `BTC`. The map path drops that arm. BTC and SOL only ever reach this hook when the account has the native token in balance data, which is also the only way the list hook surfaces them, so they resolve on the held-token path before the fallback runs.

packages/core-mobile/app/new/features/meld/hooks/useMeldSupportedCryptoCurrencies.ts
B0Y3R-AVA: Same fix. Both hooks now go through the shared `useMeldContractTokenMap`, and the Solana pass inside it is a `find` for USDC instead of a loop over all ~13.7k. The map is cached at module scope keyed on the query array references, so the three `useMeldTokenWithBalance` consumers the amount screen mounts share one build rather than each retaining a separate 57k-entry map.

packages/core-mobile/app/new/features/meld/hooks/useMeldSupportedCryptoCurrencies.ts
B0Y3R-AVA: Moved. `asZeroBalanceToken` now lives in `features/meld/utils.ts`, along with the `passesListFilters` copy that was duplicated across both hooks and the `meldContractTokenKey` / `meldCurrencyTokenKey` helpers. The hooks no longer import from each other.

packages/core-mobile/app/new/features/meld/hooks/useMeldSupportedCryptoCurrencies.ts
B0Y3R-AVA: Good call, switched to an allowlist. `MELD_LISTABLE_TOKEN_TYPES = [NATIVE, ERC20, SPL]` in `utils.ts`, so `HYPERCORE_SPOT` and `NONERC` fall out by default instead of each new type needing its own exclusion.

bohdan-struk-avalabs (dismissed):

gergelylovas (dismissed):

packages/core-mobile/app/new/features/meld/components/SelectAmount.tsx
```diff
@@ -19,7 +19,6 @@ import React, { useCallback, useEffect, useState } from 'react'
 import { useSelector } from 'react-redux'
 import { selectSelectedCurrency } from 'store/settings/currency'
 import { ServiceProviderCategories } from '../consts'
-import { useResetMeldTokenList } from '../hooks/useResetMeldTokenList'
 import { useSelectAmount } from '../hooks/useSelectAmount'
 import { useOfframpActivityIndicator, useOfframpSessionId } from '../store'
```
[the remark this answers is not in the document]
B0Y3R-AVA: Good catch, and it is what CI was failing on. The stale `jest.mock("../hooks/useResetMeldTokenList")` is removed in 32f349c — the component no longer touches that hook, so the mock had no purpose beyond breaking module resolution. `app/new/features/meld` tests pass locally.

gergelylovas (approved):

PR #4035, opened 2026-08-12
CP-14936: Replace full token-list consumers with aggregator lookups

B0Y3R-AVA:
## Description

**Ticket: [CP-14936](https://ava-labs.atlassian.net/browse/CP-14936)**

Follow-up to CP-14934 (#4033): that PR fixed the buy-flow lag symptom, this one attacks the root cause, mobile holding the entire ~57k C-Chain token list in memory. Per the direct-client decision recorded on the ticket (no vm-modules change; the extension audit showed zero `module.getTokens()` consumers there).

> **Rebased onto main on 19 Aug.** #4033 landed in the meantime and had already rewritten the Meld token machinery, so two things changed shape versus the original diff. See "Rebase notes" below before re-reviewing.

**1. Remove dead full-list work on hot screens.** `PortfolioScreen`, `useAssetsFilterAndSort`, `SelectSendTokenScreen` and `useActivityFilterAndSearch` still ran the full `useErc20ContractTokens()` fetch plus an unmemoized `.filter(isErc20)` per render, even though `useSearchableTokenList`'s `hideZeroBalance` default (CP-14918) discards the result internally. They now call the hook with no contract tokens, mirroring `TokenManagementScreen`. Provably output-identical.

**2. Replace full-list exact-address lookups with `POST /v1/token/lookup`.**

- `useAddCustomToken`: "already exists" now checks custom tokens, held tokens and a single-address aggregator lookup. Previously it built a membership Set from the whole network catalog, rebuilt per keystroke. Validation holds a loading state while the lookup is in flight so there is no false-"valid" flash.
- `RecurringSchedulesScreen`: resolves visible rows' `tokenIn`/`tokenOut` via batched lookups instead of `useNetworkContractTokens`.
- `ActivityService.resolveUnknownTokenSymbols`: looks up just the batch's unknown SPL addresses instead of fetching the entire SPL list per activity fetch.
- Meld `SelectToken`: keeps #4033's balance-based row resolution and adds a single-token lookup so the USDC row still renders for an account that does not hold USDC. Also pins both row lookups to the active C-Chain (details below).

New shared `tokenLookupKey` helper in `common/utils/tokenLookup.ts` encodes the lookup response-map key convention, **verified against the live endpoint on-device**: the server lowercases EVM (`eip155:`) addresses but preserves case for Solana caip2Ids and base58 addresses. A naive lowercase match silently never resolves Solana tokens, and there is a regression test for exactly that.

### Rebase notes

Two conflicts against #4033 were semantic rather than textual, so they are worth a look:

- **`useSearchableERC20AndSolanaTokenList` is gone.** #4033 deleted the hook outright and left zero consumers, so the rebase took the deletion. The USDC-Solana single-token lookup that this PR originally added there is therefore **no longer part of this diff**; the SPL side gets rebuilt in #4036 instead.
- **`useSolanaTokens` survives after all.** This PR originally deleted it. #4033 added a new consumer, `useMeldContractTokenMap`, so deleting it now breaks the build and the rebase restored it. Same story for `useErc20ContractTokens`. Both are retired in #4036.
- **`SelectToken` is the union of both approaches.** #4033's `visibleTokens` filtering (token visibility, enabled chain ids, NFT exclusion, offramp zero-balance) is kept as-is. On top of it: the unheld-USDC aggregator lookup, plus an explicit `networkChainId === cChainNetwork?.chainId` guard on both the USDC and AVAX finds. The USDC guard matters because `tokenIds.USDC` is a cross-chain internalId, and the AVAX guard matters because AVAX is the native token of X- and P-Chain too, so a symbol-only match could resolve an X/P balance for a C-Chain row.

Deferred, tracked on the ticket: `useGetCChainToken` (DeFi market) still uses the full list and is migrated in #4037; retiring `getNetworkContractTokens`/`module.getTokens()` entirely is #4036.

Follow-ups filed off this review: **CP-14953** (consolidate the many token data shapes) and **CP-14956** (retire `NetworkService.fetchDeBankNetworks` or backfill `caip2Id` on the DeBank-sourced networks).

## Screenshots/Videos

JS-only change, no UI redesign. Android verified end-to-end on a physical Pixel (see Testing); iOS not separately exercised beyond unit and tsc coverage.

## Testing

**Dev Testing**

iOS: 9467
Android: 9468

> Those builds **predate the rebase**. The Meld `SelectToken` path in particular now sits on top of #4033's rewrite, so it needs a fresh build and a re-run of the buy-flow steps below before merge.

Performed on a physical Pixel (internal debug build, staging aggregator, ephemeral test wallet):

- Portfolio, Assets filter/sort, Send token select and search, Activity tab: all render correctly with the contract-token fetch removed.
- Meld Buy, token select: AVAX and USDC rows both present, USDC via the new lookup even when unheld.
- Add custom token: catalog token (JOE `0x6e84...fDd` on C-Chain) rejected with "Token already exists in the wallet", no stuck spinner (verified with instrumented logs: lookup hit, then `AlreadyExists`); malformed address gives an inline error and no spinner.
- Aggregator coverage probed live from the app (AppCheck auth): `/v2/tokens` and `/v1/token/lookup` both resolve Fuji (`eip155:43113`) assets, WAVAX by exact address, and mainnet returns the full 52,666-token catalog. Solana lookup key casing verified as described above.
- Pre-existing, not introduced here, reproduced on a `main` baseline: two silent `TypeError: Cannot convert undefined value to object` logs on an empty Solana Activity tab.

Post-rebase verification: real `tsc -p .` clean (not the yarn-wrapper no-op), eslint `--max-warnings=0` clean on every changed file, and the touched unit suites green (`tokenLookup`, `useTokenLookup`, `ActivityService`, meld) at 35 tests across 5 suites.

Edge cases: lookup failure degrades gracefully (screens fall back to balance-cache data or omit the optional row; add-custom-token fails open exactly like the old path did when its catalog fetch failed). Empty lookup results are guarded at every call site.

**QA Testing**

- Portfolio, Send, Activity and Token Management should be visually unchanged.
- Buy (Meld): token select must list AVAX and USDC on C-Chain regardless of held balance. Worth checking an account holding USDC on a non-Avalanche chain only, since that previously could satisfy the C-Chain row.
- Withdraw (Meld, offramp): AVAX row should appear only with a C-Chain AVAX balance, not an X- or P-Chain one.
- Add custom token: any token already in the Avalanche catalog (e.g. JOE) must be rejected as existing; a genuinely new ERC-20 address must resolve and add.
- Solana account with SPL activity: token symbols in Activity must resolve, not show "Unknown".
- Regression watch: memory and perf on Portfolio mount should improve, since there is no 57k-token fetch or filter.

## Checklist

[CP-14936]: https://ava-labs.atlassian.net/browse/CP-14936?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core-mobile/app/new/common/hooks/useAddCustomToken.ts
+++ b/packages/core-mobile/app/new/common/hooks/useAddCustomToken.ts
@@ -44,7 +44,12 @@ const validateAddress = (
     return AddressValidationStatus.Invalid
   }
 
-  if (tokens.some(token => token === tokenAddress) || existsInNetworkCatalog) {
+  if (
+    // Custom tokens are stored lowercased (store/customToken slice), while
+    // user input may be checksummed -- compare case-insensitively.
+    tokens.some(token => token.toLowerCase() === tokenAddress.toLowerCase()) ||
+    existsInNetworkCatalog
+  ) {
     return AddressValidationStatus.AlreadyExists
   }
 
--- a/packages/core-mobile/app/new/common/hooks/useAddCustomToken.ts
+++ b/packages/core-mobile/app/new/common/hooks/useAddCustomToken.ts
@@ -1,6 +1,9 @@
 import { useDispatch, useSelector } from 'react-redux'
 import { isAddress } from 'ethers'
-import { addCustomToken as addCustomTokenAction } from 'store/customToken'
+import {
+  addCustomToken as addCustomTokenAction,
+  selectAllCustomTokens
+} from 'store/customToken'
 import { useState, useEffect, useMemo, useCallback } from 'react'
 import { Network } from '@avalabs/core-chains-sdk'
 import Logger from 'utils/Logger'
@@ -11,13 +14,15 @@ import {
   TokenType,
   TokenWithBalanceERC20
 } from '@avalabs/vm-module-types'
-import { useNetworkContractTokens } from 'hooks/networks/useNetworkContractTokens'
 import {
   useSelectedNetwork,
   useTokenAddress
 } from 'features/tokenManagement/store'
 import { useTokensWithBalanceByNetworkForAccount } from 'features/portfolio/hooks/useTokensWithBalanceByNetworkForAccount'
 import { selectActiveAccount } from 'store/account'
+import { selectIsDeveloperMode } from 'store/settings/advanced'
+import { tokenLookupKey, useTokenLookup } from 'common/hooks/useTokenLookup'
+import { getCaip2ChainId } from 'utils/caip2ChainIds'
 
 enum AddressValidationStatus {
   Valid,
@@ -28,7 +33,8 @@ enum AddressValidationStatus {
 
 const validateAddress = (
   tokenAddress: string,
-  tokens: string[]
+  tokens: string[],
+  existsInNetworkCatalog: boolean
 ): AddressValidationStatus => {
   if (tokenAddress.length <= 10) {
     return AddressValidationStatus.TooShort
@@ -38,7 +44,7 @@ const validateAddress = (
     return AddressValidationStatus.Invalid
   }
 
-  if (tokens.some(token => token === tokenAddress)) {
+  if (tokens.some(token => token === tokenAddress) || existsInNetworkCatalog) {
     return AddressValidationStatus.AlreadyExists
   }
 
@@ -79,17 +85,29 @@ const useAddCustomToken = (callback: () => void): CustomToken => {
   const [selectedNetwork] = useSelectedNetwork()
   const chainId = selectedNetwork?.chainId
 
-  const tokens = useNetworkContractTokens(selectedNetwork)
+  const isDeveloperMode = useSelector(selectIsDeveloperMode)
+  const allCustomTokens = useSelector(selectAllCustomTokens)
   const activeAccount = useSelector(selectActiveAccount)
   const { tokens: tokensWithBalance } = useTokensWithBalanceByNetworkForAccount(
     activeAccount,
     chainId
   )
 
+  // A token the user already added or already holds counts as "already in
+  // the wallet" -- no need to pull the network's full contract-token catalog
+  // just to build this membership set.
+  const customTokensForChain = useMemo(
+    () =>
+      chainId !== undefined && selectedNetwork?.isTestnet === isDeveloperMode
+        ? allCustomTokens[chainId] ?? []
+        : [],
+    [allCustomTokens, chainId, isDeveloperMode, selectedNetwork?.isTestnet]
+  )
+
   const tokenAddresses = useMemo(
     () => [
       ...new Set([
-        ...tokens.map(t => t.address),
+        ...customTokensForChain.map(t => t.address),
         ...tokensWithBalance
           .map(t => {
             if (t.type === TokenType.ERC20) {
@@ -99,7 +117,36 @@ const useAddCustomToken = (callback: () => void): CustomToken => {
           .filter(item => item !== undefined)
       ])
     ],
-    [tokens, tokensWithBalance]
+    [customTokensForChain, tokensWithBalance]
+  )
+
+  const selectedCaip2Id = useMemo(
+    () => (chainId !== undefined ? getCaip2ChainId(chainId) : undefined),
+    [chainId]
+  )
+
+  // Restores the old useNetworkContractTokens "already exists" check without
+  // pulling the network's whole curated token list: look up just the single
+  // address the user typed against the token aggregator.
+  const networkCatalogLookupIds = useMemo(
+    () =>
+      selectedCaip2Id && isAddress(tokenAddress)
+        ? [{ caip2Id: selectedCaip2Id, address: tokenAddress }]
+        : [],
+    [selectedCaip2Id, tokenAddress]
+  )
+
+  const { data: networkCatalogLookup, isLoading: isNetworkCatalogLoading } =
+    useTokenLookup(networkCatalogLookupIds)
+
+  const existsInNetworkCatalog = useMemo(
+    () =>
+      selectedCaip2Id !== undefined &&
+      isAddress(tokenAddress) &&
+      Boolean(
+        networkCatalogLookup[tokenLookupKey(selectedCaip2Id, tokenAddress)]
+      ),
+    [networkCatalogLookup, selectedCaip2Id, tokenAddress]
   )
 
   const existingToken = useMemo(
@@ -132,7 +179,26 @@ const useAddCustomToken = (callback: () => void): CustomToken => {
       return
     }
 
-    const validationStatus = validateAddress(tokenAddress, tokenAddresses)
+    // Hold off on validating while the network-catalog lookup for this exact
+    // address is still in flight -- otherwise a token that turns out to
+    // already be in the network's catalog could flash a "valid" verdict (and
+    // kick off fetchTokenData) before the lookup resolves.
+    if (isAddress(tokenAddress) && isNetworkCatalogLoading) {
+      setIsLoading(true)
+      return
+    }
+
+    const validationStatus = validateAddress(
+      tokenAddress,
+      tokenAddresses,
+      existsInNetworkCatalog
+    )
+
+    // Unconditional: only the Valid branch below re-arms this via its own
+    // fetchTokenData().finally() -- every other branch has no async work,
+    // so without this reset the gate above would leave it stuck on.
+    setIsLoading(false)
+
     switch (validationStatus) {
       case AddressValidationStatus.Invalid:
         setToken(undefined)
@@ -164,7 +230,13 @@ const useAddCustomToken = (callback: () => void): CustomToken => {
         setErrorMessage('')
         setToken(undefined)
     }
-  }, [selectedNetwork, tokenAddress, tokenAddresses])
+  }, [
+    selectedNetwork,
+    tokenAddress,
+    tokenAddresses,
+    isNetworkCatalogLoading,
+    existsInNetworkCatalog
+  ])
 
   const addCustomToken = useCallback((): void => {
     if (token && chainId) {
--- a/packages/core-mobile/app/new/common/hooks/useAddCustomToken.ts
+++ b/packages/core-mobile/app/new/common/hooks/useAddCustomToken.ts
@@ -21,7 +21,8 @@ import {
 import { useTokensWithBalanceByNetworkForAccount } from 'features/portfolio/hooks/useTokensWithBalanceByNetworkForAccount'
 import { selectActiveAccount } from 'store/account'
 import { selectIsDeveloperMode } from 'store/settings/advanced'
-import { tokenLookupKey, useTokenLookup } from 'common/hooks/useTokenLookup'
+import { useTokenLookup } from 'common/hooks/useTokenLookup'
+import { tokenLookupKey } from 'common/utils/tokenLookup'
 import { getCaip2ChainId } from 'utils/caip2ChainIds'
 
 enum AddressValidationStatus {
--- a/packages/core-mobile/app/new/common/hooks/useTokenLookup.ts
+++ b/packages/core-mobile/app/new/common/hooks/useTokenLookup.ts
@@ -14,15 +14,28 @@ export type { TokenInfo }
 
 const STALE_TIME = 60 * 1000 // 60 seconds
 
-// Builds the token-aggregator's `/v1/token/lookup` response-map key
-// (`data.data` is keyed as `{caip2Id}-{address}`). Confirmed against the
-// live endpoint: the server only lowercases EVM (`eip155:`) addresses --
-// Solana addresses are case-sensitive base58, so lowercasing them there
-// means the key never matches what the server returns.
-export function tokenLookupKey(caip2Id: string, address: string): string {
+// The server's canonical address casing: EVM (`eip155:`) addresses are
+// lowercased, Solana addresses are case-sensitive base58 and kept verbatim --
+// a lowercased Solana address returns no results (confirmed against the live
+// endpoint). Use this for both request bodies and response-map keys so the
+// two can never disagree.
+export function normalizeLookupAddress(
+  caip2Id: string,
+  address: string
+): string {
   return caip2Id.toLowerCase().startsWith('eip155:')
-    ? `${caip2Id.toLowerCase()}-${address.toLowerCase()}`
-    : `${caip2Id}-${address}`
+    ? address.toLowerCase()
+    : address
+}
+
+// Builds the token-aggregator's `/v1/token/lookup` response-map key
+// (`data.data` is keyed as `{caip2Id}-{address}`, address cased per
+// normalizeLookupAddress above).
+export function tokenLookupKey(caip2Id: string, address: string): string {
+  const normalizedCaip2Id = caip2Id.toLowerCase().startsWith('eip155:')
+    ? caip2Id.toLowerCase()
+    : caip2Id
+  return `${normalizedCaip2Id}-${normalizeLookupAddress(caip2Id, address)}`
 }
 
 export function useTokenLookup(
--- a/packages/core-mobile/app/new/common/hooks/useTokenLookup.ts
+++ b/packages/core-mobile/app/new/common/hooks/useTokenLookup.ts
@@ -14,6 +14,17 @@ export type { TokenInfo }
 
 const STALE_TIME = 60 * 1000 // 60 seconds
 
+// Builds the token-aggregator's `/v1/token/lookup` response-map key
+// (`data.data` is keyed as `{caip2Id}-{address}`). Confirmed against the
+// live endpoint: the server only lowercases EVM (`eip155:`) addresses --
+// Solana addresses are case-sensitive base58, so lowercasing them there
+// means the key never matches what the server returns.
+export function tokenLookupKey(caip2Id: string, address: string): string {
+  return caip2Id.toLowerCase().startsWith('eip155:')
+    ? `${caip2Id.toLowerCase()}-${address.toLowerCase()}`
+    : `${caip2Id}-${address}`
+}
+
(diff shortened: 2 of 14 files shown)


packages/core-mobile/app/new/common/hooks/useAddCustomToken.ts
```diff
@@ -38,7 +44,7 @@ const validateAddress = (
     return AddressValidationStatus.Invalid
   }
 
-  if (tokens.some(token => token === tokenAddress)) {
+  if (tokens.some(token => token === tokenAddress) || existsInNetworkCatalog) {
```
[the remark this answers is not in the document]
B0Y3R-AVA: Good catch — fixed in 9472130a3. Custom tokens are stored lowercased by the reducer, so the membership check now compares case-insensitively. (The gap predates this PR — the same comparison exists on main — but this function was rewritten here anyway, so fixed in place.)

packages/core-mobile/app/new/features/swap/hooks/useFusionTokenLookup.ts
```diff
@@ -38,10 +42,8 @@ const toLookupEntry = (
     ? { caip2Id, address: id.toLowerCase() }
     : { internalId: normalizeId(id) }
```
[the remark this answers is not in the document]
B0Y3R-AVA: Fixed in 9472130a3. Added a shared `normalizeLookupAddress` (EVM addresses lowercased, Solana base58 kept verbatim — matching the server's canonicalization, which we verified against the live endpoint: a lowercased Solana address returns no results). `toLookupEntry` and `tokenLookupKey` both delegate to it now, so request bodies and response-map keys can't disagree.

packages/core-mobile/app/new/features/recurringSwap/screens/RecurringSchedulesScreen.tsx
```diff
@@ -635,6 +637,26 @@ export function RecurringSchedulesScreen({
   // the placeholder can be vertically centered instead of pinned near the top.
   const isEmpty = manageableCount === 0
 
+  // Resolve every visible row's tokenIn/tokenOut via the token-aggregator
+  // lookup (one request per unique token, via useQueries) instead of
+  // pulling the active network's whole contract-token catalog.
```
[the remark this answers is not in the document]
B0Y3R-AVA: Deliberate trade-off rather than an oversight: per-unique-token queries share the React Query cache entry (`TOKEN_LOOKUP` + token key) with every other `useTokenLookup` consumer, so tokens already resolved elsewhere in the app cost nothing here and vice versa, and the unique-token count is bounded by the visible schedule rows (small N, 60s staleTime caps refetches across the 30s poll). A single batched call would bypass that shared cache and refetch the full set every time. If schedule counts grow, the plan is a batched variant that seeds the per-token cache entries — tracked as a follow-up on CP-14936.

packages/core-mobile/app/new/common/hooks/useSolanaTokens.ts
gergelylovas, line 1: Should `useErc20ContractTokens` be removed as well?

packages/core-mobile/app/services/activity/ActivityService.tsx
```diff
@@ -215,33 +219,45 @@ export class ActivityService {
...
   ): Promise<void> {
     try {
-      const moduleTokens = await module.getTokens(mapToVmNetwork(network))
-
-      for (const token of moduleTokens) {
-        if ('address' in token && 'symbol' in token && token.symbol) {
-          tokenMap.set(token.address, {
-            symbol: token.symbol,
-            name: 'name' in token && token.name ? token.name : token.symbol
+      const caip2Id = getSolanaCaip2ChainId(network.chainId)
```
gergelylovas: the network should already have a caip2id prop. Is this getter needed? Also if this function gets called with an EVM chain, it would return a wrong `solana:1` caip2 id.

packages/core-mobile/app/new/common/hooks/useSearchableERC20AndSolanaTokenList.ts
```diff
@@ -1,12 +1,25 @@
...
 import { useSelector } from 'react-redux'
 import { selectIsSolanaSupportBlocked } from 'store/posthog/slice'
 import { LocalTokenWithBalance } from 'store/balance/types'
-import { useSolanaTokens } from './useSolanaTokens'
+import { getSolanaCaip2ChainId } from 'utils/caip2ChainIds'
 import { useErc20ContractTokens } from './useErc20ContractTokens'
 import { useSearchableTokenList } from './useSearchableTokenList'
+import { tokenLookupKey, useTokenLookup } from './useTokenLookup'
+
+const SOLANA_CAIP2_ID = getSolanaCaip2ChainId(ChainId.SOLANA_MAINNET_ID)
```
gergelylovas: should this be just `SolanaCaip2ChainId.MAINNET`

packages/core-mobile/app/new/common/hooks/useSearchableERC20AndSolanaTokenList.ts
```diff
@@ -18,16 +31,26 @@ export const useSearchableERC20AndSolanaTokenList = (
...
 
-  const usdcSolanaToken = useMemo(() => {
-    return solanaTokens.find(
-      tk =>
-        'chainId' in tk &&
-        tk.chainId === ChainId.SOLANA_MAINNET_ID &&
-        tk.address === tokenAddresses.USDC_SOLANA
-    )
-  }, [solanaTokens])
+  const usdcSolanaToken = useMemo((): SPLToken | undefined => {
```
gergelylovas: The code below could use some cleanup as well. We are creating a token which we put in a single item array then filter it again. We probably can get rid of `filteredUsdcSolanaTokenList`  and `filteredErc20TokenList` as well

packages/core-mobile/app/new/common/hooks/useTokenLookup.ts
```diff
@@ -14,6 +14,30 @@ export type { TokenInfo }
...
+// Builds the token-aggregator's `/v1/token/lookup` response-map key
+// (`data.data` is keyed as `{caip2Id}-{address}`, address cased per
+// normalizeLookupAddress above).
+export function tokenLookupKey(caip2Id: string, address: string): string {
+  const normalizedCaip2Id = caip2Id.toLowerCase().startsWith('eip155:')
+    ? caip2Id.toLowerCase()
+    : caip2Id
+  return `${normalizedCaip2Id}-${normalizeLookupAddress(caip2Id, address)}`
+}
+
```
gergelylovas: Please move these to a util.

Also `tokenLookupKey` can be simplified to a single toLowerCase conversion on the whole thing without checking again for eip155 for the address.

packages/core-mobile/app/new/features/recurringSwap/screens/RecurringSchedulesScreen.tsx
```diff
@@ -635,6 +637,26 @@ export function RecurringSchedulesScreen({
...
+  const tokenLookupIds = useMemo(() => {
+    if (!activeNetwork) return []
+    const caip2Id = getCaip2ChainId(activeNetwork.chainId)
```
gergelylovas: should the activeNetwork have a caip2id by default?

packages/core-mobile/app/new/features/meld/components/SelectToken.tsx
```diff
@@ -38,29 +39,73 @@ export const SelectToken = ({
...
-      token.type === TokenType.NATIVE && token.symbol === TokenSymbol.AVAX
+  // Unlike AVAX, USDC-C's icon is rendered, so an unheld account still needs
+  // its symbol/logo -- fetch only this one token, not the full contract list.
+  const usdcLookupIds = useMemo(
+    () => (heldUsdcAvalancheToken ? [] : [{ internalId: tokenIds.USDC }]),
+    [heldUsdcAvalancheToken]
   )
+  const { data: usdcLookupTokens } = useTokenLookup(usdcLookupIds)
+
+  const usdcAvalancheToken = useMemo(():
```
gergelylovas, line 105: General comment that doesn't necessary to be done as part of this PR. Feel like Core Mobile uses MANY token data shapes which makes reuse more difficult and results in many mappings across the application.
Try to stick to one that has all the params.

gergelylovas (changes requested):

packages/core-mobile/app/new/common/hooks/useTokenLookup.ts
B0Y3R-AVA: Moved both into `common/utils/tokenLookup.ts` in 2a2ed16d, with its own test file.

On the simplification: took it for the EVM branch, so it is one `toLowerCase()` over the whole `${caip2Id}-${address}` string now.

```ts
const key = `${caip2Id}-${address}`
return isEvmCaip2Id(caip2Id) ? key.toLowerCase() : key
```

Cannot do it unconditionally though. Solana caip2 ids are base58 as well (`solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp`), and so are the mint addresses, so lowercasing that key gets you nothing back from the aggregator. Checked it against the live endpoint before writing this.

Also dropped the bare `'eip155:'` literal for `BlockchainNamespace.EIP155` while I was in there.

packages/core-mobile/app/new/common/hooks/useSearchableERC20AndSolanaTokenList.ts
B0Y3R-AVA: Yep. Done in 2a2ed16d, three spots in that file.

packages/core-mobile/app/new/common/hooks/useSearchableERC20AndSolanaTokenList.ts
B0Y3R-AVA: Did part of this. The two memos are one now, and the return stopped building a fresh array on every render, which also means `TokenList`'s `setSplTokenIndex` effect no longer re-fires constantly.

The `.find()` has to stay, and it is not doing what it looks like it is doing. `useSearchableTokenList` merges `tokensWithBalance`, meaning the account's entire balance list across every chain, into its output regardless of what you pass as `tokens` (useSearchableTokenList.ts:99-110). So `filteredTokenList` is not the single-item array coming back out. It is that array plus everything the account holds. Drop the find and every balance-bearing token leaks into `filteredSolanaTokenList`.

Same reason both `useSearchableTokenList` calls are still there rather than folding USDC-SOL into the first one: that puts an SPL token into the list `TokenList.tsx` hands to `setTokenIndex`, which is the ERC20/native index.

Added a comment above the find so nobody deletes it for the same reason.

packages/core-mobile/app/new/common/hooks/useSolanaTokens.ts
B0Y3R-AVA: Not yet. Two consumers left: `useSearchableERC20AndSolanaTokenList` (the ERC20 half of the Meld token list) and `useGetCChainToken` in defiMarket, which goes through `useAvalancheErc20ContractTokens`.

`useGetCChainToken` is the next thing I am migrating on CP-14936, and this hook comes out with it.

packages/core-mobile/app/services/activity/ActivityService.tsx
B0Y3R-AVA: Switched to `network.caip2Id ?? getSolanaCaip2ChainId(network.chainId)` in 2a2ed16d.

Kept a fallback because `caip2Id` is optional on `Network` in core-chains-sdk, and `NetworkService.fetchDeBankNetworks` builds the arb/bsc/op/matic/base entries without one. The hardcoded consts do set it.

On the `solana:1` case: cannot happen today, `resolveUnknownTokenSymbols` returns early for anything that is not SVM before it reaches this. Still the wrong thing to have reached for first, so good catch.

packages/core-mobile/app/new/features/recurringSwap/screens/RecurringSchedulesScreen.tsx
B0Y3R-AVA: No, not by default. It is optional on `Network`, and the DeBank-sourced EVM networks (arb, bsc, op, matic, base) get constructed in `NetworkService.fetchDeBankNetworks` with no `caip2Id` at all.

So it reads `activeNetwork.caip2Id ?? getCaip2ChainId(activeNetwork.chainId)` now, in 2a2ed16d. Prefers the real thing, derives only when the backend did not give us one.

packages/core-mobile/app/new/features/meld/components/SelectToken.tsx
B0Y3R-AVA: Agreed, and it bites every time I go near this code. Just in this PR: `NetworkContractToken`, `LocalTokenWithBalance`, `SPLToken`, the aggregator's `TokenInfo`, and Meld's `CryptoCurrency`.

Not untangling it here. Worth its own ticket so it stops living in review comments.

B0Y3R-AVA: Filed as CP-14953: https://ava-labs.atlassian.net/browse/CP-14953

Wrote down the shapes with file counts so the size of it is visible: `LocalTokenWithBalance` (~99 files), `MarketToken` (~24), `NetworkContractToken` (~21), `TokenWithBalance` (~14), Meld's `CryptoCurrency` (~9), plus the aggregator's `TokenInfo` and the one-off `SPLToken`. Seven mapper files exist only to bridge them.

Picking it up next.

packages/core-mobile/app/new/features/recurringSwap/screens/RecurringSchedulesScreen.tsx
gergelylovas: What is `NetworkService.fetchDeBankNetworks` for? Is it still needed?
The network list should come form the backend, the balances from debank are already handled by the balances service and the vm-modules. So there should be no need to custom fetch networks from debank.
Also we can easily just add the caip2id for these networks as we know all of them are `eip155:`

It's fine to do in a separate cleanup PR though.

gergelylovas (dismissed):

packages/core-mobile/app/new/features/recurringSwap/screens/RecurringSchedulesScreen.tsx
B0Y3R-AVA: Filed as CP-14956: https://ava-labs.atlassian.net/browse/CP-14956

Agreed on both parts, and I kept it as a separate ticket per your note.

Wrote it up investigate-first rather than delete-first, because I did not want to assume the backend list already covers arb/bsc/op/matic/base. Step 1 is comparing `fetchNetworks()` output against the five DeBank entries in mainnet and developer mode. If it is covered, `fetchDeBankNetworks` goes entirely along with the per-chain `/proxy/debank/v1/token` calls it fires on every `getNetworks`. If there is a gap, it keeps only the gap chains and sets `caip2Id` to `eip155:${chainId}` on what it constructs, since all five are EVM.

Also captured the part that made me trip over this in the first place: these networks are the only ones built without `caip2Id`, which is the sole reason the `?? getCaip2ChainId(...)` fallback exists here and the matching one exists in `ActivityService`. Once every `Network` reliably carries the field, both fallbacks come out and there is one source of truth again.

packages/core-mobile/app/new/common/hooks/useSearchableERC20AndSolanaTokenList.ts
B0Y3R-AVA: Withdrawing this one: the file is gone.

I rebased onto main and CP-14934 (#4033) had already deleted `useSearchableERC20AndSolanaTokenList` outright, so the rebase took the deletion. The `SolanaCaip2ChainId.MAINNET` change I made in 2a2ed16d went with the rest of the file, and the USDC-Solana lookup this PR added here is no longer part of the diff.

The SPL side of the Meld list gets rebuilt in #4036, so that is where this resurfaces if it resurfaces at all.

packages/core-mobile/app/new/common/hooks/useSearchableERC20AndSolanaTokenList.ts
B0Y3R-AVA: Same as above, this file no longer exists after the rebase onto main (CP-14934 / #4033 deleted it). So this is resolved, though by deletion rather than by the cleanup you asked for. The `.find()` I defended in my earlier reply is gone too.

Worth recording the reason it existed, because it still applies to anything built on `useSearchableTokenList`: that hook merges the account's entire cross-chain balance list into its output regardless of what you pass as `tokens` (useSearchableTokenList.ts:99-110). Narrowing its result back down afterwards looks redundant but is not. Anyone who deletes that step leaks every balance-bearing token into the list.

packages/core-mobile/app/new/common/hooks/useSolanaTokens.ts
B0Y3R-AVA: This flipped after the rebase, so updating my earlier answer.

`useSolanaTokens` now has to stay, and for a new reason: CP-14934 (#4033) added a consumer that did not exist when I wrote that reply, `useMeldContractTokenMap`. Deleting the hook breaks the build, so the rebase restored it. The consumer I originally cited, `useSearchableERC20AndSolanaTokenList`, is the thing that got deleted instead.

`useErc20ContractTokens` is in the same position. Its consumers on this branch are `useMeldContractTokenMap` (also new from #4033) and `useGetCChainToken`. #4037 removes the second, and #4036 deletes `useErc20ContractTokens`, `useCChainContractTokens`, `useEthereumContractTokens` and `useSolanaTokens` together once the Meld picker stops needing them. So yes to removing it, just not in this PR.

bohdan-struk-avalabs (approved):

PR #4036, opened 2026-08-13
CP-14936: Move Meld buy picker to v2 tokens API, remove remaining full token-list machinery

B0Y3R-AVA:
## Description

**Ticket: [CP-14936](https://ava-labs.atlassian.net/browse/CP-14936)**

Top of the tokenlist stack (**#4035 → #4037 → this**). Adopts the v2 tokens API for the one consumer that genuinely needs enumeration, then deletes every remaining piece of full-catalog machinery. After this PR, `module.getTokens()` and `getNetworkContractTokens` have **zero callers** and no code path anywhere in the app downloads a whole token catalog.

**v2 adoption — Meld buy-flow token picker.** The onramp picker was the last real enumeration consumer (users must be able to pick tokens they don't hold). It now runs on `GET /v2/tokens`: paginated infinite query (200/page, C-Chain + Ethereum), server-side `keyword` search (replacing the client substring filter over a frozen in-memory catalog), exact `address` filter for pasted contract addresses, infinite scroll (new additive `renderListFooter`/`onEndReached` wiring on the shared `ListScreen`/`SelectTokenScreen`), `keepPreviousData` so typing never blanks the list, held tokens merged on top, and Meld supported-currency filtering preserved. The old "build-once" zustand token index (structurally incompatible with server-side search) is deleted.

**`useMeldTokenWithBalance`** now resolves the selected Meld currency held-first, then via a targeted lookup — coverage widens (any EVM chain + SPL, vs. C-Chain/Ethereum + one hardcoded Solana address before), and it exposes `isLoading` so the amount screen shows a spinner instead of a "no balance" flash during the round trip.

**Deleted** (each verified zero references first): `useErc20ContractTokens` / `useCChainContractTokens` / `useEthereumContractTokens` / `useAvalancheErc20ContractTokens`, `useSearchableERC20AndSolanaTokenList`, `getNetworkContractTokens` (the last `module.getTokens()` call site), `isErc20`, the Meld token-index store slices + `useResetMeldTokenList`, and `useSearchableTokenList`'s `tokens` parameter (the CP-14918 short-circuit is now structural).

**Documentation.** New `docs/token-aggregator.md`: the usage policy (never fetch a whole catalog), which endpoint each feature uses, and the concrete reasons the app is not 100% v2 — no batch lookup on v2 (single `address` per request vs. N per lookup POST), no `internalId` resolution on v2 (how cross-chain canonicals like USDC/BTC.b are addressed), and no v2 price/markets endpoints (`lookupWithPrice`, `watchlist/markets` are v1-only). All verified against the live staging aggregator via on-device probes (2026-08-12). Closing those gaps is backend work; the doc lists the revisit triggers.

## Screenshots/Videos

JS-only change. On-device Android verification pending (see Testing); picker screenshots will be attached after the device pass.

## Testing

**Dev Testing**

- Full suite at the stack tip: 367 suites / 4,520 tests passing; real `tsc` (cache cleared) and eslint clean.
- New unit coverage: `useMeldOnrampTokenPool` (30 tests: pool union/dedupe, v2 shape mapping, `chainId` presence for Meld tradability matching, `keepPreviousData`).
- Two independent adversarial review passes + a fix-layer re-review; all Critical/Important findings fixed (notably: unheld v2 tokens failing Meld's `'chainId' in token` tradability check, missing infinite-scroll wiring, and search refetch unmounting the list/keyboard).
- End-state assertion: `rg "module\.getTokens|getNetworkContractTokens|useErc20ContractTokens|useSearchableERC20AndSolanaTokenList|useNetworkContractTokens" app/` → zero hits.
- **Pending**: physical-device smoke pass (Meld picker browse / infinite scroll / server-side search / unheld-token selection, offramp unchanged) — will be run and reported here before this leaves draft.

**QA Testing**

- Buy (Meld): picker must list tokens beyond held balances (e.g. JOE, PNG unheld); scrolling loads more; typing filters server-side without the list flashing away; selecting an unheld token reaches the amount screen; providers/quotes work for it.
- Withdraw (Meld): token list unchanged (held tokens only).
- Add custom token / send / portfolio: unchanged (regression watch from the base PRs).

## Checklist

[CP-14936]: https://ava-labs.atlassian.net/browse/CP-14936?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/core-mobile/docs/token-aggregator.md
```diff
@@ -0,0 +1,72 @@
...
+  this — never hand-build these keys.
+- **Lookup entries** carry decimals under `meta.decimals[caip2Id]` and logo
+  under `meta.logoUri`; there is no `contractType`.
+- **v2 token items** have no numeric `chainId` (join
+  `networks[networkCaip2Id].chainId` from the same response), no `color`
+  field, and a nullable `contractType` (null on native rows).
+
+## Revisit when
+
+- Backend ships v2 equivalents for batch lookup (with `internalId`) or
```
gergelylovas, line 69: In general it is fine not to have v2 from everything and as you state it doesn't even have full compatibility. It's intentional. The goal wasn't to switch to V2 for everything but to not use the unpaginated `/v1/tokens` and `/v1/tokenlist` endpoints and avoid fetching the massive token lists. 
In this sense mission accomplished, I don't think at the moment more migration is needed.

packages/core-mobile/app/new/features/meld/components/SelectServiceProvider.tsx
```diff
@@ -58,7 +58,11 @@ export const SelectServiceProvider = ({
     return crytoQuotes && crytoQuotes?.length > 0
   }, [crytoQuotes])
 
-  const token = useMeldTokenWithBalance({ category })
+  // TODO(CP-14936 follow-up): this screen is only reachable after SelectAmount,
```
gergelylovas, line 61: can you please create a ticket for this so we don't forget?

gergelylovas (dismissed):

B0Y3R-AVA:
Closing this in favour of #4039 + #4040, which reach the same goal by a smaller route.

**Why it changed shape.** CP-14934 (#4033) merged while this sat in review, and it rewrote the same Meld picker this PR rewrites — `TokenList`, `useMeldTokenWithBalance`, `SelectAmount`, `SelectToken`. So `useMeldOnrampTokenPool` and `useMeldSupportedCryptoCurrencies` ended up as two competing implementations of one thing, and rebasing would have meant resolving five semantic conflicts in order to then delete the side I'd fought for.

**What decided it.** I measured the input instead of guessing: Meld returns **149** onramp currencies, of which **29** sit on the chains the picker actually scopes to. And `/v1/token/lookup` accepts up to **500** tokens per request, at 241 ms for 500 versus 68 ms for 1 — so the cost is per-request, not per-token.

29 tokens fit in a single request. Pagination, infinite scroll, debounced server-side search and the dual search layer were all sized for a problem that does not exist. That matches your own framing on the doc thread here: the goal was never full v2 adoption, it was to stop fetching the massive lists.

**What shipped instead**

- **#4039** — batches the token-lookup layer so N tokens in a tick cost one request, with no call-site changes. Also fixes a live bug: `ActivityService` sent an unbounded array, so >500 unknown SPL mints got `INVALID_PAYLOAD` and left *every* symbol in that batch as "Unknown".
- **#4040** — points `useMeldContractTokenMap` at a batched lookup and deletes `useErc20ContractTokens`, `useCChainContractTokens`, `useEthereumContractTokens`, `useSolanaTokens` and `getNetworkContractTokens`. `module.getTokens()` is now called nowhere in the app.

Verified on a Pixel 8 Pro: the buy list is **35 rows before and 35 rows after, identical currency codes in identical order**, now resolved by one 27-token request instead of a 53,121-token download.

**Salvaged from here into #4040:** `docs/token-aggregator.md`, plus deleting `isErc20` and the dead `tokens` param on `useSearchableTokenList` — both of which only became dead once the catalog hooks went.

**Your two threads:**
- *`NetworkService.fetchDeBankNetworks`* — filed as CP-14956, with the `caip2Id` backfill and removing the derive-on-miss fallbacks it forced.
- *the v2-coverage doc* — the doc came across to #4040 unchanged; no further v2 migration is planned, per your note.

**Not carried over:** the `ListScreen`/`SelectTokenScreen` pagination wiring, since nothing paginates any more. Copilot's `Number(...) ?? 0` NaN catch is moot — #4033 already replaced that line on `main`.

Branch stays put if we ever want the v2 paginated pool back.

PR #4039, opened 2026-08-20
CP-14963: Batch token-aggregator lookups into one request per tick

B0Y3R-AVA:
## Description

**Ticket: [CP-14963](https://ava-labs.atlassian.net/browse/CP-14963)**

`useTokenLookup` issues **one HTTP request per token**. Five call sites use it and three of them pass arrays, so a screen showing 8 tokens makes 8 requests. Copilot flagged this on `RecurringSchedulesScreen` during CP-14936 review and the agreed plan was a batched variant that seeds the per-token cache entries. This is that.

No call site changes. Query keys, `staleTime`, dedupe and `combine` are untouched.

### The problem is a false trade-off

This repo already contains both halves of the answer, in two sibling hooks for two sibling endpoints:

| | requests | cache sharing |
|---|---|---|
| `useTokenLookup` (per-token queries) | **N** | per-token ✅ |
| `useTokensWithPrice` (one array-keyed query) | 1 | none ❌ |

Per-token queries are deliberate: a token resolved anywhere is free everywhere, and a set that changes by one member refetches only that member. `useTokensWithPrice` keys its query on the whole array, so it makes one request but shares nothing and refetches everything when membership changes.

We want both. So the fan-out stays and the *fetches* get merged underneath it.

```mermaid
flowchart TB
    subgraph before["BEFORE - 8 tokens, 8 requests"]
        direction TB
        B0["useTokenLookup([A..H])"]
        B0 --> BQ1["query A"] --> BH1["POST /v1/token/lookup"]
        B0 --> BQ2["query B"] --> BH2["POST /v1/token/lookup"]
        B0 --> BQ3["query C..H"] --> BH3["POST x 6 more"]
    end
    subgraph after["AFTER - 8 tokens, 1 request"]
        direction TB
        A0["useTokenLookup([A..H])"]
        A0 --> AQ1["query A"]
        A0 --> AQ2["query B"]
        A0 --> AQ3["query C..H"]
        AQ1 --> CO["coalescer"]
        AQ2 --> CO
        AQ3 --> CO
        CO --> AH["1 x POST /v1/token/lookup<br/>tokens: [A..H]"]
    end
```

Same 8 cache entries in both. Only the network calls collapse.

### How the coalescer works

Each `queryFn` stops fetching and instead enqueues its token, returning a promise that the flush settles later. All the `queryFn`s in a render pass land in the same queue, one microtask later the queue is drained into a single request, and each waiter gets its own slice of the response.

```mermaid
sequenceDiagram
    participant RQ as React Query
    participant Q as coalescer
    participant LT as lookupTokens
    participant API as POST /v1/token/lookup

    RQ->>Q: enqueue(A) - schedules flush
    RQ->>Q: enqueue(B)
    RQ->>Q: enqueue(C)
    Note over Q: microtask fires<br/>queue swapped out BEFORE awaiting
    Q->>LT: lookupTokens([A,B,C])
    LT->>API: tokens: [A,B,C]
    API-->>LT: { keyA:..., keyB:..., keyC:... }
    LT-->>Q: merged data + failedTokens
    Q-->>RQ: resolve A with { keyA }
    Q-->>RQ: resolve B with { keyB }
    Q-->>RQ: resolve C with { keyC }
    Note over RQ: 3 independent cache entries
```

Two details worth a reviewer's eye:

- **The queue is swapped out before the `await`.** A token enqueued mid-flight belongs to the next batch. Clearing afterwards would discard it without settling its promise, stranding that query in `isLoading` forever with no error and no retry.
- **Each waiter gets a slice, not the whole map.** Handing everyone the merged map would produce the same rendered result, but token A's cache entry would then physically contain a snapshot of token B, and `combine`'s `Object.assign` ordering could let a stale copy overwrite a fresh entry.

### Why coalesce at all

Because the endpoint's cost is almost entirely per-request, not per-token. Measured against the live endpoint from an internal build on a physical Pixel 8 Pro:

| tokens in one request | latency |
|---:|---:|
| 1 | 68 ms |
| 10 | 65 ms |
| 50 | 89 ms |
| 100 | 148 ms |
| 200 | 167 ms |
| 300 | 186 ms |
| 500 | 241 ms |

**500x the data for 3.5x the time.** On top of that, every request separately pays an App Check token fetch across the native bridge in `appCheckFetch`, plus a possible 401/403 path that repeats the whole request. On mobile, N scattered requests also hold the cellular radio awake where one would not.

The same probe established the ceiling: the endpoint accepts at most **500 tokens** and rejects more with `INVALID_PAYLOAD` / `too_big` (`maximum: 500, inclusive: true`). That number is now a named constant with the server's error recorded next to it.

### Bug fixed on the way

`ActivityService.populateFromTokenLookup` sent an **unbounded** token array inside a `try/catch` that only warns. A Solana account whose activity batch contained more than 500 distinct unknown mints got `INVALID_PAYLOAD`, the catch swallowed it, and **every** symbol in that batch silently stayed `"Unknown"` — not just the ones past the limit. It now chunks, and a partial failure keeps the chunks that succeeded.

### Unresolved is not the same as absent

```mermaid
flowchart LR
    R{"token's chunk"} -->|"request succeeded,<br/>token present"| F["resolve with its entry"]
    R -->|"request succeeded,<br/>token absent"| E["resolve empty<br/>(genuinely not in catalog)"]
    R -->|"request failed"| X["REJECT<br/>(unknown - let React Query retry)"]
```

This is why the primitive reports failed tokens separately instead of just returning data. Resolving a failed token as empty would let React Query cache *"this token does not exist"* for the whole `staleTime` and skip the retry that would have fixed it — turning a network blip into tokens silently missing from a list.

### Also in here

- `chunkArray` promoted from a private helper in `buildRequestItemsForAccounts` (7 call sites, unchanged) to a shared util. It gained a guard: a size of `0` or negative previously **hung in an infinite loop**, and `NaN` produced a bogus `[[]]`.
- `tokenToKey` moved from `useTokensWithPrice` to `common/utils/tokenLookup`, so all three key functions live together. They are easy to confuse and must not be crossed:

| function | purpose | format |
|---|---|---|
| `tokenToKey` | React Query **cache** key | `caip2:address`, all lowercase |
| `tokenLookupResponseKey` | reading the **response** map | `caip2-address`, Solana case preserved |
| `tokenLookupKey` | the address-pair half of the above | same |

## Screenshots/Videos

JS-only change with no UI surface of its own. Verified on a physical Pixel 8 Pro against a funded multi-chain wallet; every affected screen renders identically, with names, symbols, icons and network badges resolved off batched data.

## Testing

**Dev Testing**

iOS: n/a
Android: verified on a physical Pixel 8 Pro (internal debug build via Metro, real wallet)

Temporary instrumentation counted tokens enqueued vs HTTP requests issued, then reverted:

| Screen | tokens enqueued | HTTP requests | before |
|---|---:|---:|---:|
| Earn / DeFi market | 7, then 8 | **1 + 1** | 15 |
| Meld Buy → token picker | 1 | **1** | 1 |
| Swap | 3 | **1** | 3 |
| Borrow | 0 | **0** (all cache hits) | — |
| Portfolio, DeFi portfolio | 0 | 0 | 0 |

**21 requests became 3.** The DeFi market trace shows it plainly — seven separate `queryFn` calls, one flush, one request:

```
CP14936_ENQ queued=1 ... queued=7
CP14936_BATCH tokens=7
CP14936_HTTP  tokens=7
```

Borrow returning **zero** requests is the other half of the proof: its tokens were already cached from Deposit, so the per-token sharing served the whole screen with no network, and it still rendered LINK.e on Aave with icon, health score and APY.

Automated coverage — 32 suites / 365 tests, real `tsc -p .` clean, eslint `--max-warnings=0` clean:

- `tokenLookupRequest` (16) — chunk boundaries at 499/500/501, no body over 500, partial-chunk failure keeps the survivor's data, Solana response keys keep their casing.
- `tokenLookupQueue` (8) — N tokens in one tick produce one call, each caller gets only its own slice, failed tokens reject while batch-mates resolve, a thrown request leaves nothing pending, and a token enqueued mid-flight lands in the next batch.
- `useTokenLookup.batching` (5) — drives the **real** React Query rather than mocking `useQueries`: 8 tokens → 1 request; an already-cached token costs nothing; an overlapping set `[0,1]` then `[1,2]` requests only token 2.
- `chunkArray` (20) — including 9 invalid-size cases that now throw instead of hanging.
- `ActivityService` (20, 9 new) — 1/499/500 → 1 request, 501/1000 → 2, 1001 → 3; symbols resolve from every chunk; a failed chunk still resolves the survivor; unique mints are chunked rather than transactions.

Each non-obvious behaviour was mutation-tested — I broke the implementation and confirmed the right tests failed:

| mutation | caught by |
|---|---|
| clear the queue *after* the await | 2 coalescer tests (one as a 5s hang) |
| resolve with the whole map instead of a slice | 2 coalescer tests |
| chunking disabled | 3 ActivityService tests |

**Not verified on device:** `ActivityService`'s lookup path. This wallet has no Solana transaction history, so `resolveUnknownTokenSymbols` short-circuits with nothing to resolve. It stays covered by unit tests only; exercising it for real needs a wallet with SPL activity.

**QA Testing**

- Portfolio, Send, Activity, Swap, Earn (Deposit and Borrow), Buy and Recurring Swap should all be visually unchanged.
- Token names, symbols and logos must still resolve everywhere — the regression shape here is tokens quietly rendering without a symbol or icon, not an error.
- Navigate between Earn → Deposit → Borrow: the second screen should populate without a loading flash, since its tokens are already cached.
- Solana account with SPL activity: Activity symbols must resolve rather than showing "Unknown". Worth targeting an account with a lot of distinct unknown mints, as that is the chunking path.
- Airplane-mode mid-navigation, then restore: affected rows should recover on retry rather than staying permanently blank.

## Checklist

[CP-14963]: https://ava-labs.atlassian.net/browse/CP-14963?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core-mobile/app/new/common/utils/chunkArray.ts
+++ b/packages/core-mobile/app/new/common/utils/chunkArray.ts
@@ -1,0 +1,15 @@
+export const chunkArray = <T>(array: T[], size: number): T[][] => {
+  // A size of 0 or a negative one never advances `i`, so the loop below would
+  // spin forever instead of returning a wrong answer -- a hang is far more
+  // expensive to diagnose than a throw. NaN, Infinity and fractional sizes are
+  // rejected by the same check: they all produce silently misshapen chunks.
+  if (!Number.isInteger(size) || size < 1) {
+    throw new Error(`chunkArray: size must be a positive integer, got ${size}`)
+  }
+
+  const chunks: T[][] = []
+  for (let i = 0; i < array.length; i += size) {
+    chunks.push(array.slice(i, i + size))
+  }
+  return chunks
+}
--- a/packages/core-mobile/app/new/common/utils/tokenLookup.ts
+++ b/packages/core-mobile/app/new/common/utils/tokenLookup.ts
@@ -1,7 +1,7 @@
 import { BlockchainNamespace } from '@avalabs/core-chains-sdk'
 import {
-  Caip2IdAddressPair,
-  InternalId
+  type Caip2IdAddressPair,
+  type InternalId
 } from 'utils/api/generated/tokenAggregator/aggregatorApi.client'
 
 const EVM_NAMESPACE_PREFIX = `${BlockchainNamespace.EIP155}:`
--- a/packages/core-mobile/app/new/common/utils/tokenLookup.ts
+++ b/packages/core-mobile/app/new/common/utils/tokenLookup.ts
@@ -1,4 +1,8 @@
 import { BlockchainNamespace } from '@avalabs/core-chains-sdk'
+import {
+  Caip2IdAddressPair,
+  InternalId
+} from 'utils/api/generated/tokenAggregator/aggregatorApi.client'
 
 const EVM_NAMESPACE_PREFIX = `${BlockchainNamespace.EIP155}:`
 
@@ -27,3 +31,17 @@ export const tokenLookupKey = (caip2Id: string, address: string): string => {
   const key = `${caip2Id}-${address}`
   return isEvmCaip2Id(caip2Id) ? key.toLowerCase() : key
 }
+
+export const tokenToKey = (token: Caip2IdAddressPair | InternalId): string => {
+  if ('internalId' in token) {
+    return token.internalId.toLowerCase()
+  }
+  return `${token.caip2Id.toLowerCase()}:${token.address.toLowerCase()}`
+}
+
+export const tokenLookupResponseKey = (
+  token: Caip2IdAddressPair | InternalId
+): string =>
+  'internalId' in token
+    ? token.internalId.toLowerCase()
+    : tokenLookupKey(token.caip2Id, token.address)
--- a/packages/core-mobile/app/new/common/utils/tokenLookupRequest.ts
+++ b/packages/core-mobile/app/new/common/utils/tokenLookupRequest.ts
@@ -26,6 +26,21 @@ export const lookupChunk = async (
     client: tokenAggregatorApi,
     body: { tokens }
   })
+
+  /**
+   * `tokenAggregatorApi` is created with `throwOnError: true`, so a 500 or a
+   * surviving 401 already threw before we got here and `error` is undefined.
+   * Guard anyway: the generated signature still models `error` (its
+   * `ThrowOnError` generic defaults to `false`), so nothing but that one client
+   * flag stands between us and returning `{}` for a failed request -- which
+   * would cache "these tokens do not exist" instead of retrying.
+   */
+  if (response.error !== undefined) {
+    throw new Error(
+      `Token lookup request failed: ${JSON.stringify(response.error)}`
+    )
+  }
+
   return response.data?.data ?? {}
 }
 
@@ -49,10 +64,14 @@ export const lookupTokens = async (
     }
 
     /**
-     * This chunks request never landed so we know nothing about its tokens
-     * this is different than when the request succeeds but we don't have that token in the catalog on the backend
-     * Callers must reject these rather than resolve empty so that it can retry, otherwise it will React Query will
-     * cache this as empty until the next staleTime or garbage collection when a simple retry may have re populated it
+     * This chunk's request never landed, so we know nothing about its tokens.
+     * That is different from a request that succeeded without carrying a
+     * token, which tells us the backend catalog genuinely has no entry for it.
+     *
+     * Callers must reject these rather than resolve them empty. Resolving
+     * empty would have React Query cache "this token does not exist" until the
+     * next staleTime or garbage collection, skipping the retry that would have
+     * repopulated it.
      */
     failedTokens.push(...(chunks[index] ?? []))
   })
--- a/packages/core-mobile/app/new/common/utils/tokenLookupRequest.ts
+++ b/packages/core-mobile/app/new/common/utils/tokenLookupRequest.ts
@@ -1,0 +1,61 @@
+import {
+  postV1TokenLookup,
+  type Caip2IdAddressPair,
+  type InternalId,
+  type TokenInfo
+} from 'utils/api/generated/tokenAggregator/aggregatorApi.client'
+
+import { chunkArray } from 'common/utils/chunkArray'
+import { tokenAggregatorApi } from 'utils/api/clients/aggregatedTokensApiClient'
+
+export type LookupToken = Caip2IdAddressPair | InternalId
+
+export type LookupChunkResult = { [key: string]: TokenInfo }
+
+export type TokenLookupResult = {
+  data: LookupChunkResult
+  failedTokens: LookupToken[]
+}
+
+export const TOKEN_LOOKUP_CHUNK_SIZE = 500
+
+export const lookupChunk = async (
+  tokens: LookupToken[]
+): Promise<LookupChunkResult> => {
+  const response = await postV1TokenLookup({
+    client: tokenAggregatorApi,
+    body: { tokens }
+  })
+  return response.data?.data ?? {}
+}
+
+export const lookupTokens = async (
+  tokens: LookupToken[]
+): Promise<TokenLookupResult> => {
+  if (tokens.length === 0) {
+    return { data: {}, failedTokens: [] }
+  }
+
+  const chunks = chunkArray(tokens, TOKEN_LOOKUP_CHUNK_SIZE)
+  const settled = await Promise.allSettled(chunks.map(lookupChunk))
+
+  const data: LookupChunkResult = {}
+  const failedTokens: LookupToken[] = []
+
+  settled.forEach((result, index) => {
+    if (result.status === 'fulfilled') {
+      Object.assign(data, result.value)
+      return
+    }
+
+    /**
+     * This chunks request never landed so we know nothing about its tokens
+     * this is different than when the request succeeds but we don't have that token in the catalog on the backend
+     * Callers must reject these rather than resolve empty so that it can retry, otherwise it will React Query will
+     * cache this as empty until the next staleTime or garbage collection when a simple retry may have re populated it
+     */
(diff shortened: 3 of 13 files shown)


packages/core-mobile/app/new/common/utils/tokenLookupRequest.ts
```diff
@@ -0,0 +1,61 @@
...
+
+export const lookupChunk = async (
+  tokens: LookupToken[]
+): Promise<LookupChunkResult> => {
+  const response = await postV1TokenLookup({
+    client: tokenAggregatorApi,
+    body: { tokens }
+  })
+  return response.data?.data ?? {}
+}
```
bohdan-struk-avalabs, line 45: `response.error` is never checked. Without it, a 500 or a surviving 401 resolves to {}, the chunk counts as `fulfilled`, and every token in it resolves empty instead of landing in `failedTokens`

packages/core-mobile/app/new/common/utils/chunkArray.ts
meeh0w, line 1: Why not use [`lodash.chunk`](https://lodash.com/docs/#chunk)?

meeh0w (dismissed):
Approved with a nit

packages/core-mobile/app/new/common/utils/tokenLookupRequest.ts
```diff
@@ -0,0 +1,61 @@
...
+      Object.assign(data, result.value)
+      return
+    }
+
+    /**
+     * This chunks request never landed so we know nothing about its tokens
+     * this is different than when the request succeeds but we don't have that token in the catalog on the backend
+     * Callers must reject these rather than resolve empty so that it can retry, otherwise it will React Query will
+     * cache this as empty until the next staleTime or garbage collection when a simple retry may have re populated it
+     */
```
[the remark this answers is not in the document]
B0Y3R-AVA: Good catch, that comment was carrying the most important semantic in the file and reading badly while doing it. Rewritten in 597ad7f.

packages/core-mobile/app/new/common/utils/tokenLookup.ts
```diff
@@ -1,4 +1,8 @@
 import { BlockchainNamespace } from '@avalabs/core-chains-sdk'
+import {
+  Caip2IdAddressPair,
+  InternalId
+} from 'utils/api/generated/tokenAggregator/aggregatorApi.client'
```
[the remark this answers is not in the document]
B0Y3R-AVA: Agreed, and it was inconsistent with `tokenLookupRequest.ts` right next door, which already imports both as types. Fixed in 597ad7f.

packages/core-mobile/app/services/activity/ActivityService.tsx
```diff
@@ -191,11 +188,8 @@ export class ActivityService {
         caip2Id,
         address
       }))
-      const response = await postV1TokenLookup({
-        client: tokenAggregatorApi,
-        body: { tokens }
-      })
-      const data = response.data?.data ?? {}
+
+      const { data } = await lookupTokens(tokens)
```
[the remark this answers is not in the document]
B0Y3R-AVA: You're right, and this is the one real gap in the batch. `lookupTokens` resolves on a chunk failure by design, so the `catch` here only fires if the whole call throws, which it does not for a partial outage. Those mints just stayed "Unknown" with nothing recorded anywhere.

Fixed in 597ad7f: `populateFromTokenLookup` now destructures `failedTokens` and warns with the failed/total counts when it is non-empty. Covered by a new `ActivityService` test that fails a second chunk and asserts the warning.

packages/core-mobile/app/new/common/utils/tokenLookupRequest.ts
B0Y3R-AVA: Checked this properly. The client is created with `throwOnError: true` (`aggregatedTokensApiClient.ts:16`), and the generated client honours it on both failure paths, `client.gen.ts:112` for a fetch exception and `client.gen.ts:232` for a non-ok response. So a 500 or a surviving 401 throws before we reach that line, `Promise.allSettled` records the chunk as rejected, and every token in it lands in `failedTokens`. It never resolves to `{}` today.

That said, your reading of the code is the correct one: nothing in `lookupChunk` records that dependency, and the generated signature still models `error` because its `ThrowOnError` generic defaults to `false`. One flag flip in a file nobody would think to look at turns this into exactly the silent-empty-cache bug the rest of the PR is built to avoid.

So I added the guard anyway in 597ad7f, with a comment naming `throwOnError: true` so the next person knows why it looks dead. Two tests cover it: a response carrying `error` is reported as failed rather than absent, and a sibling chunk that succeeds still keeps its data.

packages/core-mobile/app/new/common/utils/chunkArray.ts
B0Y3R-AVA: Two reasons I kept the local one.

We do not have `lodash` here, only the per-function packages (`lodash.debounce`, `.isempty`, `.isstring`, `.merge`), so this would mean a new runtime dependency plus `@types/lodash.chunk` for six lines.

The bigger one is the guard. `lodash.chunk(arr, 0)` returns `[]` silently, and that is the behaviour I was specifically trying to get rid of when I promoted this out of `buildRequestItemsForAccounts`. That file passes seven different runtime-derived sizes into it, and a bad size previously either hung in an infinite loop or produced a plausible-looking wrong answer. Throwing is the point.

Happy to swap it if you feel strongly, but the throw is doing real work for the balance path.

bohdan-struk-avalabs (approved):

PR #4043, opened 2026-08-24
CP-14964: Validate Ledger device address/pubkey replies — BLE frame corruption fails closed instead of persisting a bare P- address

B0Y3R-AVA:
## Description

**Ticket: [CP-14964]**

A Ledger BLE frame that loses its payload bytes in transit but keeps the trailing status word parses in `@avalabs/hw-app-avalanche` as **success with an empty address** (`returnCode 0x9000`, `address: ""`). The app persisted that as `` `P-${""}` `` → the literal string `"P-"` for the wallet's X/P/CoreEth addresses. `"P-"` is truthy, so it passed every `!address` guard, and because stored Ledger addresses are never re-derived, the corruption was permanent, surfacing only as an unexplained "No routes" on C→P transfers (plus Sentry BA1/BA2 and API 400s). Two production users are affected.

This PR validates device replies at the boundary instead of trusting them:

- **New `app/services/ledger/validateDeviceAddress.ts`**: validators for every reply shape the Avalanche/Eth/Solana Ledger apps return.
  - `assertDeviceBech32Address`: return-code check (only when present, since `getETHAddress` replies legitimately lack one), non-empty body, strict bech32 parse, and **HRP match** against the requested network (`avax`/`fuji`), so a valid-but-wrong-network body also fails closed.
  - `assertDeviceEvmAddress`: viem's `isAddress`, which accepts all-lowercase addresses and enforces EIP-55 on mixed-case ones, composing with the existing `ethers.getAddress` checksum layer at the call site.
  - `assertDevicePublicKey` (33-byte compressed SECP256K1) and `assertDeviceSolanaAddress` (32 bytes): a truncated reply can also deliver a short or empty **pubkey** with `0x9000`, and pubkey→bech32 hashing does not validate length, so an empty pubkey silently derives a *well-formed* address nobody controls. That is a strictly worse failure than `"P-"` because it never surfaces an error. Wired into `getPublicKeysForRange`, the `getAvalancheKeys` Ledger Live branch, and `getSolanaKeys`.
- **`LedgerService.getAllAddresses`**: all three call sites validated (EVM, CoreEth, XP). The device returns a `P-` prefixed bech32 on the CoreEth path too, so it is corruptible identically. `getAvalancheKeys` now fails closed instead of `|| ''`.
- **`AvalancheWalletService.getReadOnlySigner`**: rejects bare-prefix `addressPVM` *and* `addressCoreEth` before handing them to the wallets-SDK, and now does so before the XP provider lookup so a corrupt stored address cannot be masked by an RPC error.
- **`stripAddressPrefix.ts`**: new `isBareChainPrefix` helper (`/^[XPC]-$/`).
- **`useQuoteStreaming`**: bare-prefix from/to addresses are blocked before reaching the Fusion SDK and surface the standard quotes-unavailable error plus a distinct Sentry message, preserving the telemetry that identified the affected users instead of leaving a silent blank state.
- **Diagnostic telemetry**, so the next occurrence in the wild identifies the cause rather than leaving it inferred. Each APDU exchange adds a Sentry breadcrumb carrying frame metadata only (`cla`, `ins`, `replyLength`, `statusWord`, never payload bytes), and a validation failure captures a structured event under a stable fingerprint with `publicKeyLength` and `hashLength`. Those two fields are the discriminator: a 33-byte pubkey and 20-byte hash alongside an empty address means the frame was cut exactly at the address boundary, whereas a device returning an empty address looks different. `ledgerAppType` and `ledgerAppVersion` are set on the Sentry scope for the same reason. Address values are capped to a 12-character preview so a real address cannot leak into Sentry.

**Root cause and verification.** `processGetAddrResponse` reads the status word from the reply's last two bytes and takes the address as whatever sits between the 20-byte hash and that status word. A frame that loses its address bytes but keeps its tail therefore parses as `returnCode 0x9000` with `address: ""`. Unit tests drive the real parser through a scripted transport to demonstrate this, which is also why a return-code check cannot catch it.

Production confirms the consequence. Issue BA1 carries `queryHash: [..., [""]]`, an `xpAddresses` array holding a single empty string, thrown from `getReadOnlySigner`. The same user's events run from `1.0.36+9322` through `1.0.37+9465`, so an app update does not clear it: stored Ledger addresses are never re-derived, which is why the state is permanent and why the repair migration is tracked separately in CP-14970.

On hardware (Pixel 8 Pro + Ledger Flex), with a dev-only truncation harness injecting that frame at the transport layer: unpatched, onboarding completes silently and persists `"P-"`, then emits the same bech32 error and recurring API 400s seen in production. Patched, the same fault stops onboarding with `empty address body` and persists nothing. With injection off, onboarding is unchanged and matches the device's known-good addresses, including after the review changes that route EVM validation through viem's `isAddress`.

Scope of that result: the harness supplies the corrupt frame, so it establishes that the mechanism is sufficient and that the guard covers it. Which fault produced the two users' value is a separate question and remains open. One user shows `TransportError: Invalid tag 8`, a genuine framing failure, fifty seconds before their first bad-address error; the other shows only connect-level failures. A device or app version returning an empty address with a success code is not excluded. This is deliberately why the change validates the reply rather than diagnosing the cause: any input that yields an empty or malformed address at this boundary now fails closed, including causes we have not characterised.

The xpub-based derivation path needs no guard: `bip32.fromPublicKey` strictly rejects truncated key material, verified empirically. That asymmetry, xpubs validated by bip32 while addresses were string-concatenated raw, was the whole bug.

**Adversarial review** (two independent passes plus a delta re-review) additionally verified: no partial persistence before the new throws, no transport-state wedge, and no cold-start regression for already-corrupted accounts. Review feedback is addressed in `a86562a22`. Deliberately deferred to **CP-14970** to keep this reviewable: the repair migration for the two corrupted users, consumption-side guards (`transformXPAddresses`, the `getAddressByNetwork` chokepoint and friends), validator-specific error copy on `AppConnectionScreen`, and Sentry `walletType` tagging.

No new dependencies (viem is already a direct dependency). Not breaking; no migration.

## Screenshots/Videos

- Unpatched + injected fault: onboarding completes normally, then Account Settings shows the corrupt state, with recurring `Malformed address` API 400s in the logs.
- Patched + injected fault: onboarding fails with a retryable error at the connect step and nothing is persisted.
- Patched + healthy device: addresses screen shows the correct `P-avax1…`, matching the device baseline.

## Testing

Happy path:
1. Onboard a Ledger (BLE, Avalanche app open) → wallet created, Account Settings → Addresses shows real `0x…`, `avax1…`, `bc1…` values.
2. Swap C→P (AVAX → AVAX on P-Chain) with a healthy wallet → quotes stream normally.

Edge/error states (require the truncation harness from the ticket notes, or the unit tests):
1. Truncated XP `getAddressAndPubKey` reply during onboarding → onboarding fails with `Ledger getAddressAndPubKey(XP) returned an invalid address: empty address body`, and retry works once the fault clears.
2. Account with persisted `"P-"` → swap shows the quotes-unavailable error rather than a blank screen and emits the CP-14964 Sentry message; staking reads throw a clear address-not-available error.
3. Unit suites: `validateDeviceAddress.test.ts` (truncated frames through the real parser, HRP mismatch, pubkey and Solana lengths, telemetry fingerprint and address-preview redaction), `stripAddressPrefix.test.ts`, `useQuoteStreaming.test.ts` (bare-prefix blocks plus error surfaced), `AvalancheWalletService.test.ts` (real `getReadOnlySigner` throws).

Regression focus: Ledger onboarding (BIP44 and Ledger Live derivation, mainnet and testnet), add-account, Solana enable, C→P and P→C transfers, and staking flows on a Ledger wallet. Expected: no behaviour change on healthy devices, and any device-reply corruption now fails the operation with an explicit error instead of persisting it.

## Checklist

[CP-14964]: https://ava-labs.atlassian.net/browse/CP-14964?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core-mobile/app/services/ledger/LedgerService.ts
+++ b/packages/core-mobile/app/services/ledger/LedgerService.ts
@@ -3,6 +3,8 @@ import Transport from '@ledgerhq/hw-transport'
 import AppAvalanche from '@avalabs/hw-app-avalanche'
 import AppSolana from '@ledgerhq/hw-app-solana'
 import { NetworkVMType } from '@avalabs/core-chains-sdk'
+import * as Sentry from '@sentry/react-native'
+import { AllowedSentryBreadcrumbCategory } from 'services/sentry/types'
 import {
   getAddressDerivationPath,
   handleLedgerError
@@ -102,7 +104,9 @@ class LedgerService {
         await new Promise(res => setTimeout(res, LEDGER_TIMEOUTS.REQUEST_DELAY))
       }
       try {
-        return await originalExchange(apdu)
+        const reply = await originalExchange(apdu)
+        this.recordApduBreadcrumb(apdu, reply)
+        return reply
       } catch (error) {
         // If error is still due to busy transport, reconnect
         if (
@@ -118,11 +122,52 @@ class LedgerService {
           await new Promise(res =>
             setTimeout(res, LEDGER_TIMEOUTS.REQUEST_DELAY)
           )
-          return await originalExchange(apdu)
+          const reply = await originalExchange(apdu)
+          this.recordApduBreadcrumb(apdu, reply)
+          return reply
         }
         // Other errors should be thrown immediately
         throw error
       }
+    }
+  }
+
+  // Frame-metadata-only breadcrumb (never payload bytes) for every APDU
+  // exchange. Sentry attaches recent breadcrumbs to any error captured
+  // afterwards, so a later validateDeviceAddress capture carries the reply
+  // length and status word — the signal that distinguishes a truncated
+  // transport frame from a device/app that legitimately returned empty (CP-14964).
+  private recordApduBreadcrumb(apdu: Buffer, reply: Buffer): void {
+    try {
+      Sentry.addBreadcrumb({
+        category: AllowedSentryBreadcrumbCategory.LedgerApdu,
+        level: 'info',
+        data: {
+          cla: (apdu[0] ?? 0).toString(16).padStart(2, '0'),
+          ins: (apdu[1] ?? 0).toString(16).padStart(2, '0'),
+          replyLength: reply.length,
+          statusWord: reply.subarray(-2).toString('hex')
+        }
+      })
+    } catch {
+      // Breadcrumb capture must never break a real APDU exchange.
+    }
+  }
+
+  // Session-global Sentry tags so any later Ledger error (e.g. a
+  // validateDeviceAddress capture) carries the app that produced it —
+  // distinguishes "this Avalanche app version returns empty addresses" from
+  // transport-layer frame corruption (CP-14964).
+  private recordLedgerAppSentryTags(
+    appType: LedgerAppType,
+    version: string
+  ): void {
+    try {
+      const scope = Sentry.getGlobalScope()
+      scope.setTag('ledgerAppType', appType)
+      scope.setTag('ledgerAppVersion', version)
+    } catch {
+      // Tagging must never break the connect/poll flow.
     }
   }
 
@@ -235,6 +280,7 @@ class LedgerService {
         Logger.info(`Immediately detected app type: ${detectedAppType}`)
         this.currentAppType = detectedAppType
         this.currentAppVersion = testAppInfo.version
+        this.recordLedgerAppSentryTags(detectedAppType, testAppInfo.version)
       } catch (error) {
         Logger.info(
           'Immediate get current app info failed, will rely on polling'
@@ -406,6 +452,7 @@ class LedgerService {
           this.currentAppType = newAppType
         }
         this.currentAppVersion = appInfo.version
+        this.recordLedgerAppSentryTags(newAppType, appInfo.version)
       } catch (error) {
         Logger.error('Error polling app info', error)
         // Don't stop polling on error, just log it
@@ -623,6 +670,7 @@ class LedgerService {
         this.currentAppType = detectedAppType
       }
       this.currentAppVersion = appInfo.version
+      this.recordLedgerAppSentryTags(detectedAppType, appInfo.version)
 
       if (this.isAppCompatible(this.currentAppType, appType)) {
         Logger.info(
@@ -863,6 +911,10 @@ class LedgerService {
           evmAvalancheAddressResponse,
           networkHrp
         )}`
+        const coreEthPublicKey = assertDevicePublicKey(
+          'getAddressAndPubKey(coreEth)',
+          evmAvalancheAddressResponse
+        )
         addresses.push({
           id: `${LedgerAddressType.AVALANCHE_CORE_ETH}-${i}`,
           type: LedgerAddressType.AVALANCHE_CORE_ETH,
@@ -909,10 +961,7 @@ class LedgerService {
         // Bitcoin addresses - derive from the Avalanche app public key at the
         // EVM derivation path (matching the browser extension behavior).
         const btcAddress = getBtcAddressFromPubKey(
-          Buffer.from(
-            evmAvalancheAddressResponse.publicKey.toString('hex'),
-            'hex'
-          ),
+          coreEthPublicKey,
           isTestnet ? networks.testnet : networks.bitcoin
         )
 
--- a/packages/core-mobile/app/services/ledger/LedgerService.ts
+++ b/packages/core-mobile/app/services/ledger/LedgerService.ts
@@ -19,10 +19,15 @@ import bs58 from 'bs58'
 import { DERIVATION_PATHS, LEDGER_TIMEOUTS } from 'new/features/ledger/consts'
 import { isBitcoinCompatibleApp } from 'new/features/ledger/utils'
 import { Curve } from 'utils/publicKeys'
-import { stripAddressPrefix } from 'common/utils/stripAddressPrefix'
 import { derivePublicKey, extendedPublicKeyToXpub } from 'utils/bip32'
 import { BluetoothState } from 'services/bluetooth/types'
 import BluetoothService from 'services/bluetooth/BluetoothService'
+import {
+  assertDeviceBech32Address,
+  assertDeviceEvmAddress,
+  assertDevicePublicKey,
+  assertDeviceSolanaAddress
+} from './validateDeviceAddress'
 import {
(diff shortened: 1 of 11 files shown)


packages/core-mobile/app/services/ledger/validateDeviceAddress.ts
```diff
@@ -0,0 +1,185 @@
...
+ * code can only be enforced when it is actually present.
+ */
+export interface LedgerAddressReply {
+  address?: string
+  returnCode?: number
+  errorMessage?: string
+}
+
+const EVM_ADDRESS_REGEX = /^0x[0-9a-fA-F]{40}$/
+
```
bohdan-struk-avalabs: Can you use `isAddress` instead from `viem`?

packages/core-mobile/app/services/ledger/validateDeviceAddress.ts
```diff
@@ -0,0 +1,185 @@
+/**
+ * Validation for address replies from the Avalanche Ledger app.
+ *
+ * `getAllAddresses` used to trust these replies unconditionally, so a reply
+ * carrying an empty address string was persisted as a bare chain alias
+ * (`P-${''}` === `'P-'`). That is truthy, so it survived every downstream
+ * emptiness guard and only surfaced much later as an unexplained "No routes"
+ * on a C->P transfer. See CP-14964.
+ */
```
meeh0w: I'd remove that comment; it's just noise at this point. No need to explain what another function used to do. The current implementation and its tests are enough documentation for what's expected.

What I would expect, though, is for all transactional features to validate the address (if they don't already), such that these weird UI states don't happen even if onboarding "fails successfully" for whatever reason.

meeh0w (dismissed):
Approved with a nit.

packages/core-mobile/app/services/ledger/validateDeviceAddress.ts
```diff
@@ -0,0 +1,185 @@
...
+  address?: string
+  returnCode?: number
+  errorMessage?: string
+}
+
+const EVM_ADDRESS_REGEX = /^0x[0-9a-fA-F]{40}$/
+
+const fail = (call: string, reason: string): never => {
+  throw new Error(`Ledger ${call} returned an invalid address: ${reason}`)
+}
```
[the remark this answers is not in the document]
B0Y3R-AVA: Good catch. Split the subject out so the message matches what actually failed, so a truncated public key now reads "returned an invalid public key" instead of blaming the address. Kept the address wording identical since the ticket notes and captured device logs quote it.

packages/core-mobile/app/services/ledger/validateDeviceAddress.ts
B0Y3R-AVA: Done, swapped the hand rolled regex for viem isAddress. Worth noting it is a bit stricter: it accepts all lowercase but enforces EIP-55 on mixed case, so a corrupt mixed case reply now fails right here with a clear message instead of falling through to ethers getAddress. Nothing that works today breaks, since that call site already threw on a bad checksum. Added a test that fails against the old regex and passes now, and confirmed a real Flex onboarding still completes on device.

packages/core-mobile/app/services/ledger/validateDeviceAddress.ts
B0Y3R-AVA: Removed the comment. On the wider point I agree, and that is the part I deliberately kept out of this PR to keep the diff reviewable. Filed CP-14970 for it.

The gaps I have so far: transformXPAddresses, where a bare prefix turns into a single empty string in xpAddresses and feeds the signing indices, then getAddressByNetwork as the single chokepoint that would cover swap, stuck funds and max amount at once, plus useHasXpAddresses, WalletConnect, claim fees, earn import/export and the receive QR. It also covers a migration to repair the two accounts already stored with a bare "P-", since nothing re-derives them today.

packages/core-mobile/app/services/wallet/AvalancheWalletService.ts
```diff
@@ -653,10 +656,14 @@ class AvalancheWalletService {
...
     const provXP = await NetworkService.getAvalancheProviderXP(isTestnet)
 
-    if (!account.addressPVM) {
+    if (!account.addressPVM || isBareChainPrefix(account.addressPVM)) {
       throw new Error('P-Chain address not available for account')
     }
 
+    if (!account.addressCoreEth || isBareChainPrefix(account.addressCoreEth)) {
+      throw new Error('CoreEth address not available for account')
+    }
```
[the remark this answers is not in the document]
B0Y3R-AVA: Agreed, moved both guards above the provider fetch. They only read local account data, so there is no reason to pay for an RPC lookup first or let a provider error mask a corrupted stored address.

packages/core-mobile/app/services/ledger/LedgerService.ts
```diff
@@ -848,8 +858,10 @@ class LedgerService {
         // underlying curve.
         const evmAvalancheAddressResponse =
           await avalancheApp.getAddressAndPubKey(evmPath, false, networkHrp)
-        const coreEthAddress = `C-${stripAddressPrefix(
-          evmAvalancheAddressResponse.address
+        const coreEthAddress = `C-${assertDeviceBech32Address(
```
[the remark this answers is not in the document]
B0Y3R-AVA: Good catch, fixed. That reply now gets its public key validated right next to the existing address check, and the validated Buffer feeds the BTC derivation directly, which also drops the pointless `Buffer.from(pk.toString('hex'), 'hex')` round trip.

Confirmed on device that it does not false positive: the harness truncates only the XP reply, so the CoreEth public key arriving here is genuine, and a Ledger Flex derivation still fails with the expected XP error rather than a public key error. No `invalid public key` anywhere in the logs.

Worth noting this reply is the one place where address and key material both matter, since the same response feeds CoreEth, the BTC address, and nothing re-derives either once persisted.

bohdan-struk-avalabs (approved):

PR #4044, opened 2026-08-24
CP-14968: Warn Keystone users in-app when the keystone feature flag is off

B0Y3R-AVA:
## Description

**Ticket: [CP-14968](https://ava-labs.atlassian.net/browse/CP-14968)**

Keystone is being retired across all Core platforms. Today, turning the master `keystone` PostHog gate off only disables signing: `KeystoneSignerScreen` mounts and immediately rejects, so the user sees a generic transaction failure that reads like a bug rather than an intentional sunset. Accounts and balances keep displaying normally, with no explanation anywhere in the app.

This adds in-app deprecation messaging so a user holding a Keystone wallet understands what happened and what to do next.

**What's in here**

- **Portfolio banner** — red inline notice under the balance header, shown only when the active wallet is Keystone *and* the master `keystone` gate is off. Tapping it opens the explainer.
- **Wallets screen badge** — red alert icon on Keystone wallet rows, opening the same explainer. The flag is read inside the row rather than passed as a prop, because `WalletHeaderRow`'s hand-written `arePropsEqual` does a field-level compare and would swallow a prop-borne flag change.
- **Shared explainer modal** — new `keystoneDeprecation` modal built on the same `ScrollScreen` + `renderFooter` pattern as `TransactionSuccessfulScreen`. Explains that signing is retired, that funds are safe, and that continuing with Core means moving funds to a supported wallet type. Links to the support article.
- **Signing path** — the flag-off branch in `KeystoneSignerScreen` now rejects and *replaces* the route with the explainer, instead of silently rejecting and going back. `replace` rather than `goBack` also keeps the existing `beforeRemove` POP listener from firing a second `onReject`.
**Deliberately out of scope: the X/P missing-signatures alert**

An earlier revision of this branch also rewrote the "Having trouble signing with Keystone?" alert in `vmModule/handlers/avalancheSendTransaction.ts` and deleted `DOCS_KEYSTONE_SIGNING_ERROR_URL`. That was reverted before review, and this PR no longer touches that file at all — the constant and its call site are unchanged from `main`.

The reason: that branch is gated on `walletType === KEYSTONE && !hasAllSignatures()`, **not** on the flag, and it is only reachable while Keystone is still enabled. With the gate off, `KeystoneSignerScreen` rejects on mount, so `await walletService.sign(...)` throws and control jumps to the handler's `catch` — the alert is never evaluated. So its "visit core.app" guidance only ever renders for a user whose Keystone wallet still works, which is the genuine multi-UTXO limitation it was written for and the guidance we want to keep there.

Whether that copy should change once Keystone is retired on every platform is a real question, but it is a different code path with a different trigger and belongs in its own ticket.

**Motivation and context**

Follows the Slack thread on what a user with one Keystone wallet experiences when the flag flips. The conclusion there was that behavior is acceptable but needs explicit deprecation warnings, and the treatment below matches the mock Amanda posted.

**Dependencies introduced:** none. Gated entirely on the existing `selectIsKeystoneBlocked` selector, with no feature-flag or selector changes. New Keystone imports remain separately gated by `keystone-onboarding` (CP-14967).

**Not breaking.** Nothing is hidden or blocked that wasn't already; this only adds explanation to a state that already existed.

## Screenshots/Videos

Captured on a physical Pixel 8 Pro (Android). iOS not captured.

1. **Portfolio banner** — renders under the balance header, alert icon vertically centred against the wrapped text.
2. **Explainer modal** — icon centred above the title, content vertically centred, actions pinned to the bottom.
3. **Wallets screen badge** — alert icon on the Keystone wallet row only; the mnemonic wallet below it is unmarked.
4. **Badge → modal** — tapping the badge opens the explainer without toggling row expansion.

## Testing

**Dev Testing**

Happy path (no Keystone wallet, or flag on):
1. With the `keystone` gate ON, confirm no banner on Portfolio, no badge in the wallets list, and Keystone signing behaves exactly as before.

Deprecation path:
1. Turn the master `keystone` gate OFF in PostHog.
2. With a Keystone wallet active, open Portfolio → red banner appears under the balance. Tap it → explainer modal, with Learn more opening the support article and Dismiss closing.
3. Open the wallets screen → red alert badge on the Keystone wallet row. Tap the badge → same explainer. Confirm tapping the badge does not toggle row expansion.
4. Trigger a transaction from the Keystone wallet → instead of the generic signing failure, the explainer modal appears and the request is rejected once.

Edge cases:
1. Switch the active wallet from Keystone to a mnemonic wallet → banner disappears; the badge remains on the Keystone row in the wallets list.
2. Multiple wallets where Keystone is NOT active → no Portfolio banner, but the badge still marks the Keystone row.
3. Flip the gate at runtime while Portfolio is mounted → banner appears/disappears without a restart.

**Verified on device** (physical Pixel 8 Pro, Android, gate forced off locally):

- Banner renders on Portfolio for the active Keystone wallet, and opens the explainer on tap.
- Badge appears on the Keystone wallet row and **not** on the mnemonic wallet row, confirming the per-row gating. Tapping it opens the explainer and does not toggle row expansion.
- Explainer modal layout confirmed against the design pattern used by `TransactionSuccessfulScreen`.

Not yet exercised on device: the dapp-send path through `KeystoneSignerScreen`. Reviewers should know one open question there — `ApprovalScreen` pops on an un-awaited `onApprove`, so it is worth confirming no stale `/approval` modal is left underneath the deprecation screen. Static tracing through expo-router 56.2.11 and `@react-navigation/core` 7.14.0 confirmed the reject path itself cannot hang or double-settle a request: `replace` delivers `REPLACE` (not `POP`) to the outgoing screen, so the `beforeRemove` guard correctly does not re-reject, and both promises are settle-once regardless.

**QA Testing**

Expected behavior: a user with a Keystone wallet can still see all accounts and balances after the flag flips. They cannot sign. Every path that previously dead-ended in a generic failure now explains the deprecation and links to the support article. Users with no Keystone wallet see no change anywhere.

## Checklist

- [x] I have verified the code works (Android, physical Pixel 8 Pro)
- [ ] I have included screenshots / videos of android and ios — Android only; iOS not captured
- [ ] I have updated the documentation — n/a

[CP-14968]: https://ava-labs.atlassian.net/browse/CP-14968?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core-mobile/app/new/features/keystone/hooks/useKeystoneDeprecation.test.ts
+++ b/packages/core-mobile/app/new/features/keystone/hooks/useKeystoneDeprecation.test.ts
@@ -1,0 +1,116 @@
+import { renderHook } from '@testing-library/react-hooks'
+import { WalletType } from 'services/wallet/types'
+import {
+  useIsActiveWalletKeystoneDeprecated,
+  useKeystoneDeprecation
+} from './useKeystoneDeprecation'
+
+const mockNavigate = jest.fn()
+jest.mock('expo-router', () => ({
+  useRouter: () => ({ navigate: mockNavigate })
+}))
+
+let mockIsKeystoneBlocked = false
+let mockActiveWallet: { type: WalletType } | undefined
+
+jest.mock('store/posthog', () => ({
+  selectIsKeystoneBlocked: () => mockIsKeystoneBlocked
+}))
+
+jest.mock('store/wallet/slice', () => ({
+  selectActiveWallet: () => mockActiveWallet
+}))
+
+// The selectors above are already stubbed to ignore state, so useSelector just
+// needs to invoke them; a real store would add nothing to these assertions.
+jest.mock('react-redux', () => ({
+  useSelector: (selector: (state: unknown) => unknown) => selector({})
+}))
+
+describe('useKeystoneDeprecation', () => {
+  beforeEach(() => {
+    jest.clearAllMocks()
+    mockIsKeystoneBlocked = false
+    mockActiveWallet = undefined
+  })
+
+  it('does not warn for a Keystone wallet while the gate is on', () => {
+    mockIsKeystoneBlocked = false
+
+    const { result } = renderHook(() => useKeystoneDeprecation())
+
+    expect(result.current.isKeystoneDeprecated).toBe(false)
+    expect(result.current.shouldWarnForWalletType(WalletType.KEYSTONE)).toBe(
+      false
+    )
+  })
+
+  it('warns only for Keystone wallets once the gate is off', () => {
+    mockIsKeystoneBlocked = true
+
+    const { result } = renderHook(() => useKeystoneDeprecation())
+
+    expect(result.current.shouldWarnForWalletType(WalletType.KEYSTONE)).toBe(
+      true
+    )
+    expect(result.current.shouldWarnForWalletType(WalletType.MNEMONIC)).toBe(
+      false
+    )
+    expect(result.current.shouldWarnForWalletType(WalletType.LEDGER)).toBe(
+      false
+    )
+  })
+
+  it('routes to the deprecation modal', () => {
+    const { result } = renderHook(() => useKeystoneDeprecation())
+
+    result.current.openDeprecationInfo()
+
+    expect(mockNavigate).toHaveBeenCalledWith({
+      pathname: '/keystoneDeprecation'
+    })
+  })
+})
+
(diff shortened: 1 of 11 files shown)


bohdan-struk-avalabs (approved):
Code looks good, if you don't use translations for the text :)

PR #4052, opened 2026-09-03
CP-14995: Remove Keystone code, keep imported wallets read-only

B0Y3R-AVA:
## Description

**Ticket: [CP-14995](https://ava-labs.atlassian.net/browse/CP-14995)**

Removes Keystone hardware-wallet support from Core Mobile while keeping existing imported Keystone wallets usable read-only, mirroring the extension's removal in ava-labs/core-extension#1020 (CP-14996).

**What an existing Keystone wallet can still do:** see balances, receive (view/copy addresses), be renamed, and be removed. The deprecation banner (portfolio), wallet-row badge, and explainer screen from CP-14968 are now unconditional — no feature flag involved.

**What is blocked:** all signing (send, swap, stake, message/typed-data, BTC PSBT, X/P, dapp/WalletConnect requests) via a single guard in `WalletFactory.createWallet` that throws `This wallet type is no longer supported` (same copy as extension). Adding accounts is blocked at the UI (dropdown + expanded-card row) and at the `addAccount` thunk, which also covers the `avalanche_addAccount` RPC — dapps get a clean JSON-RPC error instead of the previous hang-on-failure behavior. Keystone onboarding/import is deleted entirely.

**Kept deliberately (read paths):** `WalletType.KEYSTONE`, `KeystoneDataStorage` (stored xpubs), and pubkey/address derivation for existing accounts — `WalletService.getPublicKeyFor`/`getPublicKey` now derive Keystone pubkeys directly from stored xpubs (a faithful port of the deleted `KeystoneWallet` logic, preserving the typed `UNSUPPORTED_XP_DERIVATION` error contract that `ModuleManager.deriveKeystoneAddresses` relies on). Balance fetching never constructs a wallet object, so portfolios keep rendering.

**Feature flags:** `keystone` and `keystone-onboarding` are deleted from the code (extension parity). Flipping the orphaned flags in PostHog is a no-op; they can be archived in the console after this ships.

**Dependencies removed:** `@keystonehq/animated-qr`, `@keystonehq/keystone-sdk`, `@ngraveio/bc-ur` (plus their exclusive transitive cone, −464 lockfile lines), the `@keystonehq` jest transform entry, keystone jest manual mocks, and three orphaned keystone SVGs. No native (pod/gradle) changes — the integration was pure JS/TS.

Notes for reviewers:

- Signing attempts from a Keystone wallet surface the canonical "This wallet type is no longer supported" message (vmModule handlers pass it through instead of their generic "Failed to sign..." copy), and those expected failures log at `warn` instead of `error` so they don't pollute Sentry buckets for genuine signing bugs.
- The account-migration path (`shouldMigrateActiveAccounts`) still derives missing keystone accounts on first unlock — intentional: derivation-only, produces read-only accounts, and keeps existing portfolios complete.
- Removing a sole Keystone wallet is floor-protected like any other sole wallet (`walletsWithAccountsCount <= 1`) — pre-existing universal behavior, not keystone-specific; a keystone-only user resets via the general logout flow.
- `ModuleManager`'s KEYSTONE branch and `services/wallet/KeystoneWallet/errors.ts` are intentionally kept — they are the read-side derivation entry point and its error contract, not leftovers.

**Blocking at intent (added after the first review pass):** for a Keystone active wallet, the Send, Swap, Stake, Bridge and Withdraw action buttons now open the deprecation explainer instead of starting a flow that can only fail at the signing step. Buy, Receive and Hide/Unhide are untouched because they still work. The gate lives in the shared `ActionButtons` component, so every caller (portfolio, XP and non-XP token detail, collectibles, recurring schedules) is covered from one place, and the buttons stay enabled because a disabled button would swallow the tap and explain nothing. The `WalletFactory.createWallet` throw remains the backstop for dapp-initiated requests, where we do not own the entry point.

## Screenshots/Videos

Captured on a Pixel 8 Pro with a real Keystone wallet imported. Nothing changes for non-Keystone users.

### What a Keystone user sees now

| Portfolio: deprecation banner, balances still load | Tapping Send opens the explainer instead of the send flow | Receive still works, addresses unchanged |
| :---: | :---: | :---: |
| Keystone portfolio showing the deprecation banner above the action buttons | Keystone support has ended explainer, opened by tapping Send | Receive screen showing the Keystone account address |

### Add account: Keystone against a mnemonic wallet in the same list

| Keystone wallet: deprecation badge, no Add account row | Mnemonic wallet: Add account still there |
| :---: | :---: |
| My wallets, Keystone wallet expanded with a deprecation badge and no Add account row | My wallets, mnemonic wallet expanded showing its Add account button |

Both wallets are expanded in the right-hand shot, so the difference is visible in a single screen rather than across two states.

### Send tap opening the explainer

https://github.com/user-attachments/assets/7c133ed3-0f90-4e14-a40e-c14a3f9c0cc8

iOS was not exercised. The change is JS only with no native diff, and the same shared `ActionButtons` component backs both platforms.

## Testing

- Full unit suite: 378 suites / 4,656 tests, exit 0. Type check (`tsc --noEmit -p .` with `tsbuildinfo` cleared): exit 0. Lint (`--max-warnings=0`) on every changed file: exit 0.
- Read-path proof: every `WalletFactory.createWallet`/`getOrCreateWallet` call site audited, none reachable for `WalletType.KEYSTONE` except the intentional signing throw; balance polling and receive use stored account addresses only.
- Verified on hardware (Pixel 8 Pro). A real Keystone wallet was imported on a build of `main`, then the same install was pointed at this branch's bundle so the existing wallet state carried over untouched:
  - Addresses derive identically before and after removal. Both accounts' EVM addresses and the primary account's X/P address match byte for byte, which exercises `WalletService.getKeystonePublicKeyFor` against the real stored xpubs. The unit tests only cover that path with `KeystoneDataStorage` mocked.
  - The non-primary account still offers no X/P receive network, so the `UNSUPPORTED_XP_DERIVATION` contract survives. Confirmed in the logs: `Keystone non-primary account has no X/P addresses; deriving EVM/BTC only`.
  - Portfolio balances render, receive shows addresses, and the banner, badge and explainer all appear. The Keystone wallet has no Add account row on the expanded card while a mnemonic wallet in the same list still does.
  - The running bundle is genuinely free of the removed dependencies: 11,479 modules against 12,297 for a build with them still installed.
  - Approving a C-chain send (before the intent gate landed) surfaced the canonical `This wallet type is no longer supported`. With the gate in place, Send opens the explainer instead.
  - Cold start and unlock re-derive the Keystone accounts correctly.
  - Not exercised on device: Swap, Stake, Bridge and Withdraw individually (covered by unit tests through the same predicate), the dapp/WalletConnect path, rename, and remove.

- Wallet with a previously imported Keystone wallet: balances/receive/rename/remove work; every signing action fails with "This wallet type is no longer supported"; no "Add account" option appears for the Keystone wallet (dropdown or expanded card); "Add using Keystone" is gone from onboarding/access flows.
- Dapp connected to a Keystone account (WalletConnect or in-app browser): connection restores; signing requests are rejected with the same message; `avalanche_addAccount` returns an error instead of hanging.
- Send, Swap, Stake, Bridge and Withdraw on a Keystone wallet should open the "Keystone support has ended" explainer immediately rather than starting the flow. Buy and Receive should still work normally.
- Expected: blocked signing attempts are intentional and now show the deprecation copy — not a bug.

## Checklist

[CP-14995]: https://ava-labs.atlassian.net/browse/CP-14995?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/core-mobile/app/services/wallet/WalletService.tsx
```diff
@@ -406,22 +418,36 @@ class WalletService {
...
+        : async (params: {
+            derivationPath: string
+            curve: Curve
+          }): Promise<string> => {
+            const wallet = await WalletFactory.getOrCreateWallet({
+              walletId,
+              walletType
+            })
+            return wallet.getPublicKeyFor(params)
+          }
```
meeh0w: This isn't very readable 😆 Can we name the two functions before deciding which one becomes the `derivePublicKey`?

meeh0w (dismissed):
Left a minor comment.

packages/core-mobile/app/store/rpc/handlers/account/avalanche_addAccount/avalanche_addAccount.ts
```diff
@@ -60,7 +82,16 @@ class AvalancheAddAccountHandler
...
-    await dispatch(addAccount(walletId)).unwrap()
+    try {
+      await dispatch(addAccount(walletId)).unwrap()
+    } catch (error) {
+      Logger.error('avalanche_addAccount: failed to add account', error)
+      return {
+        success: false,
+        error: rpcErrors.internal(extractAddAccountErrorMessage(error))
+      }
+    }
```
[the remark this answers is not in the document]
B0Y3R-AVA: Good catch, fixed in a28cde367. Worth noting for anyone reading later: the obvious fix of reusing the shared `isUnsupportedWalletTypeError` predicate doesn't work at this call site. `unwrap()` rethrows RTK's `miniSerializeError` output, which is a plain object and never an `Error`, so the predicate's `instanceof` check would always be false and we'd still log at error. This now matches on the extracted message instead, and there are tests covering both directions: the blocked Keystone case logs at warn, a genuine failure still logs at error.

packages/core-mobile/app/services/wallet/WalletService.tsx
B0Y3R-AVA: Fair, that was dense. a28cde367 names both paths, `deriveFromStoredKeystoneXpub` and `deriveFromWallet`, with a shared `DerivePublicKeyParams` type, so the ternary now just picks between two named functions.

meeh0w (approved):

